PR #1, opened 2021-04-22
Bind-join for BrTPF

chengsijin0817:
Include an executable bind join operator for BrTPF, corresponding physical operator, and a test script for ExecOpBindJoinBRTPF

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericBindJoin.java
+++ b/src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericBindJoin.java
@@ -1,0 +1,78 @@
+package se.liu.ida.hefquin.queryplan.executable.impl.ops;
+
+import se.liu.ida.hefquin.data.SolutionMapping;
+import se.liu.ida.hefquin.data.jenaimpl.JenaBasedSolutionMapping;
+import se.liu.ida.hefquin.data.jenaimpl.JenaBasedSolutionMappingUtils;
+import se.liu.ida.hefquin.federation.FederationMember;
+import se.liu.ida.hefquin.query.Query;
+import se.liu.ida.hefquin.queryplan.executable.IntermediateResultBlock;
+import se.liu.ida.hefquin.queryplan.executable.IntermediateResultElementSink;
+import se.liu.ida.hefquin.queryproc.ExecutionContext;
+
+import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


src/test/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpTestBase.java
```diff
@@ -42,6 +37,13 @@
 		public TPFInterface getInterface() { return iface; }
 	}
 
+	protected static class BRTPFServerForTest implements BRTPFServer
+	{
+		final BRTPFInterface iface = new BRTPFInterfaceImpl();
+
+		@Override
+		public BRTPFInterface getInterface() { return iface; }
+	}
```
hartig, line 46: Can you please add an empty line between this method and the next one.

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericBindJoin.java
```diff
@@ -0,0 +1,78 @@
...
+        this.query = query;
+        this.fm = fm;
+    }
+
+    @Override
+    public int preferredInputBlockSize() {
+        // Since this algorithm processes the input solution mappings
+        // sequentially (one at a time), and input block size of 1 may
+        // reduce the response time of the overall execution process.
+        return 1;
```
hartig, line 35: This does not make sense for bind joins.

chengsijin0817: I have updated the preferredInputBlockSize for bind join (See commit db1ce88). 
According to the experiment result of the BrTPF paper, I choose 30 as the initial block size for bind join, but it can be optimized in query planning. Please let me know if it makes sense.

src/test/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpTestBase.java
chengsijin0817: Thanks, added (See commit db1ce88).

hartig:
Thanks Sijin. That's sufficient for the moment. I will merge this PR.

If you have time, think about #3 as a more efficient replacement of the approach implemented in this PR

hartig (approved):

Issue #3, opened 2021-04-22
Implementation of int.result block that is like a temporary index... 


hartig:
...with new types of methods that allow for query/index-style probing rather than just an iterator.

Once this is in place, the bind join algorithms should be re-implemented to leverage this new class (in particular, if the input block is of this new type, use it to probe with each solmap from the server response; if the input block is not of this type, use this type when populating the data structure for the request to the server).

hartig:
Now that we have the implementations of SolutionMappingsIndex in place, this TODO here can be attempted.

PR #5, opened 2021-04-29
symmetric hash join

chengsijin0817:
add symmetric hash join: executable operator, physical operator, and unit test


src/main/java/se/liu/ida/hefquin/queryplan/executable/IntermediateResultBlock.java
```diff
@@ -15,5 +16,9 @@
 	 * Returns an iterator over the intermediate result elements in this block.
 	 */
 	Iterator<SolutionMapping> iterator();
-	
+
+	/**
+	 * Returns an ArrayList over the intermediate result elements in this block.
+	 */
+	List<SolutionMapping> ArrayList();
```
hartig, line 23: Two comments about this:

First, please use the standard naming convention for method names. That is, method names begin with a lower case letter. Hence, it should be 'arrayList' in this case or, actually, better would be 'getArrayList'.

Second, I would not want to add this method to this interface. We may have implementations of this interface in which the solution mappings are not just in a list but in a more sophisticated data structure (e.g., some form of an index -- see #3). For such implementations, it would be an unnecessary and inefficient overhead to create a list that can be returned here. Hence, wherever you wanted to use this list, use the iterator instead.

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.jena.sparql.engine.join.JoinKey;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Collection;
+import java.util.List;
+
+public abstract class ExecOpGenericJoin implements BinaryExecutableOp {
+
+    protected Collection<Var> getVariables( final JenaBasedSolutionMapping JBSolMap){
```
hartig, line 17: This function should better be moved to JenaBasedSolutionMappingUtils.

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,43 @@
...
+        final Binding b = JBSolMap.asJenaBinding();
```
hartig, line 18: Please use the usual naming convention for variable and parameter names. They should begin with a lower case letter. Hence, instead of 'JBSolMap', it should be 'jbSolMap'.

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,43 @@
...
+        final Iterator<Var> variables = b.vars();
+        Collection<Var> vars = new ArrayList<>();
```
hartig, line 20: vars can be final

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,43 @@
...
+        while (variables.hasNext()) {
+            final Var v = variables.next();
+            vars.add(v);
+        }
+        return vars;
+    }
+
+    public List<Var> findJoinVars(Collection<Var> varL, Collection<Var> varR){
```
hartig, line 28: I think that this method can be static.

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,43 @@
...
+        while(keyIt.hasNext()){
+            keys.add(keyIt.next());
+        }
+        return keys;
+    }
+
+    protected abstract void process(
+            final IntermediateResultBlock inputL,
+            final IntermediateResultBlock inputR,
+            final IntermediateResultElementSink sink);
```
hartig, line 42: What is the purpose of this? You should have abstract methods only if they are used from other methods within the abstract class.

src/main/java/se/liu/ida/hefquin/queryproc/ExecutionContext.java
```diff
@@ -5,14 +5,23 @@
 public class ExecutionContext
 {
 	protected final FederationAccessManager fedAccessMgr;
+	protected final int index;
```
hartig: What is the purpose of this?

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,58 @@
...
+            final IntermediateResultBlock inputL,
+            final IntermediateResultBlock inputR,
+            final IntermediateResultElementSink sink){
+        final List<SolutionMapping> li1= inputL.ArrayList();
+        final List<SolutionMapping> li2= inputR.ArrayList();
+        Collection<Var> varL = new ArrayList<>();
+        Collection<Var> varR = new ArrayList<>();
+
+        if (inputL.size() > 0){
+            varL = getVariables((JenaBasedSolutionMapping) li1.get(0));
```
hartig, line 51: It is not guaranteed that all solution mappings coming in from the left input have the same variables. There may be a union operator in the left subtree or, eventually, we may also add support for a left join operator (OPTIONAL).
The same holds for the right input as well.

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,58 @@
...
+        while(keyIt.hasNext()){
+            keys.add(keyIt.next());
+        }
+        return keys;
+    }
+
+    protected void preprocess(
+            final IntermediateResultBlock inputL,
+            final IntermediateResultBlock inputR,
+            final IntermediateResultElementSink sink){
```
hartig, line 44: Where is this used in this class? If it isn't, then I would not put it here.

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpGenericJoin.java
```diff
@@ -0,0 +1,58 @@
...
+import se.liu.ida.hefquin.queryplan.executable.IntermediateResultElementSink;
+import org.apache.jena.sparql.engine.join.JoinKey;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Collection;
+import java.util.List;
+
+public abstract class ExecOpGenericJoin implements BinaryExecutableOp {
+    protected List<Var> keys;
```
hartig, line 17: This member variable seems weird.

First of all, it does not seem to be initialized properly.

Second, should it really be mutable? Are there any guarantees about when does it change, and through which means does it change? 

Third, the name does not really make clear what this actually is about. What are "keys" in the context of join algorithms? I assume you mean "join variables"?

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpSymmetricHashJoin.java
```diff
@@ -0,0 +1,91 @@
...
+            final Binding LHS = JBSolMapL.asJenaBinding();
+
+            // composite key
+            final ArrayList<Node> valKeyL = new ArrayList<>();
+            for (Var v : keys) {
+                valKeyL.add(LHS.get(v));
+            }
+            // Insert into left hash map
+            if (! valKeyL.isEmpty()){
+                mapL.put(new ListKey(valKeyL), JBSolMapL);
```
hartig, line 45: This replaces the previous entry at the given hash key (if there was any). That must not happen!

Note, multiple solution mappings may have the same hash key (because they have the same values for the join variables). They must all be kept in the hash table!

src/main/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpSymmetricHashJoin.java
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            // Use the listKey to probe the right hash table
+            final JenaBasedSolutionMapping MatchSolMapR = mapR.get(new ListKey(valKeyL));
```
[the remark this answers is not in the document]
hartig: Note that my second point here is related to my comment on line 45 above.

src/test/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpSymmetricHashJoinTest.java
```diff
@@ -0,0 +1,207 @@
...
+import se.liu.ida.hefquin.queryplan.executable.impl.MaterializingIntermediateResultElementSink;
+
+import java.util.Iterator;
+
+import static org.junit.Assert.*;
+
+public class ExecOpSymmetricHashJoinTest extends ExecOpTestBase
+{
+	@Test
+	public void JoinWithOneJoinVariable(){
```
hartig, line 20: Also in this file, please adhere to the usual naming conventions of method names.

src/test/java/se/liu/ida/hefquin/queryplan/executable/impl/ops/ExecOpSymmetricHashJoinTest.java
```diff
@@ -0,0 +1,207 @@
...
+				var2, NodeFactory.createURI("http://example.org/y1"),
+				var3, NodeFactory.createURI("http://example.org/z2")) );
+		input2.add( JenaBasedSolutionMappingUtils.createJenaBasedSolutionMapping(
+				var2, NodeFactory.createURI("http://example.org/y2"),
+				var3, NodeFactory.createURI("http://example.org/z3")) );
+
+		final MaterializingIntermediateResultElementSink sink = new MaterializingIntermediateResultElementSink();
+
+		final ExecOpSymmetricHashJoin op = new ExecOpSymmetricHashJoin();
+		op.preprocess(input1, input2, sink);
```
hartig, line 47: This method is protected. You shouldn't even be able to get this line compiled. 

Moreover, even if this method was public, it would never be called during normal processing. During normal processing, only the four methods of the BinaryExecutionOp interface are called.

hartig:
I have finished reviewing this PR. As you can see from my comments, there are several issues with this code.

While most of these issues may be fixed by changing the given code, I realize that there is one issue that is a bit more complicated. Namely, it is not straightforward to figure out what variables should be considered when generating the hash key. It fact, it is not even straightforward to figure out what the join variables may be. I think that this issue requires a more general solution. I will work on such a solution.

hartig (changes requested):

hartig:
I have finished reviewing this PR. As you can see from my comments, there are several issues with this code.

While most of these issues may be fixed by changing the given code, I realize that there is one issue that is a bit more complicated. Namely, it is not straightforward to figure out what variables should be considered when generating the hash key. It fact, it is not even straightforward to figure out what the join variables may be. I think that this issue requires a more general solution. I will work on such a solution (see #6).

hartig:
Sijin, now that #6 has been addressed, you can in principle continue to think about implementing the SHJ algorithm that you have started in this PR. The main difference is that, now, the method `createExecOp` in `PhysicalOperator` will be called with information about the variables that can be expected in the solution mappings that will be passed to the `ExecutableOperator`. This information can be used in the constructor of the concrete implementation of `ExecutableOperator` (e.g., to set up relevant data structures accordingly).

You may have to start in a new branch created from the current state of the main branch. However, please do not close this PR yet. My comments here may serve you as a basis for avoiding similar issues as you go along.

Also, perhaps it is better to first implement the hash table used by the SHJ algorithm. I have created an interface (`SolutionMappingsIndex`) and an empty implementation of it (`SolutionMappingsHashTable`). You can implement the latter as indicated in #10.

hartig:
@chengsijin0817 I think we can remove this PR now. #42 has replaced it. Do you agree?

chengsijin0817:
> @chengsijin0817 I think we can remove this PR now. #42 has replaced it. Do you agree?

Yes, please remove this PR. Thanks ;-)

hartig:
Closing this PR as discussed above.

PR #35, opened 2021-05-06
implement SolutionMappingsHashTable; add unit tests for SolutionMappingsHashTable

chengsijin0817:


src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
+import org.apache.jena.sparql.engine.binding.Binding;
 import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.queryplan.ExpectedVariables;
 
 public class SolutionMappingsHashTable extends SolutionMappingsIndexBase
 {
+	protected final Map<List<Node>, List<SolutionMapping>> map = new HashMap<>();
+	protected final Set<Var> joinVariables;
+
+	public SolutionMappingsHashTable(final ExpectedVariables... inputVars){
```
hartig, line 17: What's the purpose of passing potentially multiple `ExpectedVariables` objects to this constructor? Isn't it enough to pass a `Set<Var>` instead?

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
-		// TODO Auto-generated method stub
-		return false;
+		return map.isEmpty();
 	}
 
 	@Override
 	public boolean contains( final Object o ) {
-		// TODO Auto-generated method stub
-		return false;
+		return map.containsKey(o);
```
hartig, line 37: That's incorrect. You have to check whether `o` is a `SolutionMapping` that is contained in this index.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
 	}
 
 	@Override
 	public Iterator<SolutionMapping> iterator() {
```
hartig, line 41:
```suggestion
	public Iterator<SolutionMapping> iterator() {
		// The following implementation of this method is inefficient because,
		// each time it is called, it creates and populates a list of solution
		// mappings by iterating over the content of this index. However,
		// for a thread-safe implementation, we may have to live with this.
		// TODO: is there a more efficient way to implement this method?
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
 
 	@Override
 	public boolean add( final SolutionMapping e ) {
-		// TODO Auto-generated method stub
-		return false;
+		final Binding solMapBinding = e.asJenaBinding();
+		final List<Node> valKeyL = new ArrayList<>();
+
+		for (Var v : joinVariables) {
+			valKeyL.add(solMapBinding.get(v));
```
hartig, line 56: There should be a check that the given solution mappings is indeed defined for all of these variables. Throw an illegal argument exception if it is not. This would be mainly a safety measure in case there are programming errors in the code that is using this data structure.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
+		}
+
+		final List<SolutionMapping> solMapList = new ArrayList<>();
+		if (map.containsKey(valKeyL)) {
+			solMapList.addAll(map.get(valKeyL));
+		}
+		solMapList.add(e);
+
+		map.put(valKeyL, solMapList);
```
hartig, line 65: In cases in which the `map` already contains a list for the given key, it would be better to add `e` to that list instead of creating a new list every time.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
 import org.apache.jena.graph.Node;
 import org.apache.jena.sparql.core.Var;
 
+import org.apache.jena.sparql.engine.binding.Binding;
 import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.queryplan.ExpectedVariables;
 
 public class SolutionMappingsHashTable extends SolutionMappingsIndexBase
 {
+	protected final Map<List<Node>, List<SolutionMapping>> map = new HashMap<>();
```
hartig, line 14:
```suggestion
	// Having List<Node> as key type for the hash table is probably
	// not the best choice in terms of efficiency. However, it will have
	// to do for the moment.
	// TODO: can this be made more efficient?
	protected final Map<List<Node>, List<SolutionMapping>> map = new HashMap<>();
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
 	{
-		// TODO Auto-generated method stub
-		return null;
+		final Binding solMapBinding = sm.asJenaBinding();
+		final List<Node> valKeyL = new ArrayList<>();
+		List<SolutionMapping> matchingSolMaps = new ArrayList<>();
+
+		for (Var v : joinVariables) {
+			valKeyL.add(solMapBinding.get(v));
+		}
```
hartig, line 84: Given that this is the same as what you also do in the `add` method, I suggest to move this out into a separate method.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
+
+		if (! valKeyL.isEmpty()){
+			matchingSolMaps = map.get(valKeyL);
```
hartig, line 87: This may be incorrect in various cases! You assume that the given solution mapping either has none of the variables on which the index is built (in which case `valkeyL` would be empty) or it has all of these variables. You cannot make this assumption in general. The given solution mapping may have only a subset of these variables, in which case `valkeyL` would be a totally unexpected index key.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,155 @@
...
 
 	@Override
 	public Iterable<SolutionMapping> getJoinPartners( final SolutionMapping sm )
 			throws UnsupportedOperationException
 	{
-		// TODO Auto-generated method stub
-		return null;
+		final Binding solMapBinding = sm.asJenaBinding();
+		final List<Node> valKeyL = new ArrayList<>();
+		List<SolutionMapping> matchingSolMaps = new ArrayList<>();
```
hartig, line 80: You should not create the ArrayList object here. The if branch of the if-else statement below does not use it. Hence, create such an object only in the else branch.

hartig (changes requested):

hartig:
I have taken a first look and added some initial comments.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+
+	public SolutionMappingsHashTable(final Set<Var> joinVariables){
+		this.joinVariables = joinVariables;
+	}
+
 	@Override
 	public int size() {
-		// TODO Auto-generated method stub
-		return 0;
+		return map.size();
```
hartig, line 27: That's incorrect. `map.size()` gives you only the number of buckets in the hash table. However, this method should return the number of solution mappings that are in the hash table.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
 	}
 
 	@Override
 	public boolean isEmpty() {
-		// TODO Auto-generated method stub
-		return false;
+		return map.isEmpty();
```
hartig, line 32: This may also be incorrect. Depending on how the `remove` method is implementation (e.g., in possible subclasses), there may be entries in `map` but they all have empty lists as their values. Hence, if `map` is not empty, you would still have to check whether there are non-empty buckets in the hash table.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+		final List<Node> valKeys = getVarKeys(e);
+
+		if (map.containsKey(valKeys)) {
+			map.get(valKeys).add(e);
+		}
+		else {
+			final List<SolutionMapping> solMapList = new ArrayList<>();
+			solMapList.add(e);
+			map.put(valKeys, solMapList);
+		}
```
hartig, line 67:
```suggestion
		List<SolutionMapping> solMapList = map.get(valKeys);
		if ( solMapList == null) {
			solMapList = new ArrayList<>();
			map.put(valKeys, solMapList);
		}

		solMapList.add(e);
```

hartig: My suggested change above may be a bit more efficient because, in contrast to your code, mine requires only one lookup in `map`. Yours does two lookups: first the `containsKey` and then the `get`.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+		return true;
 	}
 
 	@Override
 	public void clear() {
-		// TODO Auto-generated method stub
-		
+		map.clear();
```
hartig, line 73: Before clearing `map`, can you also clear each of the lists inside `map`.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
-		// TODO Auto-generated method stub
-		return null;
+		final List<Node> valKeys = getVarKeys(sm);
+		final List<SolutionMapping> joinPartner = new ArrayList<>();
+		List<SolutionMapping> matchingSolMaps = new ArrayList<>();
+
+		if (!valKeys.isEmpty() && valKeys.size() == joinVariables.size()){
+			matchingSolMaps = map.get(valKeys);
+		}
+		else {
```
hartig, line 87: Instead of `else`, can you do `else if` here, with the condition that `valKeys` is empty.  And then add an `else` branch that throws an `IllegalStateException`.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+			final Iterator<List<SolutionMapping>> li = map.values().iterator();
```
hartig, line 88:
```suggestion
			final Iterator<SolutionMapping> li = iterator();
```

hartig: As per my suggested change, I think you can simply reuse the `iterator` method of this class.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+			while(li.hasNext()){
+				matchingSolMaps.addAll(li.next());
+			}
+		}
+
+		if (matchingSolMaps != null) {
```
hartig, line 94: This check is not needed if you change the previous if-else block as per my comment above.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
 
 	@Override
 	public Iterable<SolutionMapping> getJoinPartners( final SolutionMapping sm )
 			throws UnsupportedOperationException
 	{
-		// TODO Auto-generated method stub
-		return null;
+		final List<Node> valKeys = getVarKeys(sm);
+		final List<SolutionMapping> joinPartner = new ArrayList<>();
+		List<SolutionMapping> matchingSolMaps = new ArrayList<>();
```
hartig, line 82: Instead of a list, change this to an iterator, which has two performance-related advantages: i) you do not need to create ArrayList object here and ii) in the else-branch below, you do not need to need to materialize the iterator into this ArrayList.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
 	}
 
 	@Override
 	public Iterable<SolutionMapping> getJoinPartners( final SolutionMapping sm )
 			throws UnsupportedOperationException
 	{
-		// TODO Auto-generated method stub
-		return null;
+		final List<Node> valKeys = getVarKeys(sm);
+		final List<SolutionMapping> joinPartner = new ArrayList<>();
```
hartig, line 81: Move this line down to the loop that populates this list.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
-		// TODO Auto-generated method stub
-		return null;
+		if ( joinVariables.size() == 1 && joinVariables.contains(var) ){
+			final List<Node> valKeyL = new ArrayList<>();
+			valKeyL.add(value);
+			final List<SolutionMapping> solMaps = map.get(valKeyL);
+			return solMaps;
+		}
+		else{
+			throw new IllegalArgumentException();
```
hartig, line 115: An `UnsupportedOperationException` is better here.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
 	public Iterable<SolutionMapping> findSolutionMappings( final Var var, final Node value )
 			throws UnsupportedOperationException
 	{
-		// TODO Auto-generated method stub
-		return null;
+		if ( joinVariables.size() == 1 && joinVariables.contains(var) ){
+			final List<Node> valKeyL = new ArrayList<>();
+			valKeyL.add(value);
+			final List<SolutionMapping> solMaps = map.get(valKeyL);
+			return solMaps;
```
hartig, line 112:
```suggestion
			return map.get(valKeyL);
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
-		return null;
+		if ( joinVariables.size() == 2 && joinVariables.contains(var1) && joinVariables.contains(var2) ){
+			final List<Node> valKeyL = new ArrayList<>();
+			valKeyL.add(value1);
+			valKeyL.add(value2);
+			final List<SolutionMapping> solMaps = map.get(valKeyL);
+			return solMaps;
+		}
+		else{
+			throw new IllegalArgumentException();
```
hartig, line 133:
```suggestion
			throw new UnsupportedOperationException();
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
hartig:
```suggestion
			throw new UnsupportedOperationException();
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+		if( joinVariables.size() == 3 && joinVariables.contains(var1) && joinVariables.contains(var2) && joinVariables.contains(var3) ){
+			final List<Node> valKeyL = new ArrayList<>();
+			valKeyL.add(value1);
+			valKeyL.add(value2);
+			valKeyL.add(value3);
+			final List<SolutionMapping> solMaps = map.get(valKeyL);
+			return solMaps;
+		}
+		else{
+			throw new IllegalArgumentException();
```
hartig, line 153:
```suggestion
			throw new UnsupportedOperationException();
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
 			final Var var2, final Node value2 )
-					throws UnsupportedOperationException
+			throws UnsupportedOperationException
 	{
-		// TODO Auto-generated method stub
-		return null;
+		if ( joinVariables.size() == 2 && joinVariables.contains(var1) && joinVariables.contains(var2) ){
+			final List<Node> valKeyL = new ArrayList<>();
+			valKeyL.add(value1);
+			valKeyL.add(value2);
```
hartig, line 128: This may be incorrect. You cannot assume that `var1` and `var2` are given in the order in which these two variables appear in `joinVariables`. Hence, you have to check whether `var1` or `var2` is the first variable in `joinVariables` and, depending on this check, add `value1` first or add `value2` first.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+			final List<SolutionMapping> solMaps = map.get(valKeyL);
+			return solMaps;
```
hartig, line 130:
```suggestion
			return  map.get(valKeyL);
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
-					throws UnsupportedOperationException
+			throws UnsupportedOperationException
 	{
-		// TODO Auto-generated method stub
-		return null;
+		if( joinVariables.size() == 3 && joinVariables.contains(var1) && joinVariables.contains(var2) && joinVariables.contains(var3) ){
+			final List<Node> valKeyL = new ArrayList<>();
+			valKeyL.add(value1);
+			valKeyL.add(value2);
+			valKeyL.add(value3);
```
hartig, line 148: Same issue here.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+			final List<SolutionMapping> solMaps = map.get(valKeyL);
+			return solMaps;
```
hartig, line 150:
```suggestion
			return map.get(valKeyL);
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+		}
+		else{
+			throw new IllegalArgumentException();
+		}
 	}
 
+	public List<Node> getVarKeys(final SolutionMapping e){
```
hartig, line 157:
```suggestion
	protected List<Node> getVarKeys(final SolutionMapping e){
```

hartig: In addition to the suggested change, I suggest to rename this method to `getVarKey` because it produces only one index key.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+		final Binding solMapBinding = e.asJenaBinding();
+		final List<Node> valKeys = new ArrayList<>();
+
+		for (Var v : joinVariables) {
```
hartig, line 161:
```suggestion
		for ( final Var v : joinVariables ) {
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,171 @@
...
+			if ( !solMapBinding.contains(v) ){
+				throw new IllegalArgumentException();
+			}
+			else {
+				valKeys.add(solMapBinding.get(v));
+			}
```
hartig, line 167:
```suggestion
			final Node n = solMapBinding.get(v);
			if ( n == null ){
				throw new IllegalArgumentException();
			}
			valKeys.add(n);
```

hartig: Same idea as somewhere above; my suggested change requires only one lookup whereas your code requires two.

hartig (changes requested):

hartig:
A new code review is ready for you  ;-)

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
chengsijin0817: Updated. Pass a Set<Var> to this constructor.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,169 @@
...
 	}
 
 	@Override
 	public Iterable<SolutionMapping> getJoinPartners( final SolutionMapping sm )
 			throws UnsupportedOperationException
 	{
-		// TODO Auto-generated method stub
-		return null;
+		final List<Node> valKeys = getVarKeys(sm);
+		Iterator<SolutionMapping> matchingSolMaps;
```
hartig, line 83:
```suggestion
		final Iterator<SolutionMapping> matchingSolMaps;
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,169 @@
...
+		if (!valKeys.isEmpty() && valKeys.size() == joinVariables.size()){
+			matchingSolMaps = map.containsKey(valKeys)? map.get(valKeys).iterator() : null;
```
hartig, line 85:
```suggestion
			final List<SolutionMapping> l = map.get(valKeys);
			matchingSolMaps = (l != null) ? l.iterator() : null;
```

hartig: Again, my suggested change reduces the code from two lookups to only one.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/SolutionMappingsHashTableTest.java
```diff
@@ -0,0 +1,495 @@
...
+                var3, z2) );
+        input1.add( SolutionMappingUtils.createSolutionMapping(
+                var2, y2,
+                var3, z3) );
+
+        final Set<Var> inputVars = new HashSet<>();
+        inputVars.add(var2);
+
+        // Build SolutionMappingsHashTable for input1
+        final SolutionMappingsHashTable solMHashTable = new SolutionMappingsHashTable(inputVars);
```
hartig, line 48: For convenience, I propose to add another constructor to SolutionMappingsHashTable:

```
public SolutionMappingsHashTable( final Var ... vars ) {
```

Then, you can simplify the code here by using this constructor as follows:

```
final SolutionMappingsHashTable solMHashTable = new SolutionMappingsHashTable(var2);
```

Hence, there is no need here to create the set called inputVars

hartig: The same simplification can probably be used in the other unit tests as well.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/SolutionMappingsHashTableTest.java
```diff
@@ -0,0 +1,495 @@
...
+        final Node x1 = NodeFactory.createURI("http://example.org/x1");
+        final Node x2 = NodeFactory.createURI("http://example.org/x2");
+        final Node y1 = NodeFactory.createURI("http://example.org/y1");
+        final Node y2 = NodeFactory.createURI("http://example.org/y2");
+        final Node y3 = NodeFactory.createURI("http://example.org/y3");
+        final Node z1 = NodeFactory.createURI("http://example.org/z1");
+        final Node z2 = NodeFactory.createURI("http://example.org/z2");
+        final Node z3 = NodeFactory.createURI("http://example.org/z3");
+
+        final GenericIntermediateResultBlockImpl input1 = new GenericIntermediateResultBlockImpl();
```
hartig, line 33: The code here can be simplified. There is no need for such an input1 object. Instead, simply create the solution mappings directly when calling solMHashTable.add(...) below.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/SolutionMappingsHashTableTest.java
```diff
@@ -0,0 +1,495 @@
...
+
+        // Build SolutionMappingsHashTable for input1
+        final SolutionMappingsHashTable solMHashTable = new SolutionMappingsHashTable(inputVars);
+        for ( final SolutionMapping sm : input1.getSolutionMappings() ) {
+            solMHashTable.add(sm);
+        }
+
+        //------------------
+        // Checking SolutionMappingsHashTable
+        assertEquals( 2, solMHashTable.size() );
```
hartig, line 55: This should be 3.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,209 @@
...
+			size = size + li.next().size();
+		}
+		return size;
 	}
 
 	@Override
 	public boolean isEmpty() {
-		// TODO Auto-generated method stub
-		return false;
+		return ! ( !map.isEmpty() && size()>0 );
```
hartig, line 40:
```suggestion
		for ( final List<SolutionMapping> l : map.values() ) {
			if ( ! l.isEmpty() ) {
				return false;
			}
		}
		return true;
```

hartig: See my suggested change above. It is more efficient than your implementation. More precisely, for cases in which `map.isEmpty()` is `true`, they are equal. However, if `map.isEmpty()` is false, my suggested change does not need to iterate over all values of the `map`; instead, as soon as one value is found that is nonempty, the iteration can stop.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,209 @@
...
 	}
 
 	@Override
 	public boolean contains( final Object o ) {
-		// TODO Auto-generated method stub
-		return false;
+		if( !(o instanceof SolutionMapping) ){
+			throw new IllegalArgumentException();
```
hartig, line 46:
```suggestion
			return false;
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,209 @@
...
+		boolean contain = false;
+		if( !map.isEmpty() ){
+			final Iterator<List<SolutionMapping>> li = map.values().iterator();
+			while( li.hasNext() ){
+				if( li.next().contains((SolutionMapping) o) ){
+					contain = true;
+				}
+			}
+		}
+		return contain;
```
hartig, line 58: This can be simplified in several ways.

First, there is no need for the `contain` variable. Instead, simply return `true` inside the loop as soon as the solution mapping is found in a list. No need to continue looking into the rest of the lists!

Second, there is no need for the `if` statement. If `map` is empty, then `map.values()` will be empty and, thus, the loop will not even be entered.

Third, instead of using the iterator, you can use a `for` loop.

Fourth, casting `o` to a `SolutionMapping` should be done before the loop. Otherwise, it is done over and over again.

Hence, overall, you can replace this whole part as follows.
```suggestion
		final SolutionMapping sm = (SolutionMapping) o;
		for ( final List<SolutionMapping> li : map.values() ) {
			if ( l.contains(sm) ) {
				return true;
			}
		}
		return false;
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,209 @@
...
 	}
 
 	@Override
 	public void clear() {
-		// TODO Auto-generated method stub
-		
+		final Iterator<List<SolutionMapping>> liValue = map.values().iterator();
+		while( liValue.hasNext() ){
+			liValue.next().clear();
+		}
```
hartig, line 95:
```suggestion
		for ( final List<SolutionMapping> l : map.values() ) {
			l.clear();
		}
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,209 @@
...
+				if( v.equals(var1) ){
+					valKeyL.add(value1);
+				}
+				else if( v.equals(var2) ){
+					valKeyL.add(value2);
+				}
+				else{
+					throw new IllegalArgumentException();
+				}
+			}
```
hartig, line 158: Thanks for fixing this issue. Can you please add a unit test that uses this method and that would fail if this fix was not done. I mean, I see that you have a unit test that uses this method but in that unit test the order of the variables `var1` and `var2` is the same when creating the hash table and when probing into the hash table with this method here. Now, it would be good to have another version of this unit test where the order is swapped when probing.

hartig (changes requested):
Here are more comments on your recent changes.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
chengsijin0817: I added a unit test for testing this method. (commit ff86aa7)

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,210 @@
...
 	{
-		// TODO Auto-generated method stub
-		return null;
+		final List<Node> valKeys = getVarKeys(sm);
+		final Iterator<SolutionMapping> matchingSolMaps;
+		if (!valKeys.isEmpty() && valKeys.size() == joinVariables.size()){
+			final List<SolutionMapping> l = map.get(valKeys);
+			matchingSolMaps = (l != null) ? l.iterator() : null;
+		}
+		else if(valKeys.isEmpty()){
```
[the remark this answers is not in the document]
hartig: Also, instead of returning an empty valKeys list, getVarKeys should better return null (in cases in which the given solution mapping doesn't have any of the join variables).

Then, after fixing getVarKeys and making it return null in the aforementioned case, you will have to swap the order of the two if-branches here.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -1,85 +1,210 @@
...
+			matchingSolMaps = iterator();
+		}
+		else {
+			throw new IllegalStateException();
```
hartig, line 114:
```suggestion
			throw new  UnsupportedOperationException();
```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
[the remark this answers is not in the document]
hartig: No, building a hash table without any variable(s) to build it on does not make any sense ;-)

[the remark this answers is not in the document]
chengsijin0817: I updated the method getVarKeys(). It returns valKeys only when the given solution mapping contains complete join variables. If the solution mapping doesn't contain complete join variables or does not have any of the join variables, this method return null. (b527f7a)

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTable.java
```diff
@@ -19,7 +19,12 @@
 	protected final Collection<Var> joinVariables;
 
 	public SolutionMappingsHashTable( final Set<Var> joinVariables ){
-		this.joinVariables = joinVariables;
+		if( joinVariables.isEmpty() ){
+			throw new UnsupportedOperationException();
```
hartig, line 23:
```suggestion
			throw new IllegalArgumentException();
```

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/SolutionMappingsHashTableTest.java
```diff
@@ -0,0 +1,297 @@
...
+
+        assertFalse( it5.hasNext() );
+
+    }
+
+    @Test
+    public void hashTableWithEmptyInputVariable() {
+
+        final Set<Var> inputVars = new HashSet<>();
+        assertThrows(UnsupportedOperationException.class,
```
hartig, line 294:
```suggestion
        assertThrows(IllegalArgumentException.class,
```

hartig (approved):

hartig:
This looks good now!

PR #41, opened 2021-05-13
Filter bind join

scferrada:
There might be room for optimisation and improvement, especially when creating the filter operation.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
@@ -54,7 +54,7 @@ public class ExecOpBindJoinSPARQLwithFILTER extends ExecOpGenericBindJoin<Triple
 			Binding b = SolutionMappingUtils.restrict(s, varsInTP).asJenaBinding();
 			Expr conjunction = null;
 			boolean firstConj = true;
-			Iterator<Var> vars = b.vars(); 
+			final Iterator<Var> vars = b.vars(); 
 			while (vars.hasNext()) {
 				final Var v = vars.next();
 				final Node uri = b.get(v);
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
@@ -48,9 +48,10 @@ public class ExecOpBindJoinSPARQLwithFILTER extends ExecOpGenericBindJoin<Triple
 	}
 
 	protected Op createFilter( final Iterable<SolutionMapping> solMaps ) {
+		if (varsInTP.isEmpty()) return new OpTriple(query.asJenaTriple());
 		Expr disjunction = null;
 		for (final SolutionMapping s : solMaps) {
-			final Binding b = SolutionMappingUtils.restrict(s, varsInTP).asJenaBinding();
+			final Binding b = SolutionMappingUtils.restrict(s.asJenaBinding(), varsInTP);
 			Expr conjunction = null;
 			final Iterator<Var> vars = b.vars(); 
 			while (vars.hasNext()) {
@@ -71,7 +72,6 @@ public class ExecOpBindJoinSPARQLwithFILTER extends ExecOpGenericBindJoin<Triple
 				disjunction = new E_LogicalOr(disjunction, conjunction);
 			}
 		}
-		if (disjunction == null) return new OpTriple(query.asJenaTriple());
 		return OpFilter.filter(disjunction, new OpTriple(query.asJenaTriple()));
 	}
 
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
@@ -1,5 +1,6 @@
 package se.liu.ida.hefquin.engine.queryplan.executable.impl.ops;
 
+import java.util.Iterator;
 import java.util.Set;
 
 import org.apache.jena.graph.Node;
@@ -12,7 +13,6 @@ import org.apache.jena.sparql.expr.E_Equals;
 import org.apache.jena.sparql.expr.E_LogicalAnd;
 import org.apache.jena.sparql.expr.E_LogicalOr;
 import org.apache.jena.sparql.expr.Expr;
-import org.apache.jena.sparql.expr.ExprList;
 import org.apache.jena.sparql.expr.ExprVar;
 import org.apache.jena.sparql.expr.nodevalue.NodeValueNode;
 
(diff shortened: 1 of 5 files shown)


scferrada:
Please ignore it a the moment, I had made a mistake.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+import se.liu.ida.hefquin.engine.query.TriplePattern;
+import se.liu.ida.hefquin.engine.query.impl.QueryPatternUtils;
+import se.liu.ida.hefquin.engine.query.impl.SPARQLGraphPatternImpl;
+import se.liu.ida.hefquin.engine.queryproc.ExecutionContext;
+
+public class ExecOpBindJoinSPARQLwithFILTER extends ExecOpGenericBindJoin<TriplePattern,SPARQLEndpoint> {
+	
+	protected final Set<Var> varsInTP;
+
+	public ExecOpBindJoinSPARQLwithFILTER(TriplePattern query, SPARQLEndpoint fm) {
```
hartig, line 35:
```suggestion
	public ExecOpBindJoinSPARQLwithFILTER( final TriplePattern query, final SPARQLEndpoint fm ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+		super(query, fm);
+		varsInTP = QueryPatternUtils.getVariablesInPattern(query);
+	}
+
+	@Override
+	protected Iterable<? extends SolutionMapping> fetchSolutionMappings(Iterable<SolutionMapping> solMaps,
```
hartig, line 41:
```suggestion
	protected Iterable<? extends SolutionMapping> fetchSolutionMappings( final Iterable<SolutionMapping> solMaps,
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+			ExecutionContext execCxt) {
```
hartig, line 42:
```suggestion
			final ExecutionContext execCxt) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+		final Op op = createFilter(solMaps);
+		final SPARQLGraphPattern pattern = new SPARQLGraphPatternImpl(op);
+		final SPARQLRequest request = new SPARQLRequestImpl(pattern);
+		final SolMapsResponse response = execCxt.getFederationAccessMgr().performRequest(request, fm);
+		return response.getSolutionMappings();
+	}
+
+	private Op createFilter(Iterable<SolutionMapping> solMaps) {
```
hartig, line 50:
```suggestion
	protected Op createFilter( final Iterable<SolutionMapping> solMaps ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+		Expr disjunction = null;
+		boolean firstDisj = true;
+		for (final SolutionMapping s : solMaps) {
+			Binding b = SolutionMappingUtils.restrict(s, varsInTP).asJenaBinding();
```
hartig, line 54: This can be done more efficiently: the restrict method that you use here creates a SolutionMapping object, which is not actually needed here. Therefore, please add another restrict method to SolutionMappingUtils that takes a Binding as argument (rather than a SolutionMapping) and that returns a Binding. Then, use that method here. Additionally, you can then use that method also in the implementation of the existing restrict method.

hartig, line 54:
```suggestion
			final Binding b = SolutionMappingUtils.restrict(s, varsInTP).asJenaBinding();
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+			Expr conjunction = null;
+			boolean firstConj = true;
+			Iterator<Var> vars = b.vars();
```
hartig, line 57:
```suggestion
			final Iterator<Var> vars = b.vars(); 
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+		for (final SolutionMapping s : solMaps) {
+			Binding b = SolutionMappingUtils.restrict(s, varsInTP).asJenaBinding();
+			Expr conjunction = null;
+			boolean firstConj = true;
+			Iterator<Var> vars = b.vars(); 
+			while (vars.hasNext()) {
+				final Var v = vars.next();
+				final Node uri = b.get(v);
+				final Expr expr = new E_Equals(new ExprVar(v), new NodeValueNode(uri));
+				if (firstConj) {
```
hartig, line 62: There's no need for this additional program variable. Just check whether 'conjunction' is null ;-)

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+					conjunction = expr;
+					firstConj = false;
+				} else {
+					conjunction = new E_LogicalAnd(conjunction, expr);
+				}
+			}
+			if (conjunction == null)
+				continue;
+			if (firstDisj) {
```
hartig, line 71: Same here. There's no need for this additional program variable. Just check whether 'disjunction' is null.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
```diff
@@ -0,0 +1,82 @@
...
+				disjunction = conjunction;
+				firstDisj = false;
+			} else {
+				disjunction = new E_LogicalOr(disjunction, conjunction);
+			}
+		}
+		if (disjunction == null) return new OpTriple(query.asJenaTriple());
```
hartig, line 78: One of the cases in which this may happen is if triple pattern of this operator has no variables (i.e.,  varsInTP is empty). This is a special case which, for performance reasons, should  be handled separately already at the beginning of this method.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithFILTER.java
```diff
@@ -0,0 +1,32 @@
...
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpTPAdd;
+
+public class PhysicalOpBindJoinWithFILTER extends BasePhysicalOpSingleInputJoin {
+
+	protected PhysicalOpBindJoinWithFILTER(UnaryLogicalOp lop) {
+		super(lop);
+	}
+
+	@Override
+	public UnaryExecutableOp createExecOp(ExpectedVariables... inputVars) {
```
hartig, line 18:
```suggestion
	public UnaryExecutableOp createExecOp( final ExpectedVariables... inputVars ) {
```

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithFILTER.java
scferrada: Yes, but the other case is when `solMaps` is empty, and it is not possible to check an Iterable's size. I added the empty `varsInTP` condition at the beginning, but kept the `disjunction == null` check at the end.

hartig (approved):
Great. Thanks for fixing the minor issues mentioned in my previous code review!

PR #43, opened 2021-05-18
Hash table based on two variables

chengsijin0817:


src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.util.*;
+
+public class SolutionMappingsHashTableBasedOnTwoVars extends SolutionMappingsHashTable{
+    protected final Map<Node, List<Node>> mapNN = new HashMap<>();
+    protected final Map<Node, List<SolutionMapping>> mapNS = new HashMap<>();
+    protected final Var joinVar1;
+    protected final Var joinVar2;
+
+    public SolutionMappingsHashTableBasedOnTwoVars( final Var joinVar1, final Var joinVar2 ) {
```
hartig, line 17:
```suggestion
    public SolutionMappingsHashTableBasedOnTwoVars( final Var joinVar1, final Var joinVar2 ) {
        assert joinVar1 != null;
        assert joinVar2 != null;
        assert ! joinVar1.equals(joinVar2);

```

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    public boolean add( final SolutionMapping e ) {
+        final Binding solMapBinding = e.asJenaBinding();
+        final Node n1 = solMapBinding.get(joinVar1);
+        final Node n2 = solMapBinding.get(joinVar2);
+
+        if (n1 == null || n2 == null){
+            throw new IllegalArgumentException();
+        }
+        else{
```
hartig, line 31: No need to put this into an else-block. Removing the else-block will make the code a bit more readable.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
```diff
@@ -0,0 +1,118 @@
...
+            List<Node> varList = mapNN.get(n1);
+            if( varList == null){
+                varList = new ArrayList<>();
+                mapNN.put(n1, varList);
+            }
+            varList.add(n2);
```
hartig, line 37: `n2` may already be contained in `varList`. In this case, there is no need to insert it twice.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.Binding;
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.data.impl.SolutionMappingUtils;
+
+import java.util.*;
+
+public class SolutionMappingsHashTableBasedOnTwoVars extends SolutionMappingsHashTable{
+    protected final Map<Node, List<Node>> mapNN = new HashMap<>();
+    protected final Map<Node, List<SolutionMapping>> mapNS = new HashMap<>();
```
hartig, line 13: I don't think that this is going to work. What would be the keys in `mapNS`? These keys should not be independent of the corresponding value for the other join variable. For instance, assume we first insert a solution mapping with var1=X and var2=Z, and after that we insert a solution mapping with var1=Y and var2=Z. Hence, both of these solution mappings have the same value for the second join variable. Then, when searching for join partners with var1=X and var2=Z, there should be an efficient way of retrieving only the relevant candidates with var1=X and var2=Z.

To fix this, the internal data structure here should be a nested map:
```
protected final Map<Node, Map<Node, List<SolutionMapping>>> map = new HashMap<>();
```
Then, the keys in the outer map (i.e., in `map`) are the RDF terms for the first join variable and the corresponding values would be further maps. The keys in each of these inner maps are the RDF terms for the second join variable, but only the ones that belong to the corresponding binding for the first variable.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
```diff
@@ -0,0 +1,118 @@
...
+    public Iterable<SolutionMapping> getJoinPartners( final SolutionMapping sm )
+            throws UnsupportedOperationException
+    {
+        final Binding solMapBinding = sm.asJenaBinding();
+        final Node n1 = solMapBinding.get(joinVar1);
+        final Node n2 = solMapBinding.get(joinVar2);
+        Iterator<SolutionMapping> matchingSolMaps;
+
+        if ( n1 == null || n2 == null ){
+            matchingSolMaps = iterator();
```
hartig, line 59: This should be done only if `n1 == null`.

For the case that `n2 == null` but `n1 != null`, you can retrieve the inner map for `n1` and use an iterator over all elements in there, which reduces the number of candidates to be considered for the compatibility checks below.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
```diff
@@ -0,0 +1,118 @@
...
+
+import org.apache.jena.graph.Node;
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.Binding;
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.data.impl.SolutionMappingUtils;
+
+import java.util.*;
+
+public class SolutionMappingsHashTableBasedOnTwoVars extends SolutionMappingsHashTable{
```
hartig, line 11: No. This class should be a subclass of `SolutionMappingsIndexBase`.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
```diff
@@ -0,0 +1,118 @@
...
+                joinPartner.add(matchSolM);
+            }
+        }
+        return joinPartner;
+    }
+
+    @Override
+    public Iterable<SolutionMapping> findSolutionMappings(
+            final Var var1, final Node value1,
+            final Var var2, final Node value2 )
```
hartig, line 79: Where are the corresponding methods for one variable and for three variables?

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
chengsijin0817: It makes sense. I have updated the data structure to a nested map.

src/main/java/se/liu/ida/hefquin/engine/datastructures/impl/SolutionMappingsHashTableBasedOnTwoVars.java
chengsijin0817: Added.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpSymmetricHashJoin.java
```diff
@@ -1,22 +1,24 @@
...
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 
 public class ExecOpSymmetricHashJoin implements BinaryExecutableOp{
 
-    protected final SolutionMappingsHashTable solMHashTableL;
-    protected final SolutionMappingsHashTable solMHashTableR;
+    protected final SolutionMappingsIndexBase solMHashTableL;
+    protected final SolutionMappingsIndexBase solMHashTableR;
```
hartig, line 21: Is it really necessary that these two are of type  SolutionMappingsIndexBase? Can't they just be of type SolutionMappingsIndex?

hartig (commented):
Just one question.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpSymmetricHashJoin.java
chengsijin0817: Yes, we can declare these two as type SolutionMappingsIndex.
Currently, there is no other implementation besides SolutionMappingsIndexBase, and no distinct method is contained in SolutionMappingsIndexBase. Exactly, it is not necessary to declare them as SolutionMappingsIndexBase.

hartig (approved):
Looks all good now. Thanks.

Issue #34, opened 2021-05-04
Implementation of UNION-based bind join at SPARQL endpoint


hartig:
Yet another alternative to the approach in #32 and in #33 is to inject the given input solution mappings into the SPARQL query by means of UNIONS with FILTER conditions. Consider the same example as in #32 (a tpAdd operator with the triple pattern (?x, foaf:knows, ?y) and two solution mappings coming from the child of the tpAdd operator: m1={ ?x -> ex:alice } and m2={ ?x -> ex:bob }). Then, the UNION-based bind join would be based on sending the following SPARQL query as a request to the SPARQL endpoint (prefix declarations omitted).
```
SELECT ?x ?y WHERE {
  { ?x foaf:knows ?y .  FILTER(?x=ex:alice) }
  UNION
  { ?x foaf:knows ?y .  FILTER(?x=ex:bob) }
}
```
The purpose of this issue here is to add an implementation of this approach to our engine, which includes:
* an executable operator called `ExecOpBindJoinSPARQLwithUNION`,
* a set of unit tests for that operator (again, based on `TestsForTPAddAlgorithms`), and
* a corresponding implementation of `UnaryPhysicalOpForLogicalOp`.

Also here, remember [the additional comment that I made for the VALUES-based option](https://github.com/LiUSemWeb/HeFQUIN/issues/32#issuecomment-832208053), which applies in this case as well.

The UNION-based approach is also described in
_Carlos Buil Aranda, Axel Polleres, Jürgen Umbrich: [Strategies for Executing Federated Queries in SPARQL1.1](https://www.carlosbuil.com/2015-iswc-strategies.pdf). ISWC 2014._

hartig:
@scferrada  ...and another one of the same kind as #32 and #33 -- the last one ;-)

PR #44, opened 2021-05-19
Union bind join

scferrada:


src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
```diff
@@ -0,0 +1,80 @@
...
+import se.liu.ida.hefquin.engine.federation.access.impl.req.SPARQLRequestImpl;
+import se.liu.ida.hefquin.engine.query.SPARQLGraphPattern;
+import se.liu.ida.hefquin.engine.query.TriplePattern;
+import se.liu.ida.hefquin.engine.query.impl.QueryPatternUtils;
+import se.liu.ida.hefquin.engine.query.impl.SPARQLGraphPatternImpl;
+import se.liu.ida.hefquin.engine.queryproc.ExecutionContext;
+
+public class ExecOpBindJoinSPARQLwithUNION extends ExecOpGenericBindJoin<TriplePattern,SPARQLEndpoint> {
+
+	private Set<Var> varsInTP;
```
hartig, line 33:
```suggestion
	protected Set<Var> varsInTP;
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
```diff
@@ -0,0 +1,80 @@
...
+		return response.getSolutionMappings();
+	}
+
+	protected Op createUnion(final Iterable<SolutionMapping> solMaps) {
+		Op tp = new OpTriple(query.asJenaTriple());
+		if (varsInTP.isEmpty()) return tp;
+		Op union = null;
+		for ( final SolutionMapping s : solMaps) {
+			final Binding b = SolutionMappingUtils.restrict(s.asJenaBinding(), varsInTP);
+			if (b.size() == 0) continue;
```
hartig, line 56: This is wrong. Every solution mapping `s` that does not have any variables in common with the triple pattern is a join partner for *every* solution mapping that can be produced for the triple pattern. Hence, as soon as the is one such `s`, the triple pattern has to be included in the request without any additional filter. In fact, in this case, there is no need to send a UNION-based request at all; in contrast, the triple pattern can be sent just by itself. Hence, I believe that the correct behavior here is the following.
```suggestion
			// If the current solution mapping does not have any variables in common with
			// the triple pattern of this operator, then every matching triple is a join partner
			// for the current solution mapping. Hence, in this case, we may simply retrieve
			// all matching triples (i.e., no need for putting together the UNION pattern).
			if (b.size() == 0) return tp;
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
```diff
@@ -0,0 +1,80 @@
...
+			Expr conjunction = null;
+			final Iterator<Var> vars = b.vars(); 
+			while (vars.hasNext()) {
+				final Var v = vars.next();
+				final Node uri = b.get(v);
+				final Expr expr = new E_Equals(new ExprVar(v), new NodeValueNode(uri));
```
hartig, line 62: There are cases in which this may result in incorrect intermediate results produced by this operator (more precisely, intermediate results that contain spurious duplicates).

The issue is twofold:
1. The call of `restrict` above may give us duplicates. Then, for each such duplicate, you add an identical copy of `conjunction` to the UNION pattern. Now, for every solution mapping that is a solution for this `conjunction`, the SPARQL endpoint returns as many duplicates as there are copies of this `conjunction` in the UNION pattern. Joining these duplicates with the given solution mappings in `solMaps` (which happens in the superclass of this class here), will result incorrect duplicates coming out of this operator. The fix for this issue is to avoid adding multiple identical copies of the same `conjunction` to the UNION pattern.
2. There are cases  in which things are even more complicated. Unfortunately! As an example, consider `varsInTP` contains two variables, say ?v1 and ?v2, and there are two input solution mappings s1 and s2 such that s1 has a value for both ?v1 and ?v2, respectively, whereas s2 has a value for ?v1 but not for ?v2.  Also assume that the value for ?v1 is the same in both solution mappings; i.e., s1(?v1)=s2(?v1). Hence, we may have something like the following: s1(?v1)=s2(?v1)=uri1, s1(?v2)=uri2, and s2(?v2) is undefined. Then, the resulting UNION pattern might be as follows:
```
{
   # this subpattern is from s1
   ?v1 :p ?v2 . FILTER( ?v1=uri1 && ?v2=uri2 )
}
UNION
{
   # this subpattern is from s2
   ?v1 :p ?v2 . FILTER( ?v1=uri1 )
}
```
Note that every triple that every solution for the first subpattern is also a solution for the second subpattern. Hence, we may introduce spurious duplicates when join the result of this query with the two input solution mappings.

Do you have a good idea to avoid these issues? ("good" in the sense of i) easy to implement and ii) not too problematic from a performance perspective)

If not, I suggest we ignore the UNION-based bind join. That is, I will copy the explanation from above into #34, mark #34 as "won't-fix", and we delete this PR.

hartig: Another option is to implement a restricted version of this operator which will explicitly be only for cases where such issues are guaranteed not to occur (which can be checked based on the `ExpectedVariables` of the input operator, together with the variables of the given triple pattern). Perhaps that's a good compromise!?

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
scferrada: On the meantime I have committed this change. However, I ask if the fact that one of the restricted solution mappings has size 0 implies that every solution mapping will have a size 0. I believe the implication is not correct, thus I would have to add `tp` to the union when the binding size is 0 instead of returning just the triple pattern.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
scferrada: For the first issue, we can maintain then a set of `conjunction` before creating the UNION Op object.

Given such a set, the second issue might be addressed by simply comparing all pairs of filters, but as you pointed out it would be very costly.

As for checking the `ExpectedVariables` I believe we would incur in a similar problem (having to check every pair of possibilities), unless we check that the variables in the mappings and triple pattern are only certain variables (but in that case, the mappings are not bound to use all of the certain variables, so that might be the check?).

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
hartig: In general we cannot assume that implication, I believe. On the other hand, there are cases where the implication holds. However, in these cases, this can be figured out already before the query execution process; namely, based on the ExpectedVariables of the input operator together with the variables of the given triple pattern. 

Having said that, no matter what, it will always be correct to simply return the triple pattern in all of these cases. The reason is that you use this method here as a basis for the fetchSolutionMappings method above, and that method is just meant to fetch all solution mappings from the SPARQL endpoint that are compatible with at least one of the given solution mappings. Sending only the triple pattern achieves that. The actual join happens then in the superclass after fetchSolutionMappings is called.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
hartig: The idea with checking the ExpectedVariables is to check that the set of *possible* variables of the input operator is empty. If that's the case, then every solution mapping from the input operator is guaranteed to have the same variables and, thus, the weird special cases are ruled out. Right? (please double check my reasoning here) 

Of course, it would still be necessary to avoid duplicates as per the first of the two points in my comment above.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
scferrada: alright!

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
hartig: The `ExpectedVariables` do not really "restrict[] the solution mappings fed to a tpAdd operator." Instead, the `ExpectedVariables` (of the operator that is the child of the tpAdd operator) give us information about these solution mappings. That's the main purpose of this data structure. It tells us what is expected to come out of an operator and, thus, it enables us to know what is expected to come in to the parent operator (which is a tpAdd operator in this case).

I was also thinking about the other two types of bind join algorithms for SPARQL endpoints.

For the FILTER-based version, I think there is no actual problem regarding correctness. The duplicates in the FILTER are not nice in terms of unnecessary data shipping from the engine to the SPARQL endpoint but, at least, they do not cause spurious duplicates in the query result (every triple that matches the triple pattern can satisfy the filter condition only once, even if it may satisfy multiple disjuncts of the condition).

For the VALUES-based version, unfortunately, you are right. There, spurious duplicates may occur.

scferrada: I'm seeing that we never call the `createExecOp` function in the tests, thus the check of the expected variables will never occur, thus I will have no information on how the check will affect the test queries. 

Maybe the method `createExecOpForTest` should create a new `PhysicalOpBindJoinWithVALUES` object and ask it to `createExecOp`. Would that be okay?

hartig: Good observation! Yes, do it as you propose.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNIONTest.java
```diff
@@ -0,0 +1,75 @@
...
+	@Override
+	protected SPARQLEndpoint createFedMemberForTest( final Graph dataForMember) {
+		return new SPARQLEndpointForTest(dataForMember);
+	}
+
+	@Override
+	protected UnaryExecutableOp createExecOpForTest( final TriplePattern tp, final SPARQLEndpoint fm) {
+		final LogicalOpTPAdd tpadd = new LogicalOpTPAdd(fm, tp);
+		final PhysicalOpBindJoinWithUNION bindjoin = new PhysicalOpBindJoinWithUNION(tpadd); 
+		return bindjoin.createExecOp(bindjoin.getExpectedVariables(new ExpectedVariables() {
```
scferrada, line 62: Here I create a new `PhysicalOpBindJoinWithUNION` object in order to call `createExecOp`, which receives a list of `ExpectedVariables`. To do that I call the `getExpectedVariables` method of the logical Op. In turn, this method also receives a list of `ExpectedVariables`, which I'm defining as empty.

For all tests, the possible variables are always empty. Even in the spurious duplicates test that I just created. So I'm inclined to believe that the condition that needs to be checked is another (it can also be that my intuition of the expected variables here is wrong). But I can't figure out how to check it anywhere else than in running time in the `fetchSolutionMappings` method.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithUNION.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class PhysicalOpBindJoinWithUNION extends BasePhysicalOpSingleInputJoin {
+
+	public PhysicalOpBindJoinWithUNION( final UnaryLogicalOp lop) {
+		super(lop);
+	}
+
+	@Override
+	public UnaryExecutableOp createExecOp(final ExpectedVariables... inputVars) {
+		for (final ExpectedVariables ev : inputVars) {
```
hartig, line 19: You do not need to iterate here. The size of `inputVars` should be one. In fact, you should add an assert-check for that:
```
assert inputVars.length == 1;
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithUNION.java
```diff
@@ -0,0 +1,38 @@
...
+			System.out.println(ev.getCertainVariables());
```
hartig, line 20: Please remove this.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithUNION.java
```diff
@@ -0,0 +1,38 @@
...
+			if (ev.getPossibleVariables().size() > 0) {
```
hartig, line 21:
```suggestion
			if ( ! ev.getPossibleVariables().isEmpty() ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithUNION.java
```diff
@@ -0,0 +1,38 @@
...
+				throw new IllegalArgumentException("Illegal child operator for tpAdd. Possible variables should be empty.");
```
hartig, line 22:
```suggestion
				// The executable operator for this physical operator (i.e., ExecOpBindJoinSPARQLwithUNION)
				// can work correctly only in cases in which all input solution mappings are for the exact
				// same set of variables. This can be guaranteed only if the set of possible variables from
				// the child operator is empty.
				throw new IllegalArgumentException("Nonempty set of possible variables.");
```

hartig: In contrast to your exception message, it is *not* generally necessary that the child of a tpAdd has an empty set of possible variables. In contrast, that's only necessary if we want to implement the (logical) tpAdd operator by using the ExecOpBindJoinSPARQLwithUNION.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNIONTest.java
hartig: What you do here makes no sense. Usually (i.e., during our query planning process), the `ExpectedVariables` object that is passed to the `createExecOp` method is the `ExpectedVariables` object from the child operator and not from the operator itself (see the class `QueryPlanCompilerImpl` which is where this happens). Now, for unit tests, it should be an `ExpectedVariables` object that reflects the expected variables (certain and possible) of the solution mappings that the unit test will later on send as input to the `ExecOpBindJoinSPARQLwithUNION`.

Generally, the purpose of using `ExpectedVariables` objects during query planning is precisely that certain things do *not* need to be checked anymore at running time. For instance, if variable ?x is in the certain variables in the `ExpectedVariables` object that is passed to a `createExecOp` method, then the executable operator created by that `createExecOp` method can assume that *every* solution mapping that will later be sent to it as input is guaranteed to contain a binding for ?x. Similarly, if variable ?y is in the possible variables in the `ExpectedVariables` object that is passed to a `createExecOp` method, then the executable operator created by that `createExecOp` method can assume that *some* solution mappings that will later be sent to it as input have a binding for ?y; it may then turn out at runtime that every input solution mapping has a binding for ?y, but there is no guarantee for this.

hartig (changes requested):
Why are there all of a sudden so many files in this PR?

Anyways. I have added some more comments in response to some of your new comments.

Additionally, I do not see how/whether you have changed the `createUnion` method such that it addresses the first of the two issues that I mention in https://github.com/LiUSemWeb/HeFQUIN/pull/44/files#r635169740

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
```diff
@@ -0,0 +1,89 @@
...
+				if (conjunction == null){
+					conjunction = expr;
+				} else {
+					conjunction = new E_LogicalAnd(conjunction, expr);
+				}
+			}
+			conjunctions.add(conjunction);
```
hartig, line 75: This works correctly only if the iterator `vars` is guaranteed to return the variables in the same order for every `b`. I am not sure that we can make this assumption. Therefore, to be on the safe side, I propose i) to change the type of `varsInTP` to be a `List` rather than a `Set` and, then, ii) to iterate over `varsInTP` instead of iterating over `b.vars()`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
```diff
@@ -0,0 +1,89 @@
...
+		final SPARQLGraphPattern pattern = new SPARQLGraphPatternImpl(op);
+		final SPARQLRequest request = new SPARQLRequestImpl(pattern);
+		final SolMapsResponse response = execCxt.getFederationAccessMgr().performRequest(request, fm);
+		return response.getSolutionMappings();
+	}
+
+	protected Op createUnion(final Iterable<SolutionMapping> solMaps) {
+		final Op tp = new OpTriple(query.asJenaTriple());
+		if (varsInTP.isEmpty()) return tp;
+		Op union = null;
```
hartig, line 54: Move this one below such that it is directly before the for-loop that uses and assigns this program variable.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/TestsForTPAddAlgorithms.java
```diff
@@ -63,7 +65,20 @@ protected void _tpWithJoinOnObject() {
 		dataForMember.add( Triple.create(y1,p,z2) );
 		dataForMember.add( Triple.create(y2,p,z3) );
 
-		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp);
+		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp, new ExpectedVariables() {
+			@Override
+			public Set<Var> getCertainVariables() {
+				Set<Var> set = new HashSet<>();
```
hartig, line 71:
```suggestion
				final Set<Var> set = new HashSet<>();
```

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/TestsForTPAddAlgorithms.java
```diff
@@ -139,7 +154,20 @@ protected void _tpWithJoinOnSubjectAndObject() {
 		dataForMember.add( Triple.create(x1,y1,z1) );
 		dataForMember.add( Triple.create(x2,y2,z2) );
 
-		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp);
+		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp, new ExpectedVariables() {
+			@Override
+			public Set<Var> getCertainVariables() {
+				Set<Var> set = new HashSet<>();
```
hartig, line 160:
```suggestion
				final Set<Var> set = new HashSet<>();
```

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/TestsForTPAddAlgorithms.java
```diff
@@ -199,7 +227,19 @@ protected void _tpWithoutJoinVariable() {
 		dataForMember.add( Triple.create(y1,p,z1) );
 		dataForMember.add( Triple.create(y2,p,z2) );
 
-		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp);
+		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp, new ExpectedVariables() {
+			@Override
+			public Set<Var> getCertainVariables() {
+				Set<Var> set = new HashSet<>();
```
hartig, line 233:
```suggestion
				final Set<Var> set = new HashSet<>();
```

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/TestsForTPAddAlgorithms.java
```diff
@@ -307,25 +367,77 @@ protected void _tpWithEmptyResponses() {
 
 		final Graph dataForMember = GraphFactory.createGraphMem();
 
-		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp);
+		final Iterator<SolutionMapping> it = runTest(input, dataForMember, tp, new ExpectedVariables() {
+			@Override
+			public Set<Var> getCertainVariables() {
+				Set<Var> set = new HashSet<>();
```
hartig, line 373:
```suggestion
				final Set<Var> set = new HashSet<>();
```

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/TestsForTPAddAlgorithms.java
```diff
@@ -307,25 +367,77 @@ protected void _tpWithEmptyResponses() {
...
+
+		final Graph dataForMember = GraphFactory.createGraphMem();
+		dataForMember.add( Triple.create(s1,p,o1) );
+		dataForMember.add( Triple.create(s2,p,o2) );
+
+		assertThrows(IllegalArgumentException.class,  () -> {
+			runTest(input, dataForMember, tp, new ExpectedVariables() {
+				@Override
+				public Set<Var> getCertainVariables() {
+					Set<Var> set = new HashSet<>();
```
hartig, line 411:
```suggestion
					final Set<Var> set = new HashSet<>();
```

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/TestsForTPAddAlgorithms.java
```diff
@@ -307,25 +367,77 @@ protected void _tpWithEmptyResponses() {
...
+					set.add(var1);
+					return set;
+				}
+
+				@Override
+				public Set<Var> getPossibleVariables() {
+					Set<Var> set = new HashSet<>();
```
hartig, line 418:
```suggestion
					final Set<Var> set = new HashSet<>();
```

hartig (changes requested):
Some more things that need to be changed.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithUNION.java
```diff
@@ -0,0 +1,88 @@
...
+import se.liu.ida.hefquin.engine.federation.access.impl.req.SPARQLRequestImpl;
+import se.liu.ida.hefquin.engine.query.SPARQLGraphPattern;
+import se.liu.ida.hefquin.engine.query.TriplePattern;
+import se.liu.ida.hefquin.engine.query.impl.QueryPatternUtils;
+import se.liu.ida.hefquin.engine.query.impl.SPARQLGraphPatternImpl;
+import se.liu.ida.hefquin.engine.queryproc.ExecutionContext;
+
+public class ExecOpBindJoinSPARQLwithUNION extends ExecOpGenericBindJoin<TriplePattern,SPARQLEndpoint> {
+
+	protected Set<Var> varsInTP;
```
hartig, line 32: Can you please get rid of this one. We don't need to duplicate data in our objects.

To get rid of it, you will have to change the helper function `SolutionMappingUtils.restrict` such that the second parameter is a `Collection<Var>` rather than a `Set<Var>`.

hartig (changes requested):
Now, there is only one small thing left. See my new reivew comment.

hartig (approved):
Perfect! Thanks!!

Issue #27, opened 2021-05-03
Print query plans


hartig:
We need a way to print query plans (logical, physical, executable), mainly for debugging purposes.

hartig:
This issue has been addressed in #47 for logical plans and for physical plans. What remains is to address is for executable plans as well.

PR #47, opened 2021-05-24
first implementation for printing logical plans

scferrada:
Here is a first attempt on printing Logical plans, based in the available Visitor and Walker classes.  I added a second walk method to transverse the tree from top down. 

Since the visitor does not know the level it is at, tabulation currently works as a best guess. This can be improved by coupling the walk method with the visitor class to indicate the current depth, the parent of the operator and the position of the current Op in the child list.

I'm opening the PR to discuss those issues as well as the visitor/walker structure in general.


hartig:
related to #27

hartig (changes requested):
I have added several review comments.

Additionally, I think that the current idea with the indentation is not going to work at all. A better idea would be to extend the `WalkerVisitor` class inside `LogicalPlanWalker` such that it may be initialized with two separate `LogicalPlanVisitor` objects, where one of them is called as a "before visitor" (before going into subplans) and the other as an "after visitor" (after having gone into subplans). Hence, the code of the extended `WalkerVisitor` class may look as follows:
```java
protected static class WalkerVisitor {
	protected final LogicalPlanVisitor beforeVisitor;
	protected final LogicalPlanVisitor afterVisitor;

	public WalkerVisitor( final LogicalPlanVisitor beforeVisitor, final LogicalPlanVisitor afterVisitor ) {
		this.beforeVisitor = beforeVisitor;
		this.afterVisitor = afterVisitor;
	}

	public void walk( final LogicalPlan plan ) {
		if ( beforeVisitor != null ) {
			plan.getRootOperator().visit(beforeVisitor);
		}
		for ( int i = 0; i < plan.numberOfSubPlans(); ++i ) {
			walk( plan.getSubPlan(i) );
		}
		if ( afterVisitor != null ) {
			plan.getRootOperator().visit(afterVisitor);
		}
	}
}
```
Of course, the static method `walk` in `LogicalPlanWalker` has to be extended accordingly.

Once this is in place, there can be a `LogicalPlanPrinter` class that has two implementations of `LogicalPlanVisitor` (one of which as the corresponding "before visitor" and the other one as the "after visitor") such that both access the same `indentLevel` variable of the `LogicalPlanPrinter`. Then, for instance, for the join operator, the "before visitor" prints the info about the join operator and, afterwards, increases the `indentLevel`, and the "after visitor" simply decreases the `indentLevel` again.

hartig (changes requested):
Only one small thing -- see my review comment.

Except for that, it looks good! Good job.

Please only address the last remaining issue in this PR, instead of also adding something for the executable plans here. Once this PR is merged, you can create a separate PR for the printer of the executable plans.

scferrada:
> Please only address the last remaining issue in this PR, instead of also adding something for the executable plans here. Once this PR is merged, you can create a separate PR for the printer of the executable plans.

Given that my other PR was merged I will need to rebase from main, to include the UNION physical operator

hartig:
Oops! I didn't realize there was a conflict between the PRs.

.gitignore
```diff
@@ -25,3 +25,9 @@ hs_err_pid*
 /.classpath
 /.project
 /dependency-reduced-pom.xml
+*.iml
+<<<<<<< HEAD
+/.idea
+=======
+/.idea
+>>>>>>> 6e96eb9d74b6839c61a31117a25ad65797e00edd
```
hartig, line 33: This doesn't look right.

hartig (changes requested):

scferrada:
I added a new commit with a small change and I think now you can merge.

hartig:
Thanks! Merged.

PR #50, opened 2021-05-28
Printing of ExecutablePlans

scferrada:
Here the walker goes through the `ResultElementIterator` recursive structure.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/ExecutablePlan.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/ExecutablePlan.java
@@ -1,8 +1,18 @@
 package se.liu.ida.hefquin.engine.queryplan;
 
+import se.liu.ida.hefquin.engine.queryplan.executable.impl.pullbased.ResultElementIterator;
 import se.liu.ida.hefquin.engine.queryproc.QueryResultSink;
+
+import java.util.NoSuchElementException;
 
 public interface ExecutablePlan
 {
 	void run( QueryResultSink resultSink );
+
+    /**
+     * Returns the root operator of this plan.
+     */
+    ExecutableOperator getRootOperator();
+
+    ResultElementIterator getIterator();
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBinaryUnion.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBinaryUnion.java
@@ -1,6 +1,7 @@
 package se.liu.ida.hefquin.engine.queryplan.executable.impl.ops;
 
 import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.queryplan.executable.ExecutablePlanVisitor;
 import se.liu.ida.hefquin.engine.queryplan.executable.IntermediateResultBlock;
 import se.liu.ida.hefquin.engine.queryplan.executable.IntermediateResultElementSink;
 import se.liu.ida.hefquin.engine.queryproc.ExecutionContext;
@@ -13,6 +14,11 @@ public class ExecOpBinaryUnion implements BinaryExecutableOp{
 		// sequentially (one at a time), an input block size of 1 may
 		// reduce the response time of the overall execution process.
 		return 1;
+	}
+
+	@Override
+	public void visit(final ExecutablePlanVisitor visitor) {
+		visitor.visit(this);
 	}
 
 	@Override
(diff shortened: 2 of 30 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryplan/ExecutablePlan.java
```diff
@@ -1,8 +1,18 @@
...
 public interface ExecutablePlan
 {
 	void run( QueryResultSink resultSink );
+
+    /**
+     * Returns the root operator of this plan.
+     */
+    ExecutableOperator getRootOperator();
+
+    ResultElementIterator getIterator();
```
hartig, line 17: This needs to be moved to the PullBasedPlan class. There will also be push-based plans in the future and these will not be based on iterators but based on queues between the operators.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBinaryUnion.java
```diff
@@ -1,28 +1,34 @@
...
 
 public class ExecOpBinaryUnion implements BinaryExecutableOp{
 
-	@Override
+	@Override
 	public int preferredInputBlockSize() {
-		// Since this algorithm processes the input solution mappings
-		// sequentially (one at a time), an input block size of 1 may
+		// Since this algorithm processes the input solution mappings
+		// sequentially (one at a time), an input block size of 1 may
```
hartig, line 14: What has changed here? It seems nothing.

There are many such lines in this PR that have not actually changed in their content. Perhaps the change is only in the whitespace or the line break character. Please try to avoid such things in the future. That will make my life with the code review easier ;-)

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/ResultBlockIterator.java
```diff
@@ -7,4 +7,5 @@
 public interface ResultBlockIterator extends Iterator<IntermediateResultBlock>
 {
 
+    ResultElementIterator getElementIterator();
```
hartig, line 10: Such a method should not be part of this interface. Not all types of ResultBlockIterator will be based on a ResultElementIterator. For instance, there will be a type of ResultBlockIterator that is based on a temporary index that is built on the fly for the solution mappings produced/retrieved by some operator.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/ResultElementIterWithBinaryExecOp.java
```diff
@@ -27,6 +29,18 @@ public BinaryExecutableOp getOp() {
 		return opRunnerThread.getOp();
 	}
 
+	@Override
+	public int getArity() {
+		return 2;
+	}
+
+	@Override
+	public ResultElementIterator getSubIterator(final int i) throws NoSuchElementException {
```
hartig, line 38: Providing access to the input iterators is a bit dangerous because this opens the possibility that they are accidentally consumed elsewhere even if these iterators are meant to be consumed here.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/ResultElementIterator.java
```diff
@@ -1,10 +1,16 @@
...
 import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.queryplan.ExecutableOperator;
 
 public interface ResultElementIterator extends Iterator<SolutionMapping>
 {
+    ExecutableOperator getOp();
 
+    int getArity();
+
+    ResultElementIterator getSubIterator(final int i) throws NoSuchElementException;
```
hartig, line 15: Same comment here. Exposing the child iterators is dangerous and may lead to bugs.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/pullbased/ResultElementIterator.java
```diff
@@ -1,10 +1,16 @@
...
 
 import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.queryplan.ExecutableOperator;
 
 public interface ResultElementIterator extends Iterator<SolutionMapping>
 {
+    ExecutableOperator getOp();
```
hartig, line 11: Exposing the internal operator here is dangerous. It may lead to bugs if the operator is accidentally used somewhere else.

hartig (changes requested):
Unfortunately, I don't think that the proposed approach is so suitable. The main issue is that it opens options for potential bugs that would never be possible with the current interfaces. Additionally, some things are not going to work for other implementations of the interfaces that I am already planning. See my review comments for the details.

hartig:
We decided to abandon this PR for the moment. The plan is to implement push-based plans which should be better anyways (because they allow for true inter-operator parallelism). Then, the implementation of the pull-based plans will serve more as a fallback for testing and, thus, it may not be necessary to be able to print them.

hartig:
I am closing this PR now. We don't really need this functionality anymore.

Issue #10, opened 2021-05-01
implement SolutionMappingsHashTable


hartig:
This data structure may serve as a basis for implementations of both the hash join algorithm and the symmetric hash join algorithm.

hartig:
While this issue has, in principle, been addressed by PRs #35 #42 and#43, I would still like to clean up the corresponding classes a bit and add some documentation.

hartig:
PR #52 has brought in the additional changes that I still wanted to have before closing this issue here. Now that PR #52 has been merged, we can close the issue.

PR #52, opened 2021-06-03
Improve hash-table implementations of SolutionMappingsIndex

hartig:
This PR brings some restructuring and improvements (functionality, potential performance issues, source code documentation) for the hash-table implementations of `SolutionMappingsIndex`.

The previous implementation was generic in the sense that it included functionality that would be needed only in specific cases (checking for compatibility after doing the actual index look-ups, creating and returning copies of the index buckets to ensure that nothing unexpected happens in cases in which there is a concurrent thread that writes to these buckets while another thread still consumes the result from the index). In cases in which this functionality would not be needed, it adds an unnecessary overhead in terms of performance. Therefore, I have separated the functionalities into different classes that can be plugged into one another depending on the specific situation in which they are used. Additionally, I have implemented all the `findSolutionMappings` methods.

There are still a few TODOs remaining in this PR:
* [ ] Currently, the unit tests still fail (and, in fact, need some work to cover the different implementations).
* [ ] The two operators that use these indexes need to be extended to use the classes in the best possible way.

This PR is related to #10 

/cc @chengsijin0817

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
@@ -6,6 +6,11 @@ import java.util.NoSuchElementException;
 import se.liu.ida.hefquin.engine.data.SolutionMapping;
 import se.liu.ida.hefquin.engine.data.impl.SolutionMappingUtils;
 
+/**
+ * This is an iterator of solution mappings that consumes another iterator
+ * and passes on only the solution mappings that are compatible to a given
+ * solution mapping.
+ */
 public class SolutionMappingsIteratorWithSolMapFilter implements Iterator<SolutionMapping>
 {
 	protected final Iterator<SolutionMapping> input;
--- a/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
@@ -1,0 +1,41 @@
+package se.liu.ida.hefquin.engine.data.utils;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.data.impl.SolutionMappingUtils;
+
+public class SolutionMappingsIteratorWithSolMapFilter implements Iterator<SolutionMapping>
+{
+	protected final Iterator<SolutionMapping> input;
+	protected final SolutionMapping sm;
+
+	protected SolutionMapping nextOutputElement = null;
+
+	public SolutionMappingsIteratorWithSolMapFilter( final Iterator<SolutionMapping> input, final SolutionMapping sm ) {
+		this.input = input;
+		this.sm = sm;
+	}
+
+	@Override
+	public boolean hasNext() {
+		while ( nextOutputElement == null && input.hasNext() ) {
+			final SolutionMapping nextInputElement = input.next();
+			if ( SolutionMappingUtils.compatible(sm, nextInputElement) ) {
+				nextOutputElement = nextInputElement;
+			}
+		}
+
+		return ( nextOutputElement != null );
+	}
+
+	@Override
+	public SolutionMapping next() {
+		if ( ! hasNext() )
+			throw new NoSuchElementException();
+
+		return nextOutputElement;
+	};
+
+}
--- a/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
@@ -25,6 +25,7 @@ public class SolutionMappingsIteratorWithSolMapFilter implements Iterator<Soluti
 
 	@Override
 	public boolean hasNext() {
+		nextOutputElement = null;
 		while ( nextOutputElement == null && input.hasNext() ) {
 			final SolutionMapping nextInputElement = input.next();
 			if ( SolutionMappingUtils.compatible(sm, nextInputElement) ) {
@@ -37,8 +38,8 @@ public class SolutionMappingsIteratorWithSolMapFilter implements Iterator<Soluti
 
 	@Override
 	public SolutionMapping next() {
-		if ( ! hasNext() )
-			throw new NoSuchElementException();
+		//if ( ! hasNext() )
+		//	throw new NoSuchElementException();
 
 		return nextOutputElement;
 	};
--- a/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
@@ -25,7 +25,6 @@ public class SolutionMappingsIteratorWithSolMapFilter implements Iterator<Soluti
 
 	@Override
 	public boolean hasNext() {
-		nextOutputElement = null;
 		while ( nextOutputElement == null && input.hasNext() ) {
 			final SolutionMapping nextInputElement = input.next();
 			if ( SolutionMappingUtils.compatible(sm, nextInputElement) ) {
@@ -38,10 +37,12 @@ public class SolutionMappingsIteratorWithSolMapFilter implements Iterator<Soluti
 
 	@Override
 	public SolutionMapping next() {
-		//if ( ! hasNext() )
-		//	throw new NoSuchElementException();
+		if ( ! hasNext() )
+			throw new NoSuchElementException();
 
-		return nextOutputElement;
+		final SolutionMapping output = nextOutputElement;
+		nextOutputElement = null;
+		return output;
 	};
 
 }
--- a/src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableOverCollectionOfListsTest.java
+++ b/src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableOverCollectionOfListsTest.java
@@ -1,12 +1,8 @@
 package se.liu.ida.hefquin.engine.data.utils;
 
-import org.apache.jena.graph.Node;
-import org.apache.jena.graph.NodeFactory;
-import org.apache.jena.sparql.core.Var;
 import org.apache.jena.sparql.engine.binding.Binding;
 import org.junit.Test;
 import se.liu.ida.hefquin.engine.data.SolutionMapping;
-import se.liu.ida.hefquin.engine.data.impl.SolutionMappingUtils;
 
 import java.util.*;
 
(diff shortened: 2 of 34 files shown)


src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIteratorWithSolMapFilter.java
```diff
@@ -0,0 +1,46 @@
...
+	@Override
+	public boolean hasNext() {
+		while ( nextOutputElement == null && input.hasNext() ) {
+			final SolutionMapping nextInputElement = input.next();
+			if ( SolutionMappingUtils.compatible(sm, nextInputElement) ) {
+				nextOutputElement = nextInputElement;
+			}
+		}
+
+		return ( nextOutputElement != null );
```
chengsijin0817, line 35: After assigning the first matching solution mapping to 'nextOutputElement', method hasNext() will always return **True** as ''nextOutputElement' is not initlized when calling hasNext() the next time.

hartig: Good catch!

However, your proposed solution in commit https://github.com/LiUSemWeb/HeFQUIN/pull/52/commits/364e773c192d95858c4294380dc927904a80cad1 is also not correct, for the following two reasons: First, simply setting `nextOutputElement` to `null` at the begin of `hasNext()` is problematic because `hasNext()` may be called multiple times before `next()` is called; if that's the case, some output elements may never be returned. Second, removing the call of `hasNext()` at the begin of `next()` is problematic because `next()` may actually be called without calling `hasNext()`; for such cases, `hasNext()` has to be called from within `next()` to make sure we are in a reliable state.

I have another solution that I will commit.

hartig:
@chengsijin0817 I have checked the part where you have used the new index classes in the two hash-based operators. It looked good. I only have cleaned up the code a bit to improve its readability. In this context, I have also added a helper class called `ExpectedVariablesUtils`. See commit https://github.com/LiUSemWeb/HeFQUIN/pull/52/commits/df717da59b9ed816c2cff885afa6569015e6773d

The new helper class `ExpectedVariablesUtils` can later also be used in all other places where set operations about `ExpectedVariables` are currently done in place. However, this should not be in the given PR. Therefore, I have created #53 so that we don't forget to do it once this PR has been merged.

Now what remains to be done for this PR are some more unit tests, both for the different new index classes and for the helper classes in the new `se.liu.ida.hefquin.engine.data.utils` package.

chengsijin0817:
> @chengsijin0817 I have checked the part where you have used the new index classes in the two hash-based operators. It looked good. I only have cleaned up the code a bit to improve its readability. In this context, I have also added a helper class called `ExpectedVariablesUtils`. See commit [df717da](https://github.com/LiUSemWeb/HeFQUIN/commit/df717da59b9ed816c2cff885afa6569015e6773d)
> 
> The new helper class `ExpectedVariablesUtils` can later also be used in all other places where set operations about `ExpectedVariables` are currently done in place. However, this should not be in the given PR. Therefore, I have created #53 so that we don't forget to do it once this PR has been merged.
> 
> Now what remains to be done for this PR are some more unit tests, both for the different new index classes and for the helper classes in the new `se.liu.ida.hefquin.engine.data.utils` package.

@hartig I have added unit tests for helper classes and new index classes. Please check it in the latest commit. (4884d94 and 24d730a)

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableWithOneVarFilterTest.java
```diff
@@ -0,0 +1,80 @@
...
+import se.liu.ida.hefquin.engine.datastructures.impl.SolutionMappingsHashTableBasedOnOneVar;
+import se.liu.ida.hefquin.engine.datastructures.impl.SolutionMappingsIndexBase;
+
+import java.util.*;
+
+import static org.junit.Assert.*;
+
+public class SolutionMappingsIterableWithOneVarFilterTest {
+    @Test
+    public void hashTableWithOneInputVariable() {
```
hartig, line 19: Why is this test called something with "hashTable"? Hash tables are not relevant for the these helper classes.

Similarly, in many of the other test class for these classes you have "hashTable..." tests.

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableWithOneVarFilterTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        final Node y1 = NodeFactory.createURI("http://example.org/y1");
+        final Node y2 = NodeFactory.createURI("http://example.org/y2");
+        final Node z1 = NodeFactory.createURI("http://example.org/z1");
+        final Node z2 = NodeFactory.createURI("http://example.org/z2");
+        final Node z3 = NodeFactory.createURI("http://example.org/z3");
+        final Node p = NodeFactory.createURI("http://example.org/p");
+
+        // create SolutionMappingsIndexBase
+        final SolutionMappingsIndexBase solMHashTable = new SolutionMappingsHashTableBasedOnOneVar(var2);
```
hartig, line 32: Why do you create a hash table here? Testing the helper classes should be independent of the hash table implementations. Can't you just create a `List<SolutionMapping>`, add the sequence of input solution mappings for the test into that list, and then give the `iterator()` of that list to the constructor of the helper class.

The same seems to be the case for all the test classes about the helper classes.

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableWithOneVarFilterTest.java
```diff
@@ -0,0 +1,80 @@
...
+                var2, y2,
+                var3, z3) );
+        final Iterable<SolutionMapping> allSolMap = solMHashTable.getAllSolutionMappings();
+
+        // iterate over the subset of solution mappings that have a given value for var2
+        final Iterable<SolutionMapping> solMapAfterFilter1 = new SolutionMappingsIterableWithOneVarFilter( allSolMap, var2, y1 );
+        final Iterator<SolutionMapping> it1 = solMapAfterFilter1.iterator();
+        assertTrue( it1.hasNext() );
+        final Binding bIt1 = it1.next().asJenaBinding();
+        assertEquals( 2, bIt1.size() );
```
hartig, line 49: It would be good to check that the solution mappings are also coming out in the correct order. To this end, you have to check that each `Binding` has the correct bindings for the variables.

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableWithOneVarFilterTest.java
```diff
@@ -0,0 +1,80 @@
...
+        final Binding bIt2 = it1.next().asJenaBinding();
+        assertEquals( 2, bIt2.size() );
+
+        assertFalse( it1.hasNext() );
+
+        // no solution mappings that have p for var3
+        final Iterable<SolutionMapping> solMapAfterFilter2 = new SolutionMappingsIterableWithOneVarFilter( allSolMap, var3, p );
+        final Iterator<SolutionMapping> it2 = solMapAfterFilter2.iterator();
+
+        assertFalse( it2.hasNext() );
```
hartig, line 61: Can you please split this off into a separate unit test.

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableWithOneVarFilterTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        // return all solution mappings: no solution mappings that have value for var1
```
hartig, line 63: Also here, split this off into a separate unit test.

Given that all these unit tests use the same initialization code (e.g., creation of the same sequence of input solution mappings), that code may be moved to a helper method for this test class.

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableWithOneVarFilterTest.java
```diff
@@ -0,0 +1,80 @@
...
+        final Iterable<SolutionMapping> solMapAfterFilter3 = new SolutionMappingsIterableWithOneVarFilter( allSolMap, var1, y1 );
+        final Iterator<SolutionMapping> it3 = solMapAfterFilter3.iterator();
+        assertTrue( it3.hasNext() );
+        final Binding bIt31 = it3.next().asJenaBinding();
+        assertEquals( 2, bIt31.size() );
```
hartig, line 68: Also here, the test should check that the solution mappings come out in the correct order.

hartig (commented):
Sijin, I have looked at the unit tests and have added some comments and requests for changes. While these comments are in only one of the files, they apply to all these files with test classes about the helper classes.

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableWithOneVarFilterTest.java
chengsijin0817: @hartig Thanks for the comments. I have created separate unit tests for each use case.

chengsijin0817:
@hartig Thank you very much for your prompt reply. 
I have fixed all the aforementioned issues in the latest commit. Please check unit tests in related files. Thanks

src/test/java/se/liu/ida/hefquin/engine/data/utils/TestsForSolutionMappingsIterableWithFilter.java
```diff
@@ -0,0 +1,87 @@
...
+ * This is a class with helper functions for testing SolutionMappingsIterableWithFilter
+ */
+public class TestsForSolutionMappingsIterableWithFilter
+{
+	final Var var1 = Var.alloc("v1");
+	final Var var2 = Var.alloc("v2");
+	final Var var3 = Var.alloc("v3");
+	final Var var4 = Var.alloc("v4");
+	final Var var5 = Var.alloc("v4");
+	final Var var6 = Var.alloc("v4");
```
hartig, line 26: @chengsijin0817 is it correct that all these program variables are for the same query variable?

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableOverCollectionOfListsTest.java
```diff
@@ -0,0 +1,110 @@
...
+        // Iterate over all solution mappings contained in a collection
+        final Iterator<SolutionMapping> it = new SolutionMappingsIterableOverCollectionOfLists(solMapCollection).iterator();
+
+        assertHasNext( it, "http://example.org/y1", var2, "http://example.org/z1", var3 );
+        assertHasNext( it, "http://example.org/y1", var2, "http://example.org/z2", var3 );
+        assertHasNext( it, "http://example.org/y2", var2, "http://example.org/z3", var3 );
+
+        assertHasNext( it, "http://example.org/y1", var2, "http://example.org/z1", var3, "http://example.org/x1", var1 );
+        assertHasNext( it, "http://example.org/y1", var2, "http://example.org/z2", var3, "http://example.org/x1", var1 );
+        assertHasNext( it, "http://example.org/y2", var2, "http://example.org/z2", var3, "http://example.org/x2", var1 );
```
hartig, line 61: @chengsijin0817 Please take a look at what I did here. That makes reading the tests much easier. Can you please do the same in the other test classes here in this PR.

hartig (commented):
@chengsijin0817 Thanks a lot for addressing my earlier comments! I have cleaned up the unit testing code a bit and added two more comments that I would you you to take a look at.

src/test/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingsIterableOverCollectionOfListsTest.java
chengsijin0817: @hartig Thanks a lot for your work. 
I have updated the other test classes according to your suggestions.

src/test/java/se/liu/ida/hefquin/engine/data/utils/TestsForSolutionMappingsIterableWithFilter.java
chengsijin0817: Fixed.

hartig:
Sijin, thanks a lot for your help on this PR!! It's ready to be merged now :-)

PR #61, opened 2021-07-01
Cypher translations

scferrada:
This PR is not complete, I just implemented the triple pattern translations. This is just to check for the proposed structure.

Next I have to add a full BGP-star translation, as well as the translation from the Neo4j response to SolutionMappings.


hartig:
That's a good start!

Note that I haven't checked in detail whether the rules are captured correctly. I can do that later once the PR is more complete.

One thing that needs to be changed at some point is how the implementation captures the LPG-to-RDF* configuration. This configuration should not be fixed but it should be possible (eventually) to load such a configuration from a file. Then, there needs to be a Java class that represents such a configuration and a separate class that provides the functionality i) to apply the mappings of any given configuration, ii) to apply the inverse of these mappings, and iii) to check whether some URI/bnode is in the image of one of these mappings. Note also that some of the mappings in such a configuration may also map to bnodes.

scferrada:
Now there is a first version of full BGP translation. Mind that this produces redundant MATCH clauses. The triples in the BGP are sorted as to provide a deterministic mapping of variables.

hartig:
@scferrada I think that everything you had in this initial PR has been covered by the various other PRs that you created in the meantime. Please double check and, if you agree, close this PR as it is no longer needed.

scferrada:
There's still the code to find "certain nodes" and translate BGPs, so I'd like to keep this open until those are also addressed.

scferrada:
With PRs #244 and #245, this PR is now obsolete and I will close it.

PR #64, opened 2021-07-05
cardinality estimation

chengsijin0817:


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/PhysicalPlanVarSpecificCardinalityTable.java
```diff
@@ -0,0 +1,48 @@
...
+        return false;
+    }
+
+    public int get ( final PhysicalPlan lop, final Var v ) {
+        if ( map.containsKey(lop) ){
+            Map<Var, Integer> mapV = map.get(lop);
+            if(mapV.containsKey(v)){
+                return mapV.get(v);
+            }
+            throw new IllegalArgumentException();
```
hartig, line 43: You shouldn't throw an exception in the case of cache misses. 

As in the case of the other cache class, I would make the return type of this method Integer instead of int, and then return null in case of cache misses.

hartig (changes requested):
So far, I only had time to look at the two classes for the cardinality caches.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/PhysicalPlanCardinalityTable.java
```diff
@@ -0,0 +1,23 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+
+import java.util.*;
+
+public class PhysicalPlanCardinalityTable {
```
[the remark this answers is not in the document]
chengsijin0817: Thanks, I have renamed the cache classes (a4c11b3)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/PhysicalPlanVarSpecificCardinalityTable.java
chengsijin0817: Solved.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+            cardinality = resp.getMetadataSize();
+        } else
+            throw new IllegalArgumentException("Unsupported combination of federation member (type: " + fm.getClass().getName() + ") and request type (" + req.getClass().getName() + ")");
+
+        ppCardinalityTable.add( pp, cardinality );
+        return cardinality;
+    }
+
+    public int getJoinCardinalityEstimation( final PhysicalPlan pp, final FederationAccessManager mgr ) throws FederationAccessException {
+        if ( ppCardinalityTable.contains(pp) ){ return ppCardinalityTable.get(pp); }
```
hartig, line 62: Same issue as above, avoid calling 'contains' in addition calling 'get'.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+
+        final PhysicalOperatorForLogicalOperator lop = (PhysicalOperatorForLogicalOperator) pp.getRootOperator();
+        if ( !(lop.getLogicalOperator() instanceof LogicalOpJoin)){
```
hartig, line 65: As above, move this test before the cache lookup in line 62.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+        final PhysicalPlan pp1 = pp.getSubPlan(0);
+        final PhysicalPlan pp2 = pp.getSubPlan(1);
+
+        final int cardinality = joinCardinality( pp1, pp2, mgr );
+        ppCardinalityTable.add(pp, cardinality);
+
+        return cardinality;
+    }
+
+    public int getTPAddCardinalityEstimation( final PhysicalPlan pp, final FederationAccessManager mgr ) throws FederationAccessException {
```
hartig, line 78: Same issue here (exception).

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+        if (ppCardinalityTable.contains(pp)){ return ppCardinalityTable.get(pp); }
```
hartig, line 79: Same issue here (avoid calling 'contains').

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+
+        final PhysicalOperatorForLogicalOperator pop = (PhysicalOperatorForLogicalOperator) pp.getRootOperator();
+        final LogicalOperator lop = pop.getLogicalOperator();
+        if ( !(lop instanceof LogicalOpTPAdd)){
```
hartig, line 83: Same issue here (move up).

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+
+        final int cardinality = joinCardinality( pp1, reqBGP, mgr );
+        ppCardinalityTable.add( pp, cardinality );
+
+        return cardinality;
+    }
+
+    private int joinCardinality( final PhysicalPlan pp1, final PhysicalPlan pp2, final FederationAccessManager mgr ) throws FederationAccessException {
+        final Set<Var> certainJoinVars = ExpectedVariablesUtils.intersectionOfCertainVariables( pp1.getExpectedVariables(), pp2.getExpectedVariables() );
+        final Set<Var> possibleJoinVars = ExpectedVariablesUtils.intersectionOfPossibleVariables( pp1.getExpectedVariables(), pp2.getExpectedVariables() );
```
hartig, line 116: The possible join variables is the union of **all** variables in both plans, minus the certain join variables.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+        else if (lop instanceof LogicalOpUnion){
+            final PhysicalPlan pp1 = pp.getSubPlan(0);
+            final PhysicalPlan pp2 = pp.getSubPlan(1);
+
+            final int c1 = getCardinalityEstimationOfSpecificVar( pp1, v, mgr );
+            final int c2 = getCardinalityEstimationOfSpecificVar( pp2, v, mgr );
+
+            cardinality = c1 + c2;
+        }
```
hartig, line 187: Add an 'else' branch and throw an IllegalArgumentException in this case.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,265 @@
...
+
+        ppVarCardinalityTable.add(pp, v, cardinality);
+        return cardinality;
+    }
+
+    private int joinCardinalityBasedOnVar( PhysicalPlan pp1, PhysicalPlan pp2, Var v, FederationAccessManager mgr ) throws FederationAccessException {
```
hartig, line 193: Replace 'private' by 'protected'. This way, potential subclasses can provide extended functionality.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalitiesCache.java
```diff
@@ -0,0 +1,23 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+
+import java.util.*;
+
+public class CardinalitiesCache {
+    protected final Map<PhysicalPlan, Integer> map = new HashMap<>();
+
+    public boolean add ( final PhysicalPlan lp, final Integer card ) {
```
hartig, line 10: Why does this method return a boolean?
... and what is the meaning of this return value?
... and why does implementation of this method always return true?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,291 @@
...
+    protected final CardinalitiesCache cardinalitiesCache = new CardinalitiesCache();
+    protected final VarSpecificCardinalitiesCache varSpecificCardinalitiesCache = new VarSpecificCardinalitiesCache();
+    protected final QueryProcContext ctxt;
+
+    public CardinalityEstimation(QueryProcContext ctxt) {
+        assert ctxt != null;
+        this.ctxt = ctxt;
+    }
+
+    public int getCardinalityEstimationOfLeafNode( final PhysicalPlan pp ) throws QueryOptimizationException, FederationAccessException {
```
hartig, line 38: Please read my previous comment about the exceptions again. I think that this method (and the other public methods of this class) should not throw a FederationAccessException but they may only throw a QueryOptimizationException.

Hence, in the implementation of this method, the function calls that may throw a FederationAccessException should be wrapped in a try-catch block where the catch branch would then throw a new QueryOptimizationException as described in my previous comment.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,291 @@
...
+                }
+                else c = getCardinalityEstimationOfSpecificVar( pp2, v );
+                cardinality = max(c, cardinality);
+            }
+        }
+
+        return cardinality;
+    }
+
+    public int getCardinalityEstimationOfSpecificVar( final PhysicalPlan pp, final Var v ) throws FederationAccessException, QueryOptimizationException {
```
hartig, line 172: Does this method have to be public?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,291 @@
...
+        if ( varCachedCard != null ) {
+            return varCachedCard;
+        }
+
+        final PhysicalOperatorForLogicalOperator pop = (PhysicalOperatorForLogicalOperator) pp.getRootOperator();
+        final LogicalOperator lop = pop.getLogicalOperator();
+
+        int cardinality = 0;
+        if (lop instanceof LogicalOpRequest) {
+            cardinality = getCardinalityEstimationOfLeafNode(pp );
```
hartig, line 183: Is this part really correct? It may very well be! I just observe that this part is completely independent of the given variable v, which _might_ be a mistake, or not. Please double-check.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/VarSpecificCardinalitiesCache.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.jena.sparql.core.Var;
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class VarSpecificCardinalitiesCache {
+    protected final Map<PhysicalPlan, Map<Var, Integer>> map = new HashMap<>();
+
+    public boolean add ( final PhysicalPlan lop, final Var v, final Integer card ) {
```
hartig, line 12: Also here, what is the meaning of the boolean that this method returns?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/VarSpecificCardinalitiesCache.java
```diff
@@ -0,0 +1,48 @@
...
+        else mapIn.put(v, card);
+
+        return true;
+    }
+
+    public boolean contains ( final PhysicalPlan lop, final Var v ) {
+        if ( map.containsKey(lop) ){
+            Map<Var, Integer> mapV = map.get(lop);
+            if( mapV.containsKey(v) ){ return true; }
+            return false;
```
hartig, line 32: This line, together with the previous one, can be simplified as follows.

return mapV.containsKey(v);

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/VarSpecificCardinalitiesCache.java
```diff
@@ -0,0 +1,48 @@
...
+            mapV.put(v, card);
+            map.put(lop, mapV);
+        }
+        else mapIn.put(v, card);
+
+        return true;
+    }
+
+    public boolean contains ( final PhysicalPlan lop, final Var v ) {
+        if ( map.containsKey(lop) ){
```
hartig, line 29: You can avoid this call of 'containsKey' by using 'get' directly and, then, checking whether the obtained Map is not null.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/VarSpecificCardinalitiesCache.java
```diff
@@ -0,0 +1,48 @@
...
+            Map<Var, Integer> mapV = map.get(lop);
+            if( mapV.containsKey(v) ){ return true; }
+            return false;
+        }
+        return false;
+    }
+
+    public Integer get ( final PhysicalPlan lop, final Var v ) {
+        if ( map.containsKey(lop) ){
```
hartig, line 38: This call of 'containsKey' can be avoided as well.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/VarSpecificCardinalitiesCache.java
```diff
@@ -0,0 +1,48 @@
...
+            Map<Var, Integer> mapV = map.get(lop);
+            if(mapV.containsKey(v)){
```
hartig, line 40: There is no need for this if test. If the Map does not contain the given variable, then 'get' will return null.

hartig (changes requested):

hartig (approved):
This looks good now and I would merge it. Please confirm that you are also okay with merging it now.

chengsijin0817:
> This looks good now and I would merge it. Please confirm that you are also okay with merging it now.

chengsijin0817:
> This looks good now and I would merge it. Please confirm that you are also okay with merging it now.

Thanks. I am okay with merging it now.

PR #66, opened 2021-07-09
Cost model

chengsijin0817:


hartig:
Do you want me to look at this one as well or are you still working on it?

chengsijin0817:
> Do you want me to look at this one as well or are you still working on it?

Yes, please take a look at this PR. Thanks.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/RootOperatorCostCache.java
```diff
@@ -0,0 +1,23 @@
...
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class RootOperatorCostCache {
+    protected final Map<PhysicalPlan, List<Integer>> map = new HashMap<>();
+
+    public void add ( final PhysicalPlan pp, final List<Integer> metrics ) {
```
hartig, line 12: Please introduce an actual interface for the parameter 'metric', where this interface has a specific getter method for each of the five metrics, e.g., getNumberOfRequests().

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,304 @@
...
+
+        double overallCost = 0;
+        for (  int i = 0; i < measureMetrics.size(); i++ ){
+            overallCost = overallCost + measureMetrics.get(i) * weight.get(i);
+        }
+
+        return overallCost;
+    }
+
+    protected List<Integer> costMetrics( final PhysicalPlan pp, final List<Integer> overallMetrics ) throws QueryOptimizationException {
```
hartig, line 46: This return type is totally meaningless. Please introduce an actual interface for the objects that this method returns, where this interface has a specific getter method for each of the five metrics, e.g., getNumberOfRequests().

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,304 @@
...
+            for ( int j = 0; j < li.size(); j++ ){
+                overallMetrics.set( j, overallMetrics.get(j) + li.get(j) ) ;
+            }
+            costMetrics( children.get(i), overallMetrics);
+        }
+
+        return overallMetrics;
+    }
+
+    protected List<PhysicalPlan> findChildren(final PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 60: This method doesn't make much sense. Instead of materializing all the subplans again into another (temporary) list, you can simply do loop of line 64 at line 49.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,304 @@
...
+
+    protected List<Integer> costMetrics( final PhysicalPlan pp, final List<Integer> overallMetrics ) throws QueryOptimizationException {
+        final List<PhysicalPlan> children = findChildren(pp);
+
+        for ( int i = 0; i < children.size(); i++ ){
+            final List<Integer> li = metricsOfRootOperator( children.get(i) );
+            for ( int j = 0; j < li.size(); j++ ){
+                overallMetrics.set( j, overallMetrics.get(j) + li.get(j) ) ;
+            }
+            costMetrics( children.get(i), overallMetrics);
```
hartig, line 54: This implementation is really hard to understand. In particular, I don't get how this is supposed to work with passing the list called 'overallMetrics' through the different levels of recursive calls here. Maybe it is just a matter of naming of the methods (e.g., the name 'costMetrics' is not really helpful in terms of making clear what this method does; similar for 'metricsOfRootOperator') .

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,304 @@
...
+    protected List<Integer> metricsOfRootOperator(final PhysicalPlan pp ) throws QueryOptimizationException {
+        final List<Integer> metrics = new ArrayList<Integer>();
+
+        metrics.add( numberOfRequests( pp.getRootOperator(), pp ) );
+        metrics.add( shippedRDFTermsForRequests( pp.getRootOperator(), pp ) );
+        metrics.add( shippedRDFVarsForRequests( pp.getRootOperator(), pp ) );
+        metrics.add( shippedRDFTermsForResponses( pp.getRootOperator(), pp ) );
+        metrics.add( shippedRDFVarsForResponses( pp.getRootOperator(), pp ) );
+
+        rootOperatorCostCache.add( pp, metrics );
```
hartig, line 79: If you add something to this cache here, shouldn't you also try to find it in the cache at the begin of this method?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,304 @@
...
+        for ( int i = 0; i < numChildren; ++i ){
+            children.add( pp.getSubPlan(i) );
+        }
+        return children;
+    }
+
+    protected List<Integer> metricsOfRootOperator(final PhysicalPlan pp ) throws QueryOptimizationException {
+        final List<Integer> metrics = new ArrayList<Integer>();
+
+        metrics.add( numberOfRequests( pp.getRootOperator(), pp ) );
```
hartig, line 73: For this line and the following four lines, I wonder: why do you need to pass these two parameters to these functions? I mean, if you only pass the PhysicalPlan, then the functions can get the respective root operator themselves. Or, do you call these functions from somewhere else again but then with operators that are not the root operator of the plan being passed?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalitiesCache.java
```diff
@@ -0,0 +1,22 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+
+import java.util.*;
+
+public class CardinalitiesCache {
```
hartig, line 7: This is the latest version of this class, right?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
```diff
@@ -0,0 +1,300 @@
...
+import se.liu.ida.hefquin.engine.queryplan.physical.PhysicalOperatorForLogicalOperator;
+import se.liu.ida.hefquin.engine.queryplan.physical.impl.PhysicalOpRequest;
+import se.liu.ida.hefquin.engine.queryplan.physical.impl.PhysicalPlanWithNullaryRootImpl;
+import se.liu.ida.hefquin.engine.queryplan.utils.ExpectedVariablesUtils;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+import se.liu.ida.hefquin.engine.queryproc.QueryProcContext;
+
+import java.util.*;
+
+public class CardinalityEstimation {
```
hartig, line 28: This is the latest version of this class, right?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/VarSpecificCardinalitiesCache.java
```diff
@@ -0,0 +1,35 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils;
+
+import org.apache.jena.sparql.core.Var;
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class VarSpecificCardinalitiesCache {
```
hartig, line 9: This is the latest version of this class, right?

hartig (changes requested):
Here are some initial comments about this PR. Please let me know once you have addressed them.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
chengsijin0817: Yes, all functions can get the respective root operator themselves. 
I have grouped these functions into a new class: CostOfRootOperator, to explicitly show that these functions provide the cost of the root operator (not the total cost of a given physical plan)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalitiesCache.java
chengsijin0817: Yes

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/CardinalityEstimation.java
chengsijin0817: Yes

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/utils/VarSpecificCardinalitiesCache.java
chengsijin0817: Yes

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,304 @@
...
+import se.liu.ida.hefquin.engine.queryplan.physical.PhysicalOperatorForLogicalOperator;
+import se.liu.ida.hefquin.engine.queryplan.physical.impl.*;
+import se.liu.ida.hefquin.engine.queryplan.utils.ExpectedVariablesUtils;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+import se.liu.ida.hefquin.engine.queryproc.QueryProcContext;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.CardinalityEstimation;
+
+import java.util.*;
+
+public class CostModel extends CardinalityEstimation{
```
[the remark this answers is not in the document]
chengsijin0817: Done

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
chengsijin0817: Fixed

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
chengsijin0817: I have renamed related functions and reorganized them. Please find the latest version in Class: CostOfPhysicalPlan

chengsijin0817:
> Here are some initial comments about this PR. Please let me know once you have addressed them.

@hartig Thank you very much for your work. 
I tried to solve all issues mentioned in your comments. Please feel free to check it. Thanks.

hartig (changes requested):
I'm afraid you misunderstood my previous comments. I wanted you to create an interface (and corresponding implementation) that captures the multidimensional cost of plans (i.e., something like the current `Metrics` interface but with methods that are actual getter methods rather than functions that determine the values for any given physical plan). Additionally, I wanted you to create interfaces (and corresponding implementations) that capture the functionality to determine the multidimensional cost for any given physical plan (i.e., that produce an object of the first type of interface). Your current classes seem to be a mixture of both types of things. This should be separated. Please let me know if I have managed to make myself more clear now.

chengsijin0817:
> I'm afraid you misunderstood my previous comments. I wanted you to create an interface (and corresponding implementation) that captures the multidimensional cost of plans (i.e., something like the current `Metrics` interface but with methods that are actual getter methods rather than functions that determine the values for any given physical plan). Additionally, I wanted you to create interfaces (and corresponding implementations) that capture the functionality to determine the multidimensional cost for any given physical plan (i.e., that produce an object of the first type of interface). Your current classes seem to be a mixture of both types of things. This should be separated. Please let me know if I have managed to make myself more clear now.

@hartig Thank you very much for your explanation. 
According to your comments, I created an interface **Metrics** and corresponding implementation **MetricsImpl** to capture the multidimensional cost. In addition, I added an interface **CostOfPhysicalPlan** to determine the multidimensional cost for a given physical plan. The corresponding implementation can be found in **CostOfPhysicalPlanImpl**. 
Please let me know if I misunderstood any part or missed any information.

hartig (changes requested):
I have added some comments about various details. Additionally, I have some more general comments:

1/ The naming of the new interfaces (and, correspondingly, their implementation classes) does not make much sense from a code readability perspective. `Metrics` is too generic and `CostOfPhysicalPlan` sounds like a class that represents the cost of a physical plan (rather than the functionality to determine such cost), and `CostOfRootOperator` has the same problem. Therefore, I propose you rename these things as follows: `Metrics` should be renamed to `CostOfPhysicalPlan`, `CostOfPhysicalPlan` should be renamed to `CostFunctionsForPhysicalPlans`, and `CostOfRootOperator` should be renamed to `CostFunctionsForRootOperators`.

2/ The latter two (`CostFunctionsForPhysicalPlans` and `CostFunctionsForRootOperators`) should be defined as an interface with a corresponding implementation class, respectively.

3/ `CostFunctionsForPhysicalPlans` should be extended with another public method that returns a `CostOfPhysicalPlan` object for a given physical plan. Internally, the implementation of this method may then simply use the other methods to obtain the values for the `CostOfPhysicalPlan` object. This new method can then be the main method that is called from the outside (i.e., from your `CostModel` class).

chengsijin0817:
> I have added some comments about various details. Additionally, I have some more general comments:
> 
> 1/ The naming of the new interfaces (and, correspondingly, their implementation classes) does not make much sense from a code readability perspective. `Metrics` is too generic and `CostOfPhysicalPlan` sounds like a class that represents the cost of a physical plan (rather than the functionality to determine such cost), and `CostOfRootOperator` has the same problem. Therefore, I propose you rename these things as follows: `Metrics` should be renamed to `CostOfPhysicalPlan`, `CostOfPhysicalPlan` should be renamed to `CostFunctionsForPhysicalPlans`, and `CostOfRootOperator` should be renamed to `CostFunctionsForRootOperators`.
> 
> 2/ The latter two (`CostFunctionsForPhysicalPlans` and `CostFunctionsForRootOperators`) should be defined as an interface with a corresponding implementation class, respectively.
> 
> 3/ `CostFunctionsForPhysicalPlans` should be extended with another public method that returns a `CostOfPhysicalPlan` object for a given physical plan. Internally, the implementation of this method may then simply use the other methods to obtain the values for the `CostOfPhysicalPlan` object. This new method can then be the main method that is called from the outside (i.e., from your `CostModel` class).

Thank you very much for the comments. 
I have updated related interfaces and classes, and extend CostFunctionsForPhysicalPlans based on your comments. Please check it.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlans.java
```diff
@@ -0,0 +1,20 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.evolutionaryAlgorithm;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+
+public interface CostFunctionsForPhysicalPlans {
+
+    int getTotalNumberOfRequests( PhysicalPlan pp ) throws QueryOptimizationException;
```
hartig, line 8: Now I suggest to rename these methods to `determine...` (e.g., `determineTotalNumberOfRequests`) which makes more clear that they are not simply getter methods but they involve some processing.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperators.java
```diff
@@ -0,0 +1,20 @@
...
+
+    int getShippedRDFTermsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getShippedRDFVarsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getShippedRDFTermsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getShippedRDFVarsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getIntermediateResultsSize( final PhysicalPlan pp ) throws QueryOptimizationException;
```
hartig, line 18: Same comment here as for the methods of the `CostFunctionsForPhysicalPlans` interface: rename these methods to `determine...`

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.evolutionaryAlgorithm;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.CardinalityEstimation;
+
+public class CostFunctionsForPhysicalPlansImpl implements CostFunctionsForPhysicalPlans{
+    protected final CostFunctionsForRootOperatorsImpl costFunctionForRoot;
```
hartig, line 8:
```suggestion
    protected final CostFunctionsForRootOperators costFunctionForRoot;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public CostFunctionsForPhysicalPlansImpl(CardinalityEstimation cardEstimate) {
```
hartig, line 10:
```suggestion
    public CostFunctionsForPhysicalPlansImpl( final CardinalityEstimation cardEstimate ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        this.costFunctionForRoot = new CostFunctionsForRootOperatorsImpl(cardEstimate);
+    }
+
+    public CostOfPhysicalPlan getCostOfPhysicalPlan( final PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 14: Add this method also to the interface.

After doing so, add the annotation `@Override`

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        final int numberOfRequests = getTotalNumberOfRequests(pp);
+        final int shippedRDFTermsForRequests = getTotalShippedRDFTermsForRequests(pp);
+        final int shippedRDFVarsForRequests = getTotalShippedRDFVarsForRequests(pp);
+        final int shippedRDFTermsForResponses = getTotalShippedRDFTermsForResponses(pp);
+        final int shippedRDFVarsForResponses = getTotalShippedVarsForResponses(pp);
+        final int getIntermediateResultsSize = getTotalIntermediateResultsSize(pp);
+
+        final CostOfPhysicalPlan costOfPhysicalPlan = new CostOfPhysicalPlanImpl( numberOfRequests, shippedRDFTermsForRequests , shippedRDFVarsForRequests, shippedRDFTermsForResponses, shippedRDFVarsForResponses, getIntermediateResultsSize);
+
+        return costOfPhysicalPlan;
```
hartig, line 25:
```suggestion
        return new CostOfPhysicalPlanImpl( numberOfRequests, shippedRDFTermsForRequests , shippedRDFVarsForRequests, shippedRDFTermsForResponses, shippedRDFVarsForResponses, getIntermediateResultsSize);
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    public int getTotalNumberOfRequests( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 28:
```suggestion
    @Override
    public int getTotalNumberOfRequests( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        for ( int i = 0; i < pp.numberOfSubPlans(); i++ ){
+            final int costOfRootOp = costFunctionForRoot.getNumberOfRequests( pp.getSubPlan(i) );
+            totalNumberOfRequests += costOfRootOp ;
+            getTotalNumberOfRequests( pp.getSubPlan(i) );
+        }
+
+        return totalNumberOfRequests;
+    }
+
+    public int getTotalShippedRDFTermsForRequests( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 43:
```suggestion
    @Override
    public int getTotalShippedRDFTermsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        for ( int i = 0; i < pp.numberOfSubPlans(); i++ ){
+            final int costOfRootOp = costFunctionForRoot.getShippedRDFTermsForRequests( pp.getSubPlan(i) );
+            totalShippedRDFTermsForRequests += costOfRootOp ;
+            getTotalShippedRDFTermsForRequests( pp.getSubPlan(i) );
+        }
+
+        return totalShippedRDFTermsForRequests;
+    }
+
+    public int getTotalShippedRDFVarsForRequests( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 58:
```suggestion
    @Override
    public int getTotalShippedRDFVarsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        for ( int i = 0; i < pp.numberOfSubPlans(); i++ ){
+            final int costOfRootOp = costFunctionForRoot.getShippedRDFVarsForRequests( pp.getSubPlan(i) );
+            totalShippedRDFVarsForRequests += costOfRootOp ;
+            getTotalShippedRDFVarsForRequests( pp.getSubPlan(i) );
+        }
+
+        return totalShippedRDFVarsForRequests;
+    }
+
+    public int getTotalShippedRDFTermsForResponses( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 73:
```suggestion
    @Override
    public int getTotalShippedRDFTermsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        for ( int i = 0; i < pp.numberOfSubPlans(); i++ ){
+            final int costOfRootOp = costFunctionForRoot.getShippedRDFTermsForResponses( pp.getSubPlan(i) );
+            totalShippedRDFTermsForResponses += costOfRootOp ;
+            getTotalShippedRDFTermsForResponses( pp.getSubPlan(i) );
+        }
+
+        return totalShippedRDFTermsForResponses;
+    }
+
+    public int getTotalShippedVarsForResponses( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 88:
```suggestion
    @Override
    public int getTotalShippedVarsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        for ( int i = 0; i < pp.numberOfSubPlans(); i++ ){
+            final int costOfRootOp = costFunctionForRoot.getShippedRDFVarsForResponses( pp.getSubPlan(i) );
+            totalShippedVarsForResponses += costOfRootOp ;
+            getTotalShippedVarsForResponses( pp.getSubPlan(i) );
+        }
+
+        return totalShippedVarsForResponses;
+    }
+
+    public int getTotalIntermediateResultsSize( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 103:
```suggestion
    @Override
    public int getTotalIntermediateResultsSize( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.CardinalityEstimation;
+
+import java.util.Iterator;
+import java.util.Set;
+
+public class CostFunctionsForRootOperatorsImpl implements CostFunctionsForRootOperators {
+    protected final CardinalityEstimation cardEstimate;
+
+    public CostFunctionsForRootOperatorsImpl(CardinalityEstimation cardEstimate ) {
```
hartig, line 31:
```suggestion
    public CostFunctionsForRootOperatorsImpl( final CardinalityEstimation cardEstimate ) {
        assert cardEstimate != null;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+        this.cardEstimate = cardEstimate;
+    }
+
+    public int getNumberOfRequests( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 35:
```suggestion
    @Override
    public int getNumberOfRequests( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+            cost = 1;
+        } else if ( pop instanceof BasePhysicalOpBinaryJoin){
+            cost = 0;
+        } else
+            throw new IllegalArgumentException("Unsupported Physical Operator");
+
+        return cost;
+    }
+
+    public int getShippedRDFTermsForRequests( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 52:
```suggestion
    @Override
    public int getShippedRDFTermsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+            cost = numberOfTerms;
+        } else if ( pop instanceof BasePhysicalOpBinaryJoin ) {
+            cost = 0;
+        } else
+            throw new IllegalArgumentException("Unsupported Physical Operator");
+
+        return cost;
+    }
+
+    public int getShippedRDFVarsForRequests( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 98:
```suggestion
    @Override
    public int getShippedRDFVarsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+            cost = numberOfVars;
+        } else if ( pop instanceof BasePhysicalOpBinaryJoin ) {
+            cost = 0;
+        } else
+            throw new IllegalArgumentException("Unsupported Physical Operator");
+
+        return cost;
+    }
+
+    public int getShippedRDFTermsForResponses( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 147:
```suggestion
    @Override
    public int getShippedRDFTermsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+
+        } else if ( lop instanceof LogicalOpJoin){
+            cost = 0;
+        } else
+            throw new IllegalArgumentException("Unsupported Physical Operator");
+
+        return cost;
+    }
+
+    public int getShippedRDFVarsForResponses( PhysicalPlan pp ) throws QueryOptimizationException {
```
hartig, line 192:
```suggestion
    @Override
    public int getShippedRDFVarsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+
+        } else if ( lop instanceof LogicalOpJoin ){
+            cost = 0;
+        } else
+            throw new IllegalArgumentException("Unsupported Physical Operator");
+
+        return cost;
+    }
+
+    public int getIntermediateResultsSize(final PhysicalPlan pp) throws QueryOptimizationException {
```
hartig, line 236:
```suggestion
    @Override
    public int getIntermediateResultsSize(final PhysicalPlan pp) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,37 @@
...
+import se.liu.ida.hefquin.engine.queryproc.QueryProcContext;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.CardinalityEstimation;
+
+import java.util.*;
+
+public class CostModel {
+    protected final QueryProcContext ctxt;
+    protected final PhysicalPlanCostCache physicalPlanCostCache;
+
+    public CostModel(QueryProcContext ctxt ) {
```
hartig, line 14:
```suggestion
    public CostModel( final QueryProcContext ctxt ) {
        assert ctxt != null;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostModel.java
```diff
@@ -0,0 +1,37 @@
...
+        this.ctxt = ctxt;
+        this.physicalPlanCostCache = new PhysicalPlanCostCache();
+    }
+
+    public Double wrapUpCostAsOneValue( final PhysicalPlan pp ) throws QueryOptimizationException {
+        final CostFunctionsForPhysicalPlansImpl costFunctionsForPP = new CostFunctionsForPhysicalPlansImpl( new CardinalityEstimation(ctxt) );
```
hartig, line 20:
```suggestion
        final CostFunctionsForPhysicalPlans costFunctionsForPP = new CostFunctionsForPhysicalPlansImpl( new CardinalityEstimation(ctxt) );
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+        } else if ( pop instanceof PhysicalOpBindJoin || pop instanceof PhysicalOpBindJoinWithUNION || pop instanceof PhysicalOpBindJoinWithFILTER || pop instanceof PhysicalOpBindJoinWithVALUES){
+            cost = 1;
+        } else if ( pop instanceof PhysicalOpRequest){
+            cost = 1;
+        } else if ( pop instanceof BasePhysicalOpBinaryJoin){
+            cost = 0;
+        } else
+            throw new IllegalArgumentException("Unsupported Physical Operator");
+
+        return cost;
```
hartig, line 49:
```suggestion
        if ( pop instanceof PhysicalOpIndexNestedLoopsJoin ) {
            return getIntermediateResultsSize(pp.getSubPlan(0));
        } else if ( pop instanceof PhysicalOpBindJoin || pop instanceof PhysicalOpBindJoinWithUNION || pop instanceof PhysicalOpBindJoinWithFILTER || pop instanceof PhysicalOpBindJoinWithVALUES ) {
            // TODO: Returning 1 is not entirely correct here. The actual number of requests depends on the page size used for the bind-join requests.
            return 1;
        } else if ( pop instanceof PhysicalOpRequest ) {
            // TODO: Returning 1 is not entirely correct here. The actual number of requests depends on the page size used for the requests.
            return 1;
        } else if ( pop instanceof BasePhysicalOpBinaryJoin ) {
            return 0;
        } else {
            throw new IllegalArgumentException("Unsupported Physical Operator");
        }
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+            numberOfJoinVars = ExpectedVariablesUtils.intersectionOfCertainVariables( subPP.getExpectedVariables(), reqTP.getExpectedVariables() ).size();
+        } else if ( lop instanceof LogicalOpRequest) {
+            final DataRetrievalRequest req = ((LogicalOpRequest<?, ?>) lop).getRequest();
+            if ( req instanceof SPARQLRequest) {
+                numberOfTerms = numberOfTermsOfBGP((BGP) ((SPARQLRequest) req).getQueryPattern());
+            } else if ( req instanceof TriplePatternRequest) {
+                numberOfTerms = 3 - ((TriplePatternRequest) req).getQueryPattern().numberOfVars();
+            } else
+                throw new IllegalArgumentException("Unsupported request type (" + req.getClass().getName() + ")");
+        }
```
hartig, line 81: Please add an else-branch with an IllegalArgumentException.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+        } else if ( pop instanceof PhysicalOpBindJoinWithFILTER || pop instanceof PhysicalOpBindJoinWithVALUES || pop instanceof PhysicalOpBindJoin ){
+            cost = numberOfTerms + intermediateResultSize * numberOfJoinVars;
+        } else if ( pop instanceof PhysicalOpRequest ) {
+            cost = numberOfTerms;
+        } else if ( pop instanceof BasePhysicalOpBinaryJoin ) {
+            cost = 0;
+        } else
+            throw new IllegalArgumentException("Unsupported Physical Operator");
+
+        return cost;
```
hartig, line 95:
```suggestion
        if ( pop instanceof PhysicalOpIndexNestedLoopsJoin || pop instanceof PhysicalOpBindJoinWithUNION ){
            return intermediateResultSize * (numberOfTerms + numberOfJoinVars);
        } else if ( pop instanceof PhysicalOpBindJoinWithFILTER || pop instanceof PhysicalOpBindJoinWithVALUES || pop instanceof PhysicalOpBindJoin ){
            return numberOfTerms + intermediateResultSize * numberOfJoinVars;
        } else if ( pop instanceof PhysicalOpRequest ) {
            return numberOfTerms;
        } else if ( pop instanceof BasePhysicalOpBinaryJoin ) {
            return 0;
        } else
            throw new IllegalArgumentException("Unsupported Physical Operator");
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperators.java
```diff
@@ -0,0 +1,20 @@
...
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+
+public interface CostFunctionsForRootOperators {
+
+    int getNumberOfRequests( final PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getShippedRDFTermsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getShippedRDFVarsForRequests( final PhysicalPlan pp ) throws QueryOptimizationException;
```
hartig, line 12: Shouldn't this method be called `getShippedVarsForRequests` instead?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperators.java
```diff
@@ -0,0 +1,20 @@
...
+
+    int getShippedRDFTermsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getShippedRDFVarsForResponses( final PhysicalPlan pp ) throws QueryOptimizationException;
```
hartig, line 16: Same here, `getShippedVarsForResponses` instead?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlans.java
```diff
@@ -0,0 +1,20 @@
...
+
+    int getTotalShippedRDFTermsForRequests( PhysicalPlan pp ) throws QueryOptimizationException;
+
+    int getTotalShippedRDFVarsForRequests( PhysicalPlan pp ) throws QueryOptimizationException;
```
hartig, line 12: Shouldn't this method be called `getTotalShippedVarsForRequests` instead?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostOfPhysicalPlan.java
```diff
@@ -0,0 +1,17 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.evolutionaryAlgorithm;
+
+public interface CostOfPhysicalPlan {
+
+    int getNumberOfRequests( );
+
+    int getShippedRDFTermsForRequests( ) ;
+
+    int getShippedRDFVarsForRequests( );
```
hartig, line 9: `getShippedVarsForRequests` ?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostOfPhysicalPlan.java
```diff
@@ -0,0 +1,17 @@
...
+
+    int getShippedRDFTermsForResponses( );
+
+    int getShippedRDFVarsForResponses( );
```
hartig, line 13: `getShippedVarsForResponses`?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+            numberOfJoinVars = ExpectedVariablesUtils.intersectionOfCertainVariables( subPP.getExpectedVariables(), reqTP.getExpectedVariables() ).size();
+        } else if ( lop instanceof LogicalOpRequest ) {
+            final DataRetrievalRequest req = ((LogicalOpRequest<?, ?>) lop).getRequest();
+            if ( req instanceof SPARQLRequest) {
+                numberOfVars = numberOfVarsOfBGP((BGP) ((SPARQLRequest) req).getQueryPattern());
+            } else if ( req instanceof TriplePatternRequest) {
+                numberOfVars = ((TriplePatternRequest) req).getQueryPattern().numberOfVars();
+            } else
+                throw new IllegalArgumentException("Unsupported request type (" + req.getClass().getName() + ")");
+        }
```
hartig, line 128: else-branch with IllegalArgumentException

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperatorsImpl.java
```diff
@@ -0,0 +1,271 @@
...
+
+        final int cost;
```
hartig, line 130: Remove as in my comments above.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+        return costOfPhysicalPlan;
+    }
+
+    public int getTotalNumberOfRequests( PhysicalPlan pp ) throws QueryOptimizationException {
+        int totalNumberOfRequests = 0;
+        if ( pp.numberOfSubPlans() == 0 ){
+            totalNumberOfRequests += costFunctionForRoot.getNumberOfRequests( pp );
+        }
+
```
hartig, line 33:
```suggestion
        if ( pp.numberOfSubPlans() == 0 ){
            return costFunctionForRoot.getNumberOfRequests( pp );
        }

        int totalNumberOfRequests = 0;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        for ( int i = 0; i < pp.numberOfSubPlans(); i++ ){
+            final int costOfRootOp = costFunctionForRoot.getNumberOfRequests( pp.getSubPlan(i) );
+            totalNumberOfRequests += costOfRootOp ;
+            getTotalNumberOfRequests( pp.getSubPlan(i) );
```
hartig, line 37: I don't think that this is correct. There are two issues:

1/ The value returned by `getTotalNumberOfRequests( pp.getSubPlan(i) )` must be added to `totalNumberOfRequests`.

2/ After fixing the first issue, lines 35 and 36 become obsolete because the number of requests of the root op of the current subplan is added automatically when calling `getTotalNumberOfRequests( pp.getSubPlan(i) )`.

hartig: The same issue is also in all the following methods within this class.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+        return totalNumberOfRequests;
+    }
+
+    public int getTotalShippedRDFTermsForRequests( PhysicalPlan pp ) throws QueryOptimizationException {
+        int totalShippedRDFTermsForRequests = 0;
+        if ( pp.numberOfSubPlans() == 0 ){
+            totalShippedRDFTermsForRequests += costFunctionForRoot.getShippedRDFTermsForRequests( pp );
+        }
+
```
hartig, line 48:
```suggestion
        if ( pp.numberOfSubPlans() == 0 ){
            return costFunctionForRoot.getShippedRDFTermsForRequests( pp );
        }

        int totalShippedRDFTermsForRequests = 0;
```

hartig (changes requested):
We are getting there ;-)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
chengsijin0817: Hi Olaf, I have fixed these two issues. 
Another TODO is that the cost of **sub**query plans are not cached in the current implementation. 
Can you please check my comments in Class 'CostFunctionsForPhysicalPlansImpl' and let me know what you think?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,121 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.evolutionaryAlgorithm;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.CardinalityEstimation;
+
+public class CostFunctionsForPhysicalPlansImpl implements CostFunctionsForPhysicalPlans{
+    protected final CostFunctionsForRootOperators costFunctionForRoot;
+    // protected final PhysicalPlanCostCache physicalPlanCostCache;
```
hartig, line 9: You don't need this cache here. The purpose of this class can be to provide only the actual cost functions, and the option of caching the costs of plans can be handled completely outside of this class, exactly as it is currently done already in the 'CostModel' class. Hence, you can safely remove all these portions of commented code from this class.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForPhysicalPlansImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        */
+        int totalNumberOfRequests = costFunctionForRoot.determineNumberOfRequests( pp );
+        if ( pp.numberOfSubPlans() == 0 ){
+            return totalNumberOfRequests;
+        }
+        for ( int i = 0; i < pp.numberOfSubPlans(); i++ ){
+            totalNumberOfRequests += determineTotalNumberOfRequests( pp.getSubPlan(i) );
+            // totalNumberOfRequests += getCostOfPhysicalPlan( pp.getSubPlan(i) ).getNumberOfRequests();
+        }
+        // TODO: How to add totalNumberOfRequests of (sub)PhysicalPlans to this cache?
```
hartig, line 45: See my comment above. Caching can be taken care of outside of this class.

hartig (changes requested):
This is likely the last request for changes.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/CostFunctionsForRootOperators.java
```diff
@@ -0,0 +1,20 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.evolutionaryAlgorithm;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+
+public interface CostFunctionsForRootOperators {
+
+    int determineNumberOfRequests( final PhysicalPlan pp ) throws QueryOptimizationException;
```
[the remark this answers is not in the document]
chengsijin0817: You're right. It is more logical to take 'PhysicalOperator' as input for these methods. 
However, only passing 'PhysicalOperator' to these methods does not work because the cost of a root operator depends on the 'intermediate result size' of the subquery. 

When determining the cost of a root operator, we need to call **getSubPlan**() by passing 'PhysicalPlan' to these methods. Does it make sense?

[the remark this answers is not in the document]
hartig: Ah, yes! You are right.

hartig (approved):
Now everything looks good. Thanks for your efforts!

I will merge this PR ...

Issue #67, opened 2021-07-16
Non-blocking cardinality estimator


hartig:
The current implementation to estimate result cardinalities of leaf nodes in plans performs requests sequentially, one after another. That's highly inefficient and needs to be replaced by an implementation in which requests are performed in parallel. The purpose of this issue is to develop such an implementation. A prerequisite for it is a non-blocking, multi-threaded FederationAccessManager as per #49.

PR #68, opened 2021-07-20
New FederationAccessManager that uses response processors

hartig:
This PR is a first step towards addressing #40 and #67


hartig:
Now, this became more complex than expected.

PR #70, opened 2021-08-10
Rewriting rule

chengsijin0817:
Add a set of rewriting rules for operator tpAdd: 
Convert tpAdd to Join; 
Convert two tpAdd between different physical algorithms; 
Order tweaking of two tpAdd;


hartig:
I am just starting to look at this. Some initial observations and questions:

1. Since the rewriting rules are independent of the evolutionary algorithm, they should not be under the package for that algorithm. Moved them to a package called `se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting`.
2. Is there a reason why the return types of the methods in `Rule` are not the basic types (`boolean` and `double`) but the corresponding Java classes? If not, change that.
3. It is not clear to me whether the interface `Rule` is meant to represent the general notion of our different rewriting rules or a possible application of such a rule. Note that there are cases in which the same rule can be applied to multiple different subplans of a given plan. I think there needs to be an interface for the general notion of our rules and another interface for each possible rule application:

```java
public interface RewritingRule
{
   double getPriority();
   Set<RuleApplication> determineAllPossibleApplications(PhysicalPlan p);
}

public interface RuleApplication
{
   RewritingRule getRule();

   /** Returns the plan to which the rule is meant to be applied. */
   PhysicalPlan getPlan();

   PhysicalPlan getResultingPlan();
}
```

hartig:
Regarding the third point in my previous comment, when looking into your implementations of your `Rule` interface, I notice that you seem to assume that rewriting rules are applied to the root operator of a given plan. However, that's not correct. Instead, rewriting rules may be applied to any operator within a plan. Then, when performing the actual rule application (i.e., when creating the new plan), the subplans that produce the input for the rewritten operator can also be used directly as subplans in the new plan, and the same holds for all sibling subplans; but all parent subplans (all the way to the root) need to be created anew because they are based on the new subplan that is rooted in the rewritten operator.

hartig:
Thanks for the clarification. In this case, I will wait until you have developed this other external class to see the code for the whole process.

Can you please also add a small JavaDoc comment to the `Rule` interface to clarify the purpose of this interface (namely, that this interface is meant to capture rewriting rules that are meant to be applied to the root of a given plan, where such a given plan may be a subplan within a bigger plan). Also, the name "Rule" is too generic. Please rename the interface to `RewritingRule`.

Two more things:

1. The priorities should not be hard-coded within the classes themselves but, instead, they should be passed to the instances of these classes when these instances are created. Hence, there needs to be a constant member variable within the classes and that member is initialized based on a value given to the constructor.
2. Instead of creating the instances of the rule classes as separate class members in `RuleInstances`, they should be created as elements of a `Set<RewritingRule>`. That way, it will be easier to iterate over them when you try to identify the rules that can be applied for any given subplan. Additionally, it will make it easier if, for some reason (e.g., in some experiment), we only want to consider different subsets of all the rules.

chengsijin0817:
> Thanks for the clarification. In this case, I will wait until you have developed this other external class to see the code for the whole process.

Okay, I will let you know when I have developed this external class.

> 
> Can you please also add a small JavaDoc comment to the `Rule` interface to clarify the purpose of this interface (namely, that this interface is meant to capture rewriting rules that are meant to be applied to the root of a given plan, where such a given plan may be a subplan within a bigger plan). Also, the name "Rule" is too generic. Please rename the interface to `RewritingRule`.
> 
> Two more things:
> 
> 1. The priorities should not be hard-coded within the classes themselves but, instead, they should be passed to the instances of these classes when these instances are created. Hence, there needs to be a constant member variable within the classes and that member is initialized based on a value given to the constructor.
> 2. Instead of creating the instances of the rule classes as separate class members in `RuleInstances`, they should be created as elements of a `Set<RewritingRule>`. That way, it will be easier to iterate over them when you try to identify the rules that can be applied for any given subplan. Additionally, it will make it easier if, for some reason (e.g., in some experiment), we only want to consider different subsets of all the rules.

Sounds reasonable. I will do so, thanks.

hartig:
I notice that you have merged the main branch into the branch of this PR after the main branch had been updated when I merged the other PRs earlier today. Now I wonder: has the branch of this PR become incompatible with the main branch? I don't think it should because the other two PRs have touched completely different parts of the code base. No?

chengsijin0817:
> I notice that you have merged the main branch into the branch of this PR after the main branch had been updated when I merged the other PRs earlier today. Now I wonder: has the branch of this PR become incompatible with the main branch? I don't think it should because the other two PRs have touched completely different parts of the code base. No?

No, this branch is compatible with the main branch. 
I merged the main branch into this branch before pushing my commits, was just to check if there were any conflicts.
Should I avoid doing so in future work?

hartig:
Yes, please avoid merging updated versions of the main branch into your feature branches unless strictly necessary; that is, unless the updated main branch contains something new that is needed also in the feature branch. There are two reasons why it should be avoided: First, it may somehow mess up the commit history after such a PR is merged (some commits may then be duplicated). Second, and more important, it may result in changes that have been merged into main from the other PRs to become part of this PR as well; hence, when looking at the changes of this PR, the may be changes that have not actually been done for this PR, which makes really difficult to see what exactly is actually meant to be changed by this PR.

hartig:
I see that there are already more than 50 files in this PR. This will make reviewing the PR and understanding the essence of the implementation quite difficult. Therefore, I would like to ask you to remove all the files with the actual rules. Maybe keep one as an example. The rest you can copy somewhere else on your computer to save them for later. Then, we can come back to them in a second PR after this PR here is ready and merged; and this PR here can allow us to get the overall framework right.

chengsijin0817:
> I see that there are already more than 50 files in this PR. This will make reviewing the PR and understanding the essence of the implementation quite difficult. Therefore, I would like to ask you to remove all the files with the actual rules. Maybe keep one as an example. The rest you can copy somewhere else on your computer to save them for later. Then, we can come back to them in a second PR after this PR here is ready and merged; and this PR here can allow us to get the overall framework right.

Thanks. I have removed all files with actual rules and only keep one as an example: **RuleConvertTPAddIndexNLJToHashJoin**.

In addition, I added two classes for showing the rewriting process: **ApplicableRules** and **ApplyRewritingRule**. The first class is used for identifying all applicable rules for a given physical plan, and the second one can be used to get a new physical plan by applying a selected rule to a physical plan. 
Please feel free to add any comments.

hartig (changes requested):
At this point, I do not have any comments in the actual code. Instead, I have some general issues that need to be addressed:

1. The implementation of `ApplyRewritingRule` is incorrect because it applies the given rule in all possible places within the given plan. In our evolutionary algorithm, every "possible rule application" in a given parent plan is about changing only one part of the parent plan by applying the rule to only one subplan, even if the rule may be applied to other subplans as well.
2. Related to the previous issue, the problem is already in the implementation of `ApplicableRules`. Here, you collect all rules that can be applied to at least one subplan, but you do not record which subplan this is, and you do not account for cases in which the same rule may be applied in multiple subplans. In this context, take another look at the interfaces that I had proposed in [one of my earlier comments](https://github.com/LiUSemWeb/HeFQUIN/pull/70#issuecomment-896156978). There, the distinction between the rules and the possible applications of these rules was captured explicitly. In your current implementation, this distinction is not even implicitly captured.
3. Classes such as `IdentifyPhysicalOperatorOfTPAdd` should be independent of the actual plans to be checked. If they are dependent as in your current implementation, you always have to create objects from these classes (as you do, for instance, in `RuleConvertTPAddIndexNLJToHashJoin`), which costs unnecessary CPU cycles and can be avoided if these classes simply contain static functions (e.g., `boolean matchTPAdd(final PhysicalPlan plan)`).
4. Another comment about classes such as `IdentifyPhysicalOperatorOfTPAdd`: The code in the methods can be simplified by removing the explicit if-else blocks which are not needed here. For instance, in `matchTPAdd` you can simply do:
```java
    return pop.getLogicalOperator() instanceof LogicalOpTPAdd;
```

One other thing: there are still a lot of helper classes in this PR that are not relevant to the one example rule (e.g., you have ChangeOrderOf... classes and several irrelevant ConvertTPAddTo... classes). Can you please remove them as well, just to allow us to focus on the essential parts for the moment.

chengsijin0817:
> At this point, I do not have any comments in the actual code. Instead, I have some general issues that need to be addressed:
> 
> 1. The implementation of `ApplyRewritingRule` is incorrect because it applies the given rule in all possible places within the given plan. In our evolutionary algorithm, every "possible rule application" in a given parent plan is about changing only one part of the parent plan by applying the rule to only one subplan, even if the rule may be applied to other subplans as well.
> 2. Related to the previous issue, the problem is already in the implementation of `ApplicableRules`. Here, you collect all rules that can be applied to at least one subplan, but you do not record which subplan this is, and you do not account for cases in which the same rule may be applied in multiple subplans. 

In the current implementation, I simply pick one rule from all applicable rules and apply this rule to all applicable sub-plans.
You are right! I should have applied the rule to only one child plan and only changed one part of the parent plan. I considered this problem and managed to record the applicable rule and corresponding subplans for a given plan, but ran into a problem when using the selected rule-subplan pair to rewrite the physical plan. 
The main problem is that I only record the rule and corresponding sub-plans, but did not find a suitable way to represent the position of the sub-plan in the parent plan.
For example, a given plan contains two identical sub-plans sp1 and sp2. How to record them separately in the rule application? If keeping them only once in the set of RuleApplication, it is hard to decide which part of the parent plan to apply the selected rule application?

> In this context, take another look at the interfaces that I had proposed in [one of my earlier comments](https://github.com/LiUSemWeb/HeFQUIN/pull/70#issuecomment-896156978). There, the distinction between the rules and the possible applications of these rules was captured explicitly. In your current implementation, this distinction is not even implicitly captured.

I took another look at the interfaces you proposed. This is a good way to capture the rules and possible applications of these rules. I have only two questions regarding this proposal:
    1. For each RuleApplication, what is the expected output of the method getResultingPlan()? Is it a resulting sub-plan that applying the rule to its corresponding sub-plans, or a complete physical plan that applying the given rule application?
    2. Related to the example I mentioned above, how does this proposal deal with identical sub-plans within one parent plan?
    Please let me know if I misunderstand any point.

> 3. Classes such as `IdentifyPhysicalOperatorOfTPAdd` should be independent of the actual plans to be checked. If they are dependent as in your current implementation, you always have to create objects from these classes (as you do, for instance, in `RuleConvertTPAddIndexNLJToHashJoin`), which costs unnecessary CPU cycles and can be avoided if these classes simply contain static functions (e.g., `boolean matchTPAdd(final PhysicalPlan plan)`).

Thanks a lot for the comment. I can address this issue.

> 4. Another comment about classes such as `IdentifyPhysicalOperatorOfTPAdd`: The code in the methods can be simplified by removing the explicit if-else blocks which are not needed here. For instance, in `matchTPAdd` you can simply do:
> 
> ```java
>     return pop.getLogicalOperator() instanceof LogicalOpTPAdd;
> ```
> 
Sure.

> One other thing: there are still a lot of helper classes in this PR that are not relevant to the one example rule (e.g., you have ChangeOrderOf... classes and several irrelevant ConvertTPAddTo... classes). Can you please remove them as well, just to allow us to focus on the essential parts for the moment.

Okay

hartig:
> For example, a given plan contains two identical sub-plans sp1 and sp2. How to record them separately in the rule application?

While subplans may be identical (i.e., syntactically equivalent), I would assume that they would still be represented by two different Java objects.

> 1. For each RuleApplication, what is the expected output of the method getResultingPlan()? Is it a resulting sub-plan that applying the rule to its corresponding sub-plans, or a complete physical plan that applying the given rule application?

We can decide ;-)

I think the better choice of the two options would be for getResultingPlan() to return the complete rewritten plan rather than returning only some subplan. In other words, the returned plan should be a version of getPlan() in which the rewriting rule has been applied (once!) somewhere within the plan. By using this option, we can maintain a set of RuleApplication objects for each physical plan (which, initially, would be the union of all sets returned by determineAllPossibleApplications of all the rules) and, when a plan is picked as parent, we pick one RuleApplication object from that set and, then, only have to call getResultingPlan() to directly obtain the next candidate. I think that this process also corresponds very closely to what we have in the pseudo code of the algorithm.

> 2. Related to the example I mentioned above, how does this proposal deal with identical sub-plans within one parent plan?

Also here, I think that's not a problem because, even if subplans are identical, they are represented by different java objects. Hence, in such a case, there will be multiple separate RuleApplication objects where each of them is for one of these subplans, which internally is represented by a different PhysicalPlan object.

chengsijin0817:
> > For example, a given plan contains two identical sub-plans sp1 and sp2. How to record them separately in the rule application?
> 
> While subplans may be identical (i.e., syntactically equivalent), I would assume that they would still be represented by two different Java objects.
> 
> > 1. For each RuleApplication, what is the expected output of the method getResultingPlan()? Is it a resulting sub-plan that applying the rule to its corresponding sub-plans, or a complete physical plan that applying the given rule application?
> 
> We can decide ;-)
> 
> I think the better choice of the two options would be for getResultingPlan() to return the complete rewritten plan rather than returning only some subplan. In other words, the returned plan should be a version of getPlan() in which the rewriting rule has been applied (once!) somewhere within the plan. By using this option, we can maintain a set of RuleApplication objects for each physical plan (which, initially, would be the union of all sets returned by determineAllPossibleApplications of all the rules) and, when a plan is picked as parent, we pick one RuleApplication object from that set and, then, only have to call getResultingPlan() to directly obtain the next candidate. I think that this process also corresponds very closely to what we have in the pseudo code of the algorithm.
> 
> > 1. Related to the example I mentioned above, how does this proposal deal with identical sub-plans within one parent plan?
> 
> Also here, I think that's not a problem because, even if subplans are identical, they are represented by different java objects. Hence, in such a case, there will be multiple separate RuleApplication objects where each of them is for one of these subplans, which internally is represented by a different PhysicalPlan object.

Thanks a lot for your clarification. It makes sense. 
Then I will update interfaces and implement method getResutingPlan() to return the complete rewritten plan.

hartig:
> ... and implement method getResutingPlan() to return the complete rewritten plan.

Idea for this implementation: The classes that implement RuleApplication may have a member that is the subplan that will be replaced by the rewriting (in addition to another member that is the overall plan, i.e., the one to be returned when getPlan() is called). Then, in getResultingPlan(), the overall plan is traversed recursively until the correct subplan is reached; next, the subplan is rewritten according to the corresponding rewriting rule and then the recursion goes back up to the overall root to create the new version of the overall plan. Notice that most of these steps are generic (i.e., independent of the actual rewriting rule); hence, they can be implemented in a common base class.

One more idea to make reaching the correct subplan more efficient: The base class that implements RuleApplication may actually contain not only a member that is the subplan to be found (and to which the rewriting will then be applied) but also a list of subplans along the path from the overall root to the subplan to be found (i.e., the first entry in that list would be the overall root and the last entry would be the subplan to be found). This list can be constructed recursively in determineAllPossibleApplications and then simply provided to the constructor of the respective RuleApplication object. Then, given such a list, you can avoid doing another complete breadth-first or depth-first search of the overall plan within getResultingPlan(); instead, the list already tells you all the ancestors (parent, grandparent, etc) of the rewritten subplan that need to be recreated when producing the new version of the overall plan.

chengsijin0817:
> ... the list already tells you all the ancestors (parent, grandparent, etc) of the rewritten subplan that need to be recreated when producing the new version of the overall plan.

Thank you sooo much for the really helpful tips. This saved me!!!
I was struggling with how to find the ancestors of the rewritten sub-plan and construct a complete plan.

hartig:
One more idea: When you construct these lists during the recursion in determineAllPossibleApplications, you can use a stack of the subplans reached along the current traversal path. That way, you can easily push and pop subplans as you recurse/traverse. At each point when you want to create a RuleApplication object, you take a snapshot of the stack as an array and pass that to the constructor of the RuleApplication object. Then, the only difference to what I described above is that the subplans in the list/array within each RuleApplication object are in the reverse order (starting with the subplan to be rewritten, ending with the root of the overall plan).

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RewritingRuleBaseImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+        if ( canBeAppliedTo(plan) ) {
+            ruleApplications.add( new RuleAppliConvertTPAddIndexNLJToHashJoin((PhysicalPlan[]) subPlans.toArray()) );
+        }
+
+        final int numChildren = plan.numberOfSubPlans();
+        for (int i = 0; i < numChildren; ++i) {
+            determineAllPossibleApplications( plan.getSubPlan(i) );
+        }
+        return ruleApplications;
```
hartig, line 36: Before returning from this function, you need to remove the current plan from the stack again.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RewritingRuleBaseImpl.java
```diff
@@ -0,0 +1,41 @@
...
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.impl.RuleAppliConvertTPAddIndexNLJToHashJoin;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.Stack;
+
+public abstract class RewritingRuleBaseImpl implements RewritingRule{
+    protected Set<RuleApplication> ruleApplications = new HashSet<>();
+    protected Stack<PhysicalPlan> subPlans = new Stack<>();
```
hartig, line 12: I notice that you never clear `ruleApplications` after `determineAllPossibleApplications` has been called from the outside. Therefore, whenever `determineAllPossibleApplications` is called again from the outside, for the next plan, there may still be entries in this set. These entries should have been removed.

Ideally, `ruleApplications` and `subPlans` should not be members of this class but, instead, they are created new whenever `determineAllPossibleApplications` is called from the outside and then passed around during the recursion of an additional internal method. That way, the implementation becomes thread safe, meaning that `determineAllPossibleApplications` can even be called from multiple threads at the same time.

So, you need something like the following:

```java
    @Override
    public Set<RuleApplication> determineAllPossibleApplications( final PhysicalPlan plan ) {
        return Set<RuleApplication> _determineAllPossibleApplications( plan, new HashSet<>(), new Stack<>() );
    }

    protected Set<RuleApplication> _determineAllPossibleApplications( final PhysicalPlan plan, final Set<RuleApplication> ruleApplications, final Stack<PhysicalPlan> currentPath ) {
        // here you have the code that you currently have in determineAllPossibleApplications,
        // with the only difference that you call _determineAllPossibleApplications for the recursion
    }
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public PhysicalPlan getResultingPlan() {
+        index = 0;
+        return constructPlan( subPlans[0] );
+    }
+
+    protected PhysicalPlan constructPlan( final PhysicalPlan plan ) {
+        final PhysicalPlan[] newSubPlans = getRewrittenSubPlans(plan);
```
hartig, line 31: This doesn't look right. You don't need to (try to) rewrite all subplans recursively. Instead, begin with the last plan in `subPlans` because that's the subplan to which the actual rewriting is meant to be applied. After that, you can walk your way up in the overall plan (by going over `subPlans` in the reverse order---last to first) to recreate the parent subplans all the way back to the overall root.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/impl/RuleAppliConvertTPAddIndexNLJToHashJoin.java
```diff
@@ -0,0 +1,36 @@
...
+public class RuleAppliConvertTPAddIndexNLJToHashJoin extends RuleApplicationBaseImpl {
+    protected final ConstructSubPPBasedOnUnaryOperator helper = new ConstructSubPPBasedOnUnaryOperator();
+
+    public RuleAppliConvertTPAddIndexNLJToHashJoin( final PhysicalPlan[] subPlans ) {
+        super(subPlans);
+    }
+
+    @Override
+    public RewritingRule getRule() {
+        return new RuleConvertTPAddIndexNLJToHashJoin( 0.15 );
```
hartig, line 23: No. The base class (RuleApplicationBaseImpl) should have a member for this, which is initialized by passing the rule in the constructor (i.e., in addition to passing `subPlan`).

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryplan.physical.BinaryPhysicalOp;
+import se.liu.ida.hefquin.engine.queryplan.physical.NaryPhysicalOp;
+import se.liu.ida.hefquin.engine.queryplan.physical.UnaryPhysicalOp;
+import se.liu.ida.hefquin.engine.queryplan.physical.impl.PhysicalPlanWithBinaryRootImpl;
+import se.liu.ida.hefquin.engine.queryplan.physical.impl.PhysicalPlanWithNaryRootImpl;
+import se.liu.ida.hefquin.engine.queryplan.physical.impl.PhysicalPlanWithUnaryRootImpl;
+
+public abstract class RuleApplicationBaseImpl implements RuleApplication{
+    protected final PhysicalPlan[] subPlans;
```
hartig, line 12: The meaning and the purpose of this member should be made clearer by renaming it to: `pathToTargetSubPlan`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    protected RewritingRule rule;
+
+    public RuleApplicationBaseImpl( final PhysicalPlan[] subPlans, final RewritingRule rule ){
```
hartig, line 15:
```suggestion
    public RuleApplicationBaseImpl( final PhysicalPlan[] subPlans, final RewritingRule rule ) {
        assert subPlans.length > 0;
        assert rule != null;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    public RewritingRule getRule() {
+        return rule;
+    }
+
+    @Override
+    public PhysicalPlan getResultingPlan() {
+        final int numStep = subPlans.length;
+        PhysicalPlan rewrittenPlan = rewrittenSubPlan( subPlans[numStep-1] );
+
+        for (int i = numStep-2; i >= 0; i--) {
```
hartig, line 35: Since the `numStep` variable is _not_ meant to represent the number of a step but, instead, the total number of step**s**, it should be called `numSteps`. Also, you can immediately initialize it to `subPlans.length - 1`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+            rewrittenPlan = constructPlan( subPlans[i], rewrittenPlan, i );
+        }
+        return rewrittenPlan;
+    }
+
+    protected PhysicalPlan constructPlan( final PhysicalPlan parent, final PhysicalPlan rewrittenChild, final int indexOfPath) {
+
+        if ( parent.numberOfSubPlans() == 0){
+            return parent;
+        }
```
hartig, line 45: This case should never occur! Throw an IllegalArgumentException.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+            PhysicalPlan[] newSubPlans = getRewrittenSubPlans( parent, rewrittenChild, indexOfPath );
+            return new PhysicalPlanWithNaryRootImpl((NaryPhysicalOp) parent.getRootOperator(), newSubPlans);
+        }
+    }
+
+    protected PhysicalPlan[] getRewrittenSubPlans( final PhysicalPlan parent, final PhysicalPlan rewrittenChild, final int indexOfPath) {
+        final int numChildren = parent.numberOfSubPlans();
+        final PhysicalPlan[] children = new PhysicalPlan[numChildren];
+        for (int i = 0; i< numChildren; i++) {
+            if ( parent.getSubPlan(0) == subPlans[indexOfPath+1]) {
```
hartig, line 63: While this is correct, it seems more logical to me to pass the original version of `rewrittenChild` instead of passing the index; you may call this argument `originalChild`. I also think that this may make the intention of the code even clearer.

hartig: Additionally, there is a small bug here in line 63. It must be: `parent.getSubPlan(i)`

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    protected PhysicalPlan constructPlan( final PhysicalPlan parent, final PhysicalPlan rewrittenChild, final int indexOfPath) {
+
+        if ( parent.numberOfSubPlans() == 0){
+            return parent;
+        }
+        else if ( parent.numberOfSubPlans() == 1 ){
+            return new PhysicalPlanWithUnaryRootImpl((UnaryPhysicalOp) parent.getRootOperator(), rewrittenChild);
+        }
+        else if ( parent.numberOfSubPlans() == 2 ){
+            PhysicalPlan[] newSubPlans = getRewrittenSubPlans( parent, rewrittenChild, indexOfPath );
```
hartig, line 50:
```suggestion
            final PhysicalPlan[] newSubPlans = getRewrittenSubPlans( parent, rewrittenChild, indexOfPath );
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+            return new PhysicalPlanWithBinaryRootImpl((BinaryPhysicalOp) parent.getRootOperator(), newSubPlans[0], newSubPlans[1]);
+        }
+        else {
+            PhysicalPlan[] newSubPlans = getRewrittenSubPlans( parent, rewrittenChild, indexOfPath );
```
hartig, line 54:
```suggestion
            final PhysicalPlan[] newSubPlans = getRewrittenSubPlans( parent, rewrittenChild, indexOfPath );
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/RuleApplicationBaseImpl.java
```diff
@@ -0,0 +1,75 @@
...
+                children[i] = rewrittenChild;
+            }
+            else {
+                children[i] = parent.getSubPlan(i);
+            }
+        }
+        return children;
+    }
+
+    protected abstract PhysicalPlan rewrittenSubPlan( final PhysicalPlan plan );
```
hartig, line 73: I would call this method `rewritePlan`

hartig (changes requested):
We are getting close. There are only a few small things left.

chengsijin0817:
> We are getting close. There are only a few small things left.

Thanks, the mentioned issues have been fixed.

hartig (approved):
This looks good now. I will try to resolve current merge conflicts and, then, merge.

chengsijin0817:
> This looks good now. I will try to resolve current merge conflicts and, then, merge.

Thanks. ;-)

hartig:
Merging now.

@chengsijin0817 please wait with creating another PR for the other rules. I first want to refactor things a bit more.

Issue #76, opened 2021-08-12
Bug in greedy plan enumeration algorithm


hartig:
When working on #75 to address #74, I noticed the following bug in the class that wraps the usage of the greedy plan enumeration algorithm (i.e., the class `GreedyQueryOptimizer`, formerly known as `WrappingDPAlgorithm`). The else-branch in the function called `optimizePlan` simply returns the given plan. This else-branch is for cases in which the root operator of the given plan is not a multiway join. While it is correct that the enumeration algorithm cannot simply be applied as it is done in the if-branch for plans that have a multiway join as their root, it is incorrect to simply return such a plan because, somewhere inside the plan, there may be more multiway joins. To fix this issue, the else-branch needs to traverse into all subplans, rewrite them recursively if necessary, and then combine the resulting, potentially changed subplans using the same type of operator that is the root of the given plan.

@chengsijin0817

hartig:
After merging PR #79, the bug is now in the class called `SimpleJoinOrderingQueryOptimizer`.

PR #78, opened 2021-08-12
Address issue #76: bug in greedy plan enumeration algorithm

chengsijin0817:


hartig:
Is this one ready to be reviewed?

(PS. Sorry for having caused you extra work by merging #79 after you had already branched GreedyQueryOptimizer to work on this issue.)

chengsijin0817:
> Is this one ready to be reviewed?

Yes, please review it.
> 
> (PS. Sorry for having caused you extra work by merging #79 after you had already branched GreedyQueryOptimizer to work on this issue.)

It doesn't cost much work. Please let me know whether the revision on this issue works.

hartig:
I have looked at your fix. I am afraid it's not that easy. In fact, this change makes things worse ;-)

First of all, notice that I had added the method called `hasMultiwayJoinAsRoot` to make the code a bit more readable by separating out this slightly verbose part about checking the type of the root operator.

Now, you cannot simply replace this check in the way as you have done. The reason is that the `JoinPlanOptimizer` (which, essentially, is simply a wrapper around the "dynamic programming" algorithm you had implemented) creates a plan that combines all the given subplans via binary joins, which is a correct rewriting if the root operator of the input plan was a multiway join with all these subplans as its children. However, for any other type of root operator, it is not correct to replace it with joins of its children.

Hence, the fix needs to be something like the following.

```java
public PhysicalPlan optimizePlan( final PhysicalPlan plan ) throws QueryOptimizationException {
    final PhysicalPlan[] optSubPlans = getOptimizedSubPlans(plan);
    if ( hasMultiwayJoinAsRoot(plan) ){
        return joinPlanOptimizer.determineJoinPlan(optSubPlans);
    }
    else if ( plan.numberOfSubPlans() == 0 ) {
    	return plan; 
    }
    else {
    	//   ... TODO ...
    	// here you need to create a new plan whose root operator is equivalent
    	// to the root operator of the given plan, i.e., plan.getRootOperator(),
      	// and the children of this root operator in this new plan are the
    	// (recursively) optimized subplans, i.e., optSubPlans
    }
}
```

chengsijin0817:
> I have looked at your fix. I am afraid it's not that easy. In fact, this change makes things worse ;-)
> 
> First of all, notice that I had added the method called `hasMultiwayJoinAsRoot` to make the code a bit more readable by separating out this slightly verbose part about checking the type of the root operator.
> 
> Now, you cannot simply replace this check in the way as you have done. The reason is that the `JoinPlanOptimizer` (which, essentially, is simply a wrapper around the "dynamic programming" algorithm you had implemented) creates a plan that combines all the given subplans via binary joins, which is a correct rewriting if the root operator of the input plan was a multiway join with all these subplans as its children. However, for any other type of root operator, it is not correct to replace it with joins of its children.
> 
> Hence, the fix needs to be something like the following.
> 
> ```java
> public PhysicalPlan optimizePlan( final PhysicalPlan plan ) throws QueryOptimizationException {
>     final PhysicalPlan[] optSubPlans = getOptimizedSubPlans(plan);
>     if ( hasMultiwayJoinAsRoot(plan) ){
>         return joinPlanOptimizer.determineJoinPlan(optSubPlans);
>     }
>     else if ( plan.numberOfSubPlans() == 0 ) {
>     	return plan; 
>     }
>     else {
>     	//   ... TODO ...
>     	// here you need to create a new plan whose root operator is equivalent
>     	// to the root operator of the given plan, i.e., plan.getRootOperator(),
>       	// and the children of this root operator in this new plan are the
>     	// (recursively) optimized subplans, i.e., optSubPlans
>     }
> }
> ```

Thank you for pointing out the issue. It makes sense. 
I tried to fix the issue according to your comments. Please check it.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/SimpleJoinOrderingQueryOptimizer.java
```diff
@@ -41,14 +48,26 @@ public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimi
...
-            return joinPlanOptimizer.determineJoinPlan(subplans);
+            return joinPlanOptimizer.determineJoinPlan(optSubPlans);
         }
-        else {
-        	// TODO incorrect!! There may be multiway joins inside the subplans of the given plan.
+        else if (hasNullaryOpAsRoot(plan)){
             return plan;
         }
+        else if ( hasUnaryOpAsRoot(plan) ){
+            final PhysicalPlan newChild = optimizePlan(optSubPlans[0]);
```
hartig, line 60: It is not necessary to call `optimizePlan` for `optSubPlans[0]`.  The new subplans in `optSubPlans` have already been optimized recursively. That's what the method `getOptimizedSubPlans` (as called above) is doing.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/SimpleJoinOrderingQueryOptimizer.java
```diff
@@ -41,14 +48,26 @@ public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimi
...
+            return new PhysicalPlanWithUnaryRootImpl((UnaryPhysicalOp) plan.getRootOperator(), newChild);
+        }
+        else if ( hasBinaryOpAsRoot(plan) ){
+            final PhysicalPlan newChild1 = optimizePlan(optSubPlans[0]);
+            final PhysicalPlan newChild2 = optimizePlan(optSubPlans[1]);
```
hartig, line 65: Also here, there is no need to call `optimizePlan` again.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/SimpleJoinOrderingQueryOptimizer.java
```diff
@@ -41,14 +48,26 @@ public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimi
...
+            return new PhysicalPlanWithBinaryRootImpl((BinaryPhysicalOp) plan.getRootOperator(), newChild1, newChild2);
+        }
+        else {
```
hartig, line 68: Before this final else-branch, there needs to be another else-if for the case of n-ary root operators other than multiway joins.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/SimpleJoinOrderingQueryOptimizer.java
```diff
@@ -41,14 +48,26 @@ public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimi
...
-            final PhysicalPlan[] subplans = getOptimizedSubPlans(plan);
-            return joinPlanOptimizer.determineJoinPlan(subplans);
+            return joinPlanOptimizer.determineJoinPlan(optSubPlans);
         }
-        else {
-        	// TODO incorrect!! There may be multiway joins inside the subplans of the given plan.
+        else if (hasNullaryOpAsRoot(plan)){
             return plan;
         }
+        else if ( hasUnaryOpAsRoot(plan) ){
```
hartig, line 59: Instead of calling these helper methods (`hasNullaryOpAsRoot`, etc), I think it is safe to distinguish the cases based on the number of subplans (i.e., `plan.numberOfSubPlans()`). That way, the code becomes a bit more efficient because it avoids doing the casts to `PhysicalOperatorForLogicalOperator` that are done in each of these helper methods.

hartig (changes requested):
A few smaller things need to be changed.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/SimpleJoinOrderingQueryOptimizer.java
```diff
@@ -41,14 +48,26 @@ public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimi
...
         if ( hasMultiwayJoinAsRoot(plan) ){
-            final PhysicalPlan[] subplans = getOptimizedSubPlans(plan);
-            return joinPlanOptimizer.determineJoinPlan(subplans);
+            return joinPlanOptimizer.determineJoinPlan(optSubPlans);
         }
-        else {
-        	// TODO incorrect!! There may be multiway joins inside the subplans of the given plan.
+        else if ( hasMultiwayUnionAsRoot(plan) ){
+            return new PhysicalPlanWithNaryRootImpl((NaryPhysicalOp) plan.getRootOperator(), optSubPlans);
+        }
```
hartig, line 58: Similar to my previous comment about the other helper methods (which now you have removed as suggested), instead of using this new helper method, `hasMultiwayUnionAsRoot`, we can use the number of subplans to recognize this case. Doing so will make the final else-branch (with the exception) obsolete. Therefore, you can simple remove this else-if branch here and move the code from within this else-if branch to the final else-branch, i.e.:

```java
   // ...
   else {
      return new PhysicalPlanWithNaryRootImpl((NaryPhysicalOp) plan.getRootOperator(), optSubPlans);
   }
```

hartig (changes requested):
One final change.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/SimpleJoinOrderingQueryOptimizer.java
chengsijin0817: Solved.

hartig (approved):
Now it looks good.

Issue #84, opened 2021-08-18
Potential mistake in two cost functions


hartig:
When working on PR #83 I noticed a possible mistake in the implementation of two of the (operator-specific) cost metrics, namely the one about number of RDF terms shipped in responses and number of variables shipped in responses. These implementations are now in the classes `CFRNumberOfTermsShippedInResponses` and `CFRNumberOfVarsShippedInResponses`, respectively. The mistake that I mean is described in the TODO comment in [lines 66-68](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfTermsShippedInResponses.java#L66) and [lines 64-66](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfVarsShippedInResponses.java#L64), respectively.

@chengsijin0817 can you please take a look at this and check whether my suspicion is right. In this case, maybe you also have to change the corresponding part in the paper?

chengsijin0817:
> When working on PR #83 I noticed a possible mistake in the implementation of two of the (operator-specific) cost metrics, namely the one about number of RDF terms shipped in responses and number of variables shipped in responses. These implementations are now in the classes `CFRNumberOfTermsShippedInResponses` and `CFRNumberOfVarsShippedInResponses`, respectively. The mistake that I mean is described in the TODO comment in [lines 66-68](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfTermsShippedInResponses.java#L66) and [lines 64-66](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfVarsShippedInResponses.java#L64), respectively.
> 
[...]

Issue #85, opened 2021-08-18
Mistake in four cost functions (SPARQLRequest may not contain only a BGP)


hartig:
When working on PR #83 I noticed a mistake in the implementation of four of the (operator-specific) cost metrics, namely the ones that are about shipped terms and shipped variables. In particular, the issue is about the cases in which the request is of type `SPARQLRequest` and the number of terms/variables needs to be determined. The current implementation simply assumes that each `SPARQLRequest` contains just a BGP, which is not necessarily true.

The issue is highlighted by TODOs in the corresponding source code files; see:
* [line 67 in CFRNumberOfVarsShippedInRequests](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfVarsShippedInRequests.java#L67)
* [line 63 in CFRNumberOfTermsShippedInRequests](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfTermsShippedInRequests.java#L63)
* [line 77 in CFRNumberOfTermsShippedInResponses](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfTermsShippedInResponses.java#L77)
* [line 73 in CFRNumberOfVarsShippedInResponses](https://github.com/LiUSemWeb/HeFQUIN/blob/7f6c12c012a5d9de26d4aa30cfa03d6f2195e935/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfVarsShippedInResponses.java#L73)

Note, however, that the latter two points may disappear in the context of addressing #84.

PR #86, opened 2021-08-19
address the issue #84: potential mistake in two cost functions

chengsijin0817:


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfTermsShippedInResponses.java
```diff
@@ -61,26 +58,19 @@ else if ( lop instanceof LogicalOpBGPAdd ) {
...
+				final SPARQLRequest req = (SPARQLRequest) ((LogicalOpRequest<?, ?>) lop).getRequest();
+				final int numberOfVars = QueryPatternUtils.getVariablesInPattern( req.getQueryPattern() ).size();
 				return futureIntResSize.thenApply( intResSize -> numberOfVars * intResSize );
 			}
+			else if ( fm instanceof TPFServer || fm instanceof BRTPFServer) {
+				return futureIntResSize.thenApply( intResSize -> 3 * intResSize );
+			}
 			else {
-				throw createIllegalArgumentException(req);
+				throw createIllegalArgumentException(fm);
```
hartig, line 73: Before throwing this exception, `futureIntResSize` should be canceled. See the case of `LogicalOpTPAdd` a bit further up in this source file.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/costmodel/CFRNumberOfVarsShippedInResponses.java
```diff
@@ -59,24 +56,19 @@ else if ( lop instanceof LogicalOpBGPAdd ) {
 			}
 		}
 		else if ( lop instanceof LogicalOpRequest ) {
-			final DataRetrievalRequest req = ((LogicalOpRequest<?, ?>) lop).getRequest();
+			final CompletableFuture<Integer> futureIntResSize = initiateCardinalityEstimation(plan);
```
hartig, line 59: In this case, `futureIntResSize` should be created not here but only within the if-branch for SPARQL endpoints. The reason for this is that `futureIntResSize` is not needed for TPF/brTPF servers. We shouldn't start a cardinality estimation process if we don't actually need the cardinality estimate ;-)

src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
```diff
@@ -76,6 +76,17 @@ else if ( p instanceof BGP ) {
...
 
+	public static Set<Var> getVariablesInPattern( final SPARQLGraphPattern queryPattern ) {
+		Set<Var> result = new HashSet<>();
+		if ( queryPattern instanceof TriplePattern) {
+			result = getVariablesInPattern( (TriplePattern) queryPattern );
+		}
+		else if ( queryPattern instanceof BGP) {
+			result = getVariablesInPattern( (BGP) queryPattern);
+		}
+		return result;
```
hartig, line 87:
```suggestion
		if ( queryPattern instanceof TriplePattern ) {
			return getVariablesInPattern( (TriplePattern) queryPattern );
		}
		else if ( queryPattern instanceof BGP ) {
			return getVariablesInPattern( (BGP) queryPattern );
		}
		else {
			throw new UnsupportedOperationException("Getting the variables from arbitrary SPARQL patterns is an open TODO (type of Jena Op in the current case: " + queryPattern.asJenaOp().getClass().getName() + ").");
		}
```

hartig: No need to create a HashSet that is never used ;-)

hartig (changes requested):
Thanks for addressing the issue!

There are a few minor things that need to be changed; see my detailed comments.

Additionally, I notice that this PR is not only addressing #84, but it also addresses #85 to 50%. Since you have started it already, can you please address the remaining part of #85 as well within this PR ;-)

chengsijin0817:
> Thanks for addressing the issue!
> 
> There are a few minor things that need to be changed; see my detailed comments.

Fixed. Please check it.

> 
> Additionally, I notice that this PR is not only addressing #84, but it also addresses #85 to 50%. Since you have started it already, can you please address the remaining part of #85 as well within this PR ;-)

Issue #85 is also addressed whin this PR.

hartig (approved):
Perfect. Thanks!

PR #93, opened 2021-08-25
Rewriting rules and rule applications

chengsijin0817:


chengsijin0817:
In addition to the example rule "**RuleConvertTPAddIndexNLJToHashJoin**", I added four similar rewriting rules that convert TPAdd to Hash Join. 
Then, I will add rewriting rules that convert TPAdd to other types of Join algorithms, and add more other types of rewriting rules.

chengsijin0817:
@hartig I added a group of rewrite rules that can convert a (sub)plan with TPAdd as the root operator into a (sub)plan with join as the root operator.
It contains different variations of the physical algorithms for TPAdd and join operator, with a total of 15 rewriting rules.

For these rewriting rules, when the right-hand side is the same (i.e., the same physical algorithm for join), the plan with the TPAdd root operator can be rewritten in the same way (independent of the physical operator of TPAdd). Hence, I extracted three generic classes for simplification: **GenericRuleConvertTPAddToHashJoin**, **GenericRuleConvertTPAddToNaiveNLJ**, 
**GenericRuleConvertTPAddToSymmetricHashJoin**

Please check it and let me know if it makes sense or it is more readable to implement the specific version of **rewritePlan**() method in each rewrite rule.

hartig:
I have not looked at all of them in complete detail but, generally, it looks good.

However, there is only one thing that I wonder: It might be possible to capture multiple cases by a single class rather than one class per case. For instance, the subclasses of GenericRuleConvertTPAddToNaiveNLJ differ only in the physical operator that they expect for the tpAdd operator. In principle, the `canBeAppliedTo` method could be moved to the base class (GenericRuleConvertTPAddToNaiveNLJ) and changed such that it checks only whether the root operator is a tpAdd. Then, the base class could be made non-abstract and all the subclasses could be removed. Is there any disadvantage or any reason not to implement it this way? The only reason that I can see is that all the cases captured by such a class would have the same priority. However, they do have the same priority anyways, don't they? The advantage of implementing it this way would be that there will be much fewer rewriting rules that need to be considered (essentially, much fewer calls of the `canBeAppliedTo` method), plus fewer source code files to be maintained. What do you think?

chengsijin0817:
> I have not looked at all of them in complete detail but, generally, it looks good.
> 
> However, there is only one thing that I wonder: It might be possible to capture multiple cases by a single class rather than one class per case. For instance, the subclasses of GenericRuleConvertTPAddToNaiveNLJ differ only in the physical operator that they expect for the tpAdd operator. In principle, the `canBeAppliedTo` method could be moved to the base class (GenericRuleConvertTPAddToNaiveNLJ) and changed such that it checks only whether the root operator is a tpAdd. Then, the base class could be made non-abstract and all the subclasses could be removed. Is there any disadvantage or any reason not to implement it this way? The only reason that I can see is that all the cases captured by such a class would have the same priority. However, they do have the same priority anyways, don't they? The advantage of implementing it this way would be that there will be much fewer rewriting rules that need to be considered (essentially, much fewer calls of the `canBeAppliedTo` method), plus fewer source code files to be maintained. What do you think?

That's true. If it is not necessary to maintain a list of **specific** rules for each physical plan, then we do not have to create one class for each specific rule. In this way, the number of rewriting rules can be greatly reduced.
I can modify the implementation in this way.

chengsijin0817:
Hi @hartig , I have updated the rewriting rules of tpAdd. The number of rewriting rules for tpAdd is reduced to 15 after modification.
Currently, all rewriting rules for tpAdd have been added to this PR. Could you please check the implementation? If it looks okay, please approve and merge this PR.

hartig (approved):
Looks good now! I will merge this PR.

PR #102, opened 2021-09-23
Rewriting rules for join, union, mj, mu, and req

chengsijin0817:


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/rules/GraphPatternConstructor.java
```diff
@@ -0,0 +1,145 @@
...
+        final SPARQLQuery graphPattern2 = ((SPARQLRequest) lopReq2.getRequest()).getQuery();
+        final Element element2 = graphPattern2.asJenaQuery().getQueryPattern();
+
+        final ElementUnion elementUnion = new ElementUnion(element1);
+        elementUnion.addElement(element2);
+
+        return new SPARQLGraphPatternImpl(elementUnion);
+    }
+
+    protected static Set<TriplePattern> getTriplePatternsOfReq( final LogicalOpRequest<?, ?> lop ) {
```
hartig, line 119:
```suggestion
    public static Set<TriplePattern> getTriplePatternsOfReq( final LogicalOpRequest<?, ?> lop ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/rules/GraphPatternConstructor.java
```diff
@@ -0,0 +1,145 @@
...
+
+            if ( bgp.getTriplePatterns().size() == 0 ) {
+                throw new IllegalArgumentException( "the BGP is empty" );
+            }
+            else {
+                return new HashSet<>( bgp.getTriplePatterns() );
+            }
+        }
+        else  {
+            throw new IllegalArgumentException( "Cannot get triple patterns of the req" );
```
hartig, line 141:
```suggestion
            throw new IllegalArgumentException( "Cannot get triple patterns of the given request operator (type: " + req.getClass().getName() + ")." );
```
It may be helpful to have some more information if this exception will occur in some unexpected circumstances later ;-)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/rules/IdentifyLogicalOp.java
```diff
@@ -0,0 +1,50 @@
...
+        final LogicalOperator lop = ((PhysicalOperatorForLogicalOperator) op).getLogicalOperator();
+        return lop instanceof LogicalOpUnion;
+    }
+
+    public static boolean matchMultiwayJoin( final PhysicalOperator op ) {
+        final LogicalOperator lop = ((PhysicalOperatorForLogicalOperator) op).getLogicalOperator();
+        return lop instanceof LogicalOpMultiwayJoin;
+    }
+
+    public static boolean matchMultiwayUnion( final PhysicalOperator op ) {
```
hartig, line 36: I don't think the names of these functions are particularly suitable. I mean, if I would see them somewhere in the source code, I wouldn't immediately know what they are doing. Therefore, my proposal is to replace in these names the word "match" by the word "is"; e.g., `isTPAdd`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/rules/RuleChangeOrderOfSubPlansOfMultiwayJoin.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    protected boolean canBeAppliedTo( final PhysicalPlan plan ) {
+        final PhysicalOperator op = plan.getRootOperator();
+        if ( IdentifyLogicalOp.matchMultiwayJoin(op) ) {
+            for ( int i = 0; i < plan.numberOfSubPlans(); i ++ ) {
+                final PhysicalOperator subOp = plan.getSubPlan(i).getRootOperator();
+                if ( IdentifyLogicalOp.matchJoin(subOp) ) {
+                    return true;
+                }
+            }
```
hartig, line 27: What is this rule doing? According to its name, it changes the order of the subplans that are children of a multiway join. If that's the case, why do you need to check what type of root operator these subplans have?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/rules/RuleChangeOrderOfSubPlansOfMultiwayUnion.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    protected boolean canBeAppliedTo( final PhysicalPlan plan ) {
+        final PhysicalOperator op = plan.getRootOperator();
+        if ( IdentifyLogicalOp.matchMultiwayUnion(op) ) {
+            for ( int i = 0; i < plan.numberOfSubPlans(); i ++ ) {
+                final PhysicalOperator subOp = plan.getSubPlan(i).getRootOperator();
+                if ( IdentifyLogicalOp.matchUnion(subOp) ) {
+                    return true;
+                }
+            }
```
hartig, line 27: Similar question here. Why look at the subplans if the purpose of this rule is to simply change their order?

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/rules/RuleChangeOrderOfSubPlansOfMultiwayJoin.java
chengsijin0817: This rule does not change the order of sub-plans directly, the order of sub-plans of a multiway join is changed by rewriting the sub-plans (with a binary join as root). In another word, applying this rewriting rule will expand the child with binary join to two children of the multiway join.
Perhaps, I should rename this rewriting rule as RuleExpandBinaryJoinWithinMultiwayJoin. 
Does this name better explain the purpose of this rule? @hartig

hartig: So, this rule removes a child join of a multiway join by merging it directly into the multiway join? In this case, call the class RuleMegeChildJoinIntoMultiwayJoin.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/rewriting/rules/RuleChangeOrderOfSubPlansOfMultiwayUnion.java
hartig: RuleMergeChildUnionIntoMultiwayUnion

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
```diff
@@ -193,51 +193,24 @@ public static PhysicalPlan createPlan( final UnaryPhysicalOp rootOp,
...
-			return createPlanWithBindJoinVALUES( lop, subplan );
-		}
-		else {
-			return createPlan( lop, subplan );
+	 * The root operator uses the same physical algorithm as the given physical operator, which is usually a specific physical operator for bgpAdd.
+	 * The given subplan becomes the single child of the root operator.
+	 */
+	public static PhysicalPlan createPlanBasedOnTypeOfGivenPhysicalOp( final LogicalOpBGPAdd lop, final Class<? extends PhysicalOperator> opClass, final PhysicalPlan subplan ) {
+		switch ( opClass.getName() ) {
+			case "PhysicalOpIndexNestedLoopsJoin":
```
hartig, line 201: This may not work correctly in all cases; in particular, it doesn't cover cases in which `opClass` is a subclass of any of the classes recognized here (such as `PhysicalOpIndexNestedLoopsJoin`). To make this work for subclasses as well, I think you can use the `isAssignableFrom` method of `Class`-- try something along the following lines:
```java
if ( PhysicalOpIndexNestedLoopsJoin.class.isAssignableFrom(opClass) ) {
    return createPlanWithIndexNLJ( lop, subplan );
}
else if ...
```

...but please double check whether it really works that way.

hartig (changes requested):
Only one thing remaining.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
chengsijin0817: Thanks, I have updated this method using the approach you proposed. 
In addition, I created a unit test for rule 'RuleMergeTPAddOfBGPAddIntoBGPAdd' to check whether this method works as expected.

hartig (approved):

hartig:

PR #108, opened 2021-10-13
RandomizedJoinPlanOptimizerImpl

### LINKED ISSUE #104: Add RandomizedJoinPlanOptimizerImpl
Our SimpleJoinOrderingQueryOptimizer relies on a JoinPlanOptimizer. The only implementation that we currently have of the JoinPlanOptimizer interface is GreedyJoinPlanOptimizerImpl. The task now is to add another implementation, called RandomizedJoinPlanOptimizerImpl. What this implementation has to do is to simply create a randomly ordered join plan.

hartig:
This PR is meant to close #104

@PatchouliSenpai @SHSenpai  I have created this pull request (PR) to make it easier for me to look at and comment on the changes from within the github Web interface. Note that creating a PR does not mean you cannot make anymore changes. In contrast, if you push more commits to the branch, they will all become part of this PR.

### DIFF
--- a/.settings/org.eclipse.jdt.core.prefs
+++ b/.settings/org.eclipse.jdt.core.prefs
@@ -1,0 +1,8 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=11
+org.eclipse.jdt.core.compiler.compliance=11
+org.eclipse.jdt.core.compiler.problem.enablePreviewFeatures=disabled
+org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
+org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=ignore
+org.eclipse.jdt.core.compiler.release=enabled
+org.eclipse.jdt.core.compiler.source=11
--- a/.settings/org.eclipse.jdt.core.prefs
+++ b/.settings/org.eclipse.jdt.core.prefs
@@ -1,8 +1,0 @@
-eclipse.preferences.version=1
-org.eclipse.jdt.core.compiler.codegen.targetPlatform=11
-org.eclipse.jdt.core.compiler.compliance=11
-org.eclipse.jdt.core.compiler.problem.enablePreviewFeatures=disabled
-org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
-org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=ignore
-org.eclipse.jdt.core.compiler.release=enabled
-org.eclipse.jdt.core.compiler.source=11
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
@@ -25,7 +25,6 @@ public class RandomizedJoinPlanOptimizerImpl implements JoinPlanOptimizer {
 	 */
 	public RandomizedJoinPlanOptimizerImpl( final List<PhysicalPlan> subplans ) {
 		this.subplans = subplans;
-		this.random = new Random(); // Generates the random seed.
 	}
 	
 
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
@@ -1,0 +1,104 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.simple;
+
+import java.util.List;
+import java.util.Random;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryplan.utils.PhysicalPlanFactory;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+
+public class RandomizedJoinPlanOptimizerImpl extends JoinPlanOptimizerBase {
+	@Override // Implement initializeEnumerationAlgorithm from JoinPlanOptimizerBase.
+	public EnumerationAlgorithm initializeEnumerationAlgorithm( final List<PhysicalPlan> subplans ) {
+		return new RandomizedEnumerationAlgorithm(subplans);
+	}
+	
+	protected class RandomizedEnumerationAlgorithm implements EnumerationAlgorithm {
+		/**
+		 * Have a list of subplans.
+		 */
+		protected final List<PhysicalPlan> subplans;
+
+		/**
+		 *  When initializing the algorithm, start with the given list of subplans.
+		 *  @param subplans
+		 *  (These @ things where create automatically when I was making the comments, should I keep them in? // Simon)
+		 */
+		public RandomizedEnumerationAlgorithm( final List<PhysicalPlan> subplans ) {
+			this.subplans = subplans;
+		}
+		
+
+		/**
+		 * Also taken from Greedy and left unchanged save for the different method being called within.
+		 */
+		@Override
+		public PhysicalPlan getResultingPlan() throws QueryOptimizationException {
+			PhysicalPlan currentPlan = chooseFirstSubplan();
+
+			while ( subplans.size() > 0 ){
+				currentPlan = addNextRandomJoin(currentPlan);
+			}
+
+			return currentPlan;
+		}
+
+		
+		/**
+		 * Creating all possible plans, which are then to be picked from at random.
+		 * This method is directly copied from GreedyJoinPlanOptimizer without modification.
+		 * @param currentPlan
+		 * @return
+		 */
+		protected PhysicalPlan[] createNextPossiblePlans( final PhysicalPlan currentPlan ) {
+			final PhysicalPlan[] plans = new PhysicalPlan[ subplans.size() ];
+			for ( int i = 0; i < subplans.size(); ++i ) {
+				plans[i] = PhysicalPlanFactory.createPlanWithJoin( currentPlan, subplans.get(i) );
+			}
+			return plans;
+		}
+		
+
+		/**
+		 * Chooses the first subplan.
+		 * Instead of the best plan chosen by Greedy, this method picks a completely random plan.
+		 * @return
+		 * @throws QueryOptimizationException
+		 */
+		protected PhysicalPlan chooseFirstSubplan() throws QueryOptimizationException {
+			Random random = new Random(); // Create a random seed.
+			int indexOfRandomPlan = random.nextInt(subplans.size()); // nextInt returns an int between 0 (inclusive) and size (exclusive).
+
+			final PhysicalPlan randomPlan = subplans.get(indexOfRandomPlan);
+			subplans.remove(indexOfRandomPlan);
+			return randomPlan;
+		}
+		
+
+		/**
+		 * Next random join.
+		 * @param currentPlan
+		 * @return
+		 * @throws QueryOptimizationException
+		 */
+		protected PhysicalPlan addNextRandomJoin( final PhysicalPlan currentPlan )
+				throws QueryOptimizationException
+		{
+			final PhysicalPlan[] nextPossiblePlans = createNextPossiblePlans(currentPlan);
+
+			Random random = new Random(); // Create a random seed.
+			// We're creating another random seed...
+			// ...but would it be better performance-wise to store the random seed in the class?
+			// This would allow us to only have to generate it once, with the downside of it taking up space.
+			// I would probably opt for putting it as part of the class, because processor time is expensive and storage is cheap. // Simon
+
+			int indexOfRandomPlan = random.nextInt(subplans.size());
+
+			subplans.remove(indexOfRandomPlan);
+			return nextPossiblePlans[indexOfRandomPlan];
+		}
+		
+	}
+	
+	
+}
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
@@ -7,77 +7,44 @@ import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
 import se.liu.ida.hefquin.engine.queryplan.utils.PhysicalPlanFactory;
 import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
 
-public class RandomizedJoinPlanOptimizerImpl extends JoinPlanOptimizerBase {
-	@Override // Implement initializeEnumerationAlgorithm from JoinPlanOptimizerBase.
-	public EnumerationAlgorithm initializeEnumerationAlgorithm( final List<PhysicalPlan> subplans ) {
-		return new RandomizedEnumerationAlgorithm(subplans);
+public class RandomizedJoinPlanOptimizerImpl implements JoinPlanOptimizer {
+	
+	/**
+	 * The class contains a random seed object.
+	 * This cannot be marked final (compilation error if so).
+	 */
+	protected Random random = new Random();
+	
+	/**
+	 * Have a list of subplans.
+	 */
+	protected final List<PhysicalPlan> subplans;
+	
+	/**
+	 *  When initializing the algorithm, start with the given list of subplans.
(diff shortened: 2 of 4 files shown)


hartig:
By the way, which is your actual github account now??

SHSenpai:
The account which made the commit uses my old email address it seems, I must've been using it back in high school for something.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,104 @@
...
+		/**
+		 * Next random join.
+		 * @param currentPlan
+		 * @return
+		 * @throws QueryOptimizationException
+		 */
+		protected PhysicalPlan addNextRandomJoin( final PhysicalPlan currentPlan )
+				throws QueryOptimizationException
+		{
+			final PhysicalPlan[] nextPossiblePlans = createNextPossiblePlans(currentPlan);
```
hartig, line 87: For this randomized JoinPlanOptimizer, there is no need to create all next possible plans at each step. Instead, just randomly pick one of the remaining subplans and create the next join plan with this one only (in the way it is done within the loop in `createNextPossiblePlans`).

hartig:
Thanks for the clarification about the accounts. I wonder how this old account was even able to push to the repo :-/  Anyways, I am assuming that you will only use @SHSenpai from now on.

.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,8 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=11
+org.eclipse.jdt.core.compiler.compliance=11
+org.eclipse.jdt.core.compiler.problem.enablePreviewFeatures=disabled
+org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
+org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=ignore
+org.eclipse.jdt.core.compiler.release=enabled
+org.eclipse.jdt.core.compiler.source=11
```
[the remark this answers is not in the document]
SHSenpai: I didn't even know about these files existing. Is there a way to make this be ignored with the commit, and how should I go about removing them? Is it to be removed from the commit, the branch, or deleted directly from my PC?

SHSenpai:
I tried to change my git credentials to push from this account, and it seemingly ended up pushing from no account. Very strange, and I don't know how to solve it. I'm going to do some more troubleshooting so that the next push is hopefully from this account >.>

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+	public EnumerationAlgorithm initializeEnumerationAlgorithm( final List<PhysicalPlan> subplans ) {
+		return new RandomizedEnumerationAlgorithm(subplans);
+	}
+	
+	protected class RandomizedEnumerationAlgorithm implements EnumerationAlgorithm {
+		
+		/**
+		 * The class contains a random seed object.
+		 */
+		Random random;
```
hartig, line 21: A few points here:

1. Member variables such as this one should typically be `protected`, which makes it possible to access them from within potential future subclasses.
2. The member variable should be `final`. This way, after it is initialized (e.g., in the constructor), we cannot assign something else to it later. The idea is that we do not intend to later override the value of this member variable with another `Random` object because this way. The keyword `final` allows us to make this intention explicit and to get a compiler error in case we do anything unexpected later. So, using `final` is mainly a safety measure and it is generally a good practice to use it.
3. For member variables that do not need anything from the constructor to be initialized, it is preferable to include their initialization already in the declaration rather than in the constructor.

So, based on all of these points, this line should be changed as follows:
```suggestion
		protected final Random random = new Random();
```
By the way, notice that you can accept these code change suggestions directly here in the PR, in which case a commit is created automatically. The only thing that you need to be mindful about if you do so is that, if you later also want to do more code changes in your IDE, you have to do a `git pull` first in order to update your local branch to include these commits.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.List;
+import java.util.Random;
+
+import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
+import se.liu.ida.hefquin.engine.queryplan.utils.PhysicalPlanFactory;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+
+public class RandomizedJoinPlanOptimizerImpl extends JoinPlanOptimizerBase {
+	@Override // Implement initializeEnumerationAlgorithm from JoinPlanOptimizerBase.
```
hartig, line 11: There is no need for such comments ;-)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+		 */
+		public RandomizedEnumerationAlgorithm( final List<PhysicalPlan> subplans ) {
+			this.subplans = subplans;
+			this.random = new Random(); // Generates the random seed.
+		}
+		
+
+		/**
+		 * Also taken from Greedy and left unchanged save for the different method being called within.
+		 */
```
hartig, line 40: This comment does not make much sense in the context of this class. I understand that you added it as a way of documenting what you have done. However, before merging this PR into the main branch, such comments should be removed. Someone who looks at this code later may not know the context in which you made this comment and, then, the comment might cause confusion.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+			return currentPlan;
+		}
+		
+
+		/**
+		 * Chooses the first subplan.
+		 * Instead of the best plan chosen by Greedy, this method picks a completely random plan.
+		 * @return
+		 * @throws QueryOptimizationException
```
hartig, line 57:
```suggestion
		 * Returns one of the subplans which is picked at random.
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+		 */
+		protected PhysicalPlan chooseFirstSubplan() throws QueryOptimizationException {
+			int indexOfRandomPlan = this.random.nextInt(subplans.size()); // nextInt returns an int between 0 (inclusive) and size (exclusive).
```
hartig, line 60: Two things here:
1. When you are referring to member variables, you don't need to prefix the member variables with `this.`. While doing so is possible, it is not a common practice in Java.
2. What I mentioned above about `final` also holds for local variables such as this one. Unless a local variable is explicitly meant to be reassigned within the method, it should be marked `final` as a safety measure.

Hence, the line should be changed as follows:
```suggestion
			final int indexOfRandomPlan = random.nextInt(subplans.size()); // nextInt returns an int between 0 (inclusive) and size (exclusive).
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+			return randomPlan;
+		}
+		
+
+		/**
+		 * Next random join.
+		 * @param currentPlan
+		 * @return
+		 * @throws QueryOptimizationException
+		 */
```
hartig, line 73: Remove. It does not say much anyways.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+		protected PhysicalPlan addNextRandomJoin( final PhysicalPlan currentPlan )
+				throws QueryOptimizationException
+		{
+
+			int indexOfRandomPlan = this.random.nextInt(subplans.size());
```
hartig, line 78:
```suggestion
			final int indexOfRandomPlan = random.nextInt(subplans.size());
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+			final PhysicalPlan nextPlan = PhysicalPlanFactory.createPlanWithJoin( currentPlan, subplans.get(indexOfRandomPlan));
+			// Instead of creating a whole array of plans, create one.
+			
+			subplans.remove(indexOfRandomPlan);
+			// Remove the old subplan.
+			
+			return nextPlan; // Return the created plan.
```
hartig, line 85: While I appreciate your effort of including comments, these three should be removed. The do not really add anything that is not already clear from the code itself. In fact, the first one may even introduce confusion for anyone other than you and me. Also, putting the comments in the line below the statement that they are talking about is strange; if there are comments inside code blocks, they should be in the line before the code statements that they are talking about (or directly after the statements on the same line, as you have done for the return statement here).
```suggestion
			subplans.remove(indexOfRandomPlan);			
			return nextPlan;
```

hartig (commented):
Here are more comments.

Additionally, as I already mentioned before, for this whole class it would be better to make it directly implement `JoinPlanOptimizer` instead of making it a subclass of `JoinPlanOptimizerBase`; all the code can be merged into the `determineJoinPlan` method. Please try that.

SHSenpai:
@hartig 
I wasn't able to request you for a review for some reason (your name didn't show up in the menu), so I'm tagging you here instead.
The making of the random object into a final type did not work. I believe this is because of the random object having its state modified each time a random number is drawn from it, so that the next random number will be different. I changed it back to be non-final.

This push was more removing than adding, but returning to this algorithm after having gotten more Java proficiency in IterativeImprovement made me see how inefficient my first implementation was. This new and improved version should be much more minimalist both in amount of code and in running time.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+import se.liu.ida.hefquin.engine.queryplan.utils.PhysicalPlanFactory;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+
+public class RandomizedJoinPlanOptimizerImpl implements JoinPlanOptimizer {
+	
+	/**
+	 * The class contains a random seed object.
+	 * This cannot be marked final (compilation error if so).
+	 */
+	protected Random random = new Random();
```
hartig, line 16:
```suggestion
	 */
	protected final Random random = new Random();
```

You can make this member variable `final`. Your mistake was that, after assigning something to it here, you tried to reassign something else again in the constructor. That's not possible with `final`. In contrast, every variable marked as `final` can be assigned only once---that's the meaning of the `final` keyword in Java.

Notice that this is different from `const` in C++.  `final` in Java is only about the variable, not about the object assigned to the variable. Hence, it is still possible to call member functions that may change the internal state of the object.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+	 */
+	protected final List<PhysicalPlan> subplans;
+	
+	/**
+	 *  When initializing the algorithm, start with the given list of subplans.
+	 *  @param subplans
+	 */
+	public RandomizedJoinPlanOptimizerImpl( final List<PhysicalPlan> subplans ) {
+		this.subplans = subplans;
+		this.random = new Random(); // Generates the random seed.
```
hartig, line 29:
```suggestion
```

That's the part that is not needed because you are already assigning this member variable in its declaration.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+	/**
+	 * The class contains a random seed object.
+	 * This cannot be marked final (compilation error if so).
+	 */
+	protected Random random = new Random();
+	
+	/**
+	 * Have a list of subplans.
+	 */
+	protected final List<PhysicalPlan> subplans;
```
hartig, line 21: You should not have this member variable. The class should be stateless. In fact, you are not even using this member variable in the `determineJoinPlan` function (because there you are using the function argument with the same name). Hence, please remove this member variable, which means that you also need to remove it from the constructor.

After removing it from the constructor, and removing the part about `this.random` from the constructor as well (see my comment there), the whole constructor becomes empty and, thus, can and should be removed altogether.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+	 *  @param subplans
+	 */
+	public RandomizedJoinPlanOptimizerImpl( final List<PhysicalPlan> subplans ) {
+		this.subplans = subplans;
+		this.random = new Random(); // Generates the random seed.
+	}
+	
+
+	@Override
+	public PhysicalPlan determineJoinPlan(List<PhysicalPlan> subplans) throws QueryOptimizationException {
```
hartig, line 34:
```suggestion
	public PhysicalPlan determineJoinPlan( final List<PhysicalPlan> subplans ) throws QueryOptimizationException {

```

hartig (commented):
Almost done! There are just a few smaller things that need to be addressed. See my comments.

(By the way, the reason why you cannot request my review in this PR is because I was the one who created this first PR for you ;)

SHSenpai:
@hartig Thanks for clearing up 'final', I thought that it was the same as C++'s 'const'.
I thought about whether we would ever use the same RandomizedJoinPlanOptimizerImpl more than once and whether the random object should still be a member, but we can and it should as far as I can think - it should make it more random and also save a little bit of time if the same object is used every iterative loop in the other algorithm for example.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,36 @@
...
+	/**
+	 * The class contains a random seed object.
+	 */
+	protected final Random random = new Random();
+	
+	
+	@Override
+	public PhysicalPlan determineJoinPlan( final List<PhysicalPlan> subplans ) throws QueryOptimizationException {
+
+
```
hartig, line 21:
```suggestion
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/RandomizedJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+
+public class RandomizedJoinPlanOptimizerImpl implements JoinPlanOptimizer {
+	
+	/**
+	 * The class contains a random seed object.
+	 */
+	protected final Random random = new Random();
+	
+
```
hartig, line 17:
```suggestion
```

hartig:
This PR looks good now and I will merge it. Afterwards, I will also delete the corresponding code branch.

PR #109, opened 2021-10-14
added abstract objects for cypher query modelling

scferrada:
This is the first of a series of smaller PRs to break down #61 

In this one, I add all the abstract objects needed to model Cypher Queries.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/query/CypherQuery.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/query/CypherQuery.java
@@ -4,7 +4,19 @@ import se.liu.ida.hefquin.engine.query.cypher.CypherVar;
 
 import java.util.Set;
 
+/**
+ * Represents a Query written in the Cypher Query Language
+ */
 public interface CypherQuery extends Query {
+    /**
+     * Creates the string representation of the modelled Cypher query
+     * @return a Cypher query string
+     */
     String toString();
+
+    /**
+     * Obtains the set of variables that are defined in the MATCH clauses of the query
+     * @return a Set of CypherVar objects
+     */
     Set<CypherVar> getMatchVars();
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/query/CypherMatchQuery.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/query/CypherMatchQuery.java
@@ -6,10 +6,38 @@ import se.liu.ida.hefquin.engine.query.cypher.WhereCondition;
 
 import java.util.List;
 
+/**
+ * Represents a Cypher Query with a MATCH-WHERE-RETURN structure,
+ * For example, the query:
+ * MATCH (x)
+ * MATCH (a)-[b]->(c)
+ * WHERE a:CLASS AND b.property='value'
+ * RETURN x AS n1, c AS n2
+ *
+ * can be represented with this interface.
+ */
 public interface CypherMatchQuery extends CypherQuery {
 
+    /**
+     * Obtains the list of the patterns to be matched in the query. For the example query,
+     * this method returns a list with 2 elements, one representing each MATCH statement.
+     * @return a list of MatchClause objects
+     */
     List<MatchClause> getMatches();
+
+    /**
+     * Obtains a list of conditions, such that their conjunction represents the condition imposed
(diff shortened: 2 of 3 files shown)


src/main/java/se/liu/ida/hefquin/engine/query/CypherQuery.java
```diff
@@ -0,0 +1,34 @@
...
+import se.liu.ida.hefquin.engine.query.cypher.CypherVar;
+import se.liu.ida.hefquin.engine.query.cypher.MatchClause;
+import se.liu.ida.hefquin.engine.query.cypher.ReturnStatement;
+import se.liu.ida.hefquin.engine.query.cypher.WhereCondition;
+import se.liu.ida.hefquin.engine.query.impl.MatchCypherQuery;
+import se.liu.ida.hefquin.engine.query.impl.UnionCypherQuery;
+
+import java.util.Set;
+
+public interface CypherQuery {
```
hartig, line 12:
```suggestion
public interface CypherQuery extends Query {
```

src/main/java/se/liu/ida/hefquin/engine/query/CypherQuery.java
```diff
@@ -0,0 +1,34 @@
...
+    String toString();
+
+    Set<MatchClause> getMatches();
+    Set<WhereCondition> getConditions();
+    Set<ReturnStatement> getReturnExprs();
+    Set<CypherQuery> getUnion();
+    Set<CypherQuery> getIntersect();
```
hartig, line 19: Two comments about this:
1. I understand that, for each of these, the order of the respective elements does not make a difference in terms of query semantics. However, I would consider these classes and interfaces in this PR as elements of ASTs of Cypher queries, and there everything would have an order. Additionally, this order might make a difference in terms of performance. Therefore, I suggest to replace these sets by lists.
2. You are trying to capture three different types of Cypher queries with this one interface; as a consequence, you have a mixture of methods that are not all relevant for all of these types. This is not such a good design. Please create three separate subinterfaces where each of them extends `CypherQuery` and captures only one of the three types (e.g., `CypherMatchQuery`, `CypherUnionQuery`).

src/main/java/se/liu/ida/hefquin/engine/query/CypherQuery.java
```diff
@@ -0,0 +1,34 @@
...
+
+    void addMatchClause(final MatchClause match);
+    void addConditionConjunction(final WhereCondition cond);
+    void addReturnClause(final ReturnStatement ret);
+    void addQueryToUnion(final CypherQuery q);
```
hartig, line 24: I would like our main interfaces to represent immutable objects. Is it really necessary to be able to add elements to a `CypherQuery` object (as would be possible by using these methods)? Or can your code that assembles the queries actually create the complete `CypherQuery` object after having assembled all its elements (or, alternatively, use a constructor that produces a new `CypherQuery` object by merging two given ones)? I hope so; in which case I suggest to remove these methods from the interface.

src/main/java/se/liu/ida/hefquin/engine/query/CypherQuery.java
```diff
@@ -0,0 +1,34 @@
...
+
+    boolean isMatchQuery();
+    boolean isUnionQuery();
```
hartig, line 27: See my comment above; it is better to create separate subinterfaces for the different types of Cypher queries.

src/main/java/se/liu/ida/hefquin/engine/query/CypherQuery.java
```diff
@@ -0,0 +1,34 @@
...
+
+    CypherQuery combineWith(final CypherQuery query);
+    CypherQuery combineWithMatch(final MatchCypherQuery query);
+    CypherQuery combineWithUnion(final UnionCypherQuery query);
```
hartig, line 31: See my comment above about immutability. I would prefer to not have such methods in the interface. Helper functionality for merging two Cypher queries can be provided in a Utils class, or as part of alternative constructors in the classes that implement this interface (or the suggested subinterfaces).

hartig:
Just found a bit more time ;-)  ...so, I added some more comments. I am done for now. Please request my review again once you have addressed the comments and, of course, feel free to raise questions or concerns regarding my comments.

PR #114, opened 2021-10-20
Termination criteria

chengsijin0817:
Add another five termination criteria


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/TerminateByDistancePercAvg.java
```diff
@@ -0,0 +1,53 @@
...
+            return false;
+        }
+
+        double avgCurrentCost = currentGeneration.avgCost;
+        double avgPreviousCost = allPreviousGenerations.get( previousGenerationNr-1 ).avgCost;
+        double relDistance = ( avgPreviousCost - avgCurrentCost ) / avgPreviousCost;
+
+        if ( relDistance > percAvgThreshold ) {
+            return false;
+        }
```
hartig, line 35: I think you can remove these lines and adjust the loop in a way such it takes care of this first step as well. That way, you are not repeating the same code twice.

The same comment may apply to the other termination criteria as well.

PR #115, opened 2021-10-21
Stable point for IterativeImprovement. Not blocked and not finished.

SHSenpai:
Context, Logical-to-physical converter, as well as CostModel are made to be member variables as I can only assume that one instance of the optimization algorithm should be useable for more than one logical plan.


SHSenpai:
Creating a list(?) of items out of iterating through the returs of getResultingPlan() calls was my intention for what to do if no method already existed, so it's nice to see that I was thinking in the right realm.
(?) What is the best datatype in Java for creating a set like this where order isn't important?

The inner loop will be done soon unless I accidentally stumble on some error, but as the paper says that the outer loop is to run unless a stopping condition is met - where do I take the stopping condition? I don't see any such thing included as a parameter for the function which is to be implemented - at least QueryOptimizerContext does not contain any such member as far as I can see.

SHSenpai:
Considerable progress made, with lots of questions that I've thought of. All of them that haven't already been mentioned in the previous post are added as comments in the code.  I'm still able to continue, but this felt like a natural stopping point for today.

hartig:
> What is the best datatype in Java for creating a set like this where order isn't important?

As you can see in the return type of the `getRuleApplications` method of the `RuleApplicationsOfPlans` class mentioned in my previous comment, the datatype for sets in Java is the `Set` interface.

By using the templating mechanism in Java (which is actually called "generics" in Java, but it's very similar to class templates in C++), you can specify what types of objects such a `Set` can contain. For instance, `Set<RuleApplication>` is a set of `RuleApplication` objects, and `Set<PhysicalPlan>` is a set of `PhysicalPlan` objects.

While `Set` is an interface, a concrete class that implements this interface and that is typically used when an actual `Set` object needs to be created is `HashSet`.

> The inner loop [...]
> [...] the paper says that the outer loop [...]
> [...] where do I take the stopping condition?

I am not exactly sure what you mean by all this. Are these things that are mentioned in the paper? If so, please be a bit more explicit about that; don't assume that I know the paper inside out ;-)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+	}
+	
+	/**
+	 * Currently created as a member method, but I might end up moving this to somewhere else if I find that it is needed in more places.
+	 * That obviously depends on whether Java supports importing methods outside of objects.
+	 * In C++, I would be passing references here - is something similar possible/appropriate in Java?
+	 * If significant performance is lost by making this a separate function, I'll be baking it into optimize. I've just been taught so far at LiU that making separate smaller functions is preferable so that the code becomes a bit more readable.
+	 * @return
+	 */
+	protected List<PhysicalPlan> getNeighbours(final PhysicalPlan initialPlan, final RewritingRule rewritingRule) {
```
hartig, line 86: What is the purpose of this method? To get all neighbors that the given plan has in terms of the given rewriting rule? I thought that you actually want all possible neighbors, across all rewriting rules.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+		List<PhysicalPlan> resultList = new ArrayList<PhysicalPlan>(); // Create an empty list. Just List<PhysicalPlan> didn't work so I had to look this up.
+		
+		Set<RuleApplication> ruleApplications = rewritingRule.determineAllPossibleApplications(initialPlan);
+		
+		// Assuming that forEach is the best way to iterate through a set, and that doing it is appropriate. Converting it to an array seems patently wasteful.
+		ruleApplications.forEach((e) -> {
+			resultList.add(e.getResultingPlan());
+		});
```
hartig, line 94: Using `forEach` with the given lambda expression is not wrong. Such lambda expressions have been introduced in some later version of Java to enable Java developers to do functional programming. So far, we have not used such a functional programming style in our code; instead, we have used the more traditional imperative style. I would prefer we keep it that way instead of starting to mix. Therefore, the more traditional way to write this loop here is the following:
```suggestion
		for ( final RuleApplication ra : ruleApplications ) {
			resultList.add( ra.getResultingPlan() );
		}
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+		// localMinimum = best cost among all neighbours
+		// optimalCost = best possible cost out of everything that the algorithm has found.
+		
+		// Optimization takes place here.
+		// Given the one starting state, find a local minimum.
+		// Firstly, generate neighbours. (Come to think of it, should I use the British or American spelling of neighbours?)
+		RewritingRule rwRule = null; // Eclipse suggested initiating this by adding the "= null". This doesn't seem quite right...
+		List<PhysicalPlan> neighbours = getNeighbours(optimalPlan,rwRule);
+		
+		List<CompletableFuture<Double>> futureNeighbourCosts = null; // Again, null suggested by Eclipse. Is this right?
```
hartig, line 62: No, in this case you certainly do not want to assign `null` to this variable. Instead, you want to create an actual list and assign that one. To this end, you can do:
```suggestion
		List<CompletableFuture<Double>> futureNeighbourCosts = new ArrayList<>();
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+		List<Double> neighbourCosts;
```
hartig, line 63: This variable is also initialized (implicitly) with the value `null` rather than with an actual list object.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+		
+		while(true) {
+			for(int i = 0; i < neighbours.size(); i++ ) {
```
hartig, line 66: You can write for-loops like that. If you do not actually need the running index variable (i.e., `i` in this case), then you can also simplify this as follows:
```java
for ( final PhysicalPlan p : neighbours ) {
     ...
}
```
...and then you can use `p` directly within the body of the loop, rather than having to do `neighbours.get(i)`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+				// Here, I want to get an iterator for the first element of the list, and iterate through the list and get initiateCostEstimations for each one, filling up the futureNeighbourCosts
```
hartig, line 67: I guess that, after getting all the completable futures for the neighbors, you would then also want to resolve them in order to get the actual cost values. We already have a helper function for that: You can use `getEstimates` of the helper class `CostEstimationUtils` which you find in the package `se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils`.

hartig (commented):
Here are my detailed comments for your current code.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+	public IterativeImprovementBasedQueryOptimizer (final QueryOptimizationContext ctxt) {
+		assert ctxt != null;
+		context = ctxt;
+	}
+	
+	
+	@Override
+	public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException {
+
+		converter = context.getLogicalToPhysicalPlanConverter();
```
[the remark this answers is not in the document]
SHSenpai: Oh whoops, seems like I forgot to pull after committing those things from your previous suggestion.

SHSenpai:
@hartig 
Whoops, I'm so used to communicating with professors about the courses where they're all very familiar with the entirety of the course material. Here's the short explanation on what I mean with the outer and inner loop.

The Iterative Improvement algorithm works in two loops, where the outer loop starts by picking out a random plan and then runs the inner loop until the local minimum of that plan is found. I.e. it looks through all the neighbours, then all neighbours of the best neighbours, and so on until it has arrived at a plan which does not have a neighbour with a lower cost.
However, since the local minimum cost (no *adjacent* plans are cheaper) aren't necessarily the global minimum cost, the outer loop then repeats, picking another random plan and doing the whole thing over again, then comparing this plan to the previous found minimum to see if this one is better. This is carried out perpetually in the algorithm as the paper lays it out, so the algorithm assumes that there will be some sort of stopping condition that eventually breaks this outer loop. This is because mathematics dictate that, given that the query is large enough, the amount of possible query plans will grow exponentially so it becomes completely unfeasible to compare all of them. The algorithm will run for a 'long enough' time to produce a 'good enough' result, as we will not want the optimization algorithm to take so long that any speed saved by the optimization is cancelled out by how long the algorithm takes.

My short-term plan is to have the outer loop run a constant amount of times, such as 5, for now.

hartig:
Thanks for the detailed explanation. Very helpful!

Now I can respond to your earlier question about whether we have code for this stopping condition: Since this stopping condition is a concept of the II algorithm, we do not have anything to cover it as a general thing. Instead, it needs to be implemented as part of the implementation of the algorithm. While I am okay with your idea to use a fixed number of iteration steps for this purpose, can you please not implement it by hard coding it directly into the algorithm but, instead, by means of a dedicated Java interface for which there is on implementing class that then captures your stopping condition. That way we later can implement other possible conditions and use them as a plugin in the algorithm. For inspiration of how this interface may look like, you can take a look at the TerminationCriterion interface that we have for the evolutionary algorithm.

I also have another thing that I wonder about after reading your explanation: you write that the outer loop randomly picks plans. Do you have an idea how the authors of the paper had considered that this will be done?

SHSenpai:
@hartig
I'll take a look at the TerminationCriterion interface. As I understand it though, this means that this algorithm will no longer be an implementation of the QueryOptimizer.optimize method, is that right?

The paper has no implementation listed, and just says that it picks a random state (state = physical plan) and then goes for random downhill moves (in this context a downhill move is a singular ruleapplication resulting in a cheaper plan) until it has reached a local minimum.

For picking a random starting point for each iteration, my first idea would be to carry out X amounts of *random* moves from the initial plan, although this is obviously not truly random although it's the best idea I have right now.
For making downhill moves, I'm currently thinking about whether to
1) Select all possible moves which would be downhill and choosing a random one among them,
2) Putting all possible moves into a non-strict-order container such as a set, and choosing the first arbitrary one which is downhill. This may not be fully random, but should be faster than option 1.
3) Rather than selecting a random downhill move, select the very cheapest of all possible downhill moves. This would deviate from how the paper states it, but it may be faster depending on the cost of random selection compared to the cost of remembering which the best one is and comparing other moves to it.

hartig:
> I'll take a look at the TerminationCriterion interface. As I understand it though, this means that this algorithm will no longer be an implementation of the QueryOptimizer.optimize method, is that right?

No that's not right. Your `IterativeImprovementBasedQueryOptimizer` needs to implement `QueryOptimizer`. The stopping condition can be passed to the constructor of `IterativeImprovementBasedQueryOptimizer`. You can take a look at `EvolutionaryAlgorithmQueryOptimizer` to see how the `TerminationCriterion` criterion is used there.

> For picking a random starting point for each iteration, my first idea would be to carry out X amounts of random moves from the initial plan, although this is obviously not truly random

You are right, it would not be truly random, but I also do not have any other idea. Perhaps, instead of always starting from the initial plan, you may start from the plan that was the starting point for the previous iteration. That way, they algorithm may get a bit more diversity in terms of what starting points it may consider.

Please make X an argument that can be passed to he constructor of `IterativeImprovementBasedQueryOptimizer`.

Regarding the downhill moves, I think that your third option is reasonable.

SHSenpai:
Additionally, how do I resolve the big red 'Changes requested'? From what I can see, those changes are marked outdated.

hartig:
Here is a link to the paper (so that I do not always have to search for it again ;-)
https://dl.acm.org/doi/10.1145/93597.98740

hartig:
> The number of iterations have been made as part of the constructor to the optimizer.

While it works like that, this solution was not what I had in mind. In contrast, as described in [my earlier comment above](https://github.com/LiUSemWeb/HeFQUIN/pull/115#issuecomment-957987094), please create a Java interface that captures the notion of a stopping condition. You may call this interface `StoppingConditionForIterativeImprovement`. This interface needs to have a method that you may call `readyToStop` and that has to return a `boolean` and it should have relevant parameters. This interface would be very similar to the interface called `TerminationCriterion`.

Once you have create the interface, change the constructor of your `IterativeImprovementBasedQueryOptimizer`. The argument of that constructor needs to be such a `StoppingConditionForIterativeImprovement` object.

Finally, you can create a class that implements your `StoppingConditionForIterativeImprovement` interface and that captures the specific stopping condition to stop after a fixed number of iteration steps. We have a related implementation of the `TerminationCriterion` interface which is called `TerminatedByNumberOfGenerations`.

> It feels a little inefficient to do it like this (what if we want to run X optimization iterations on plan A but Y optimization iterations on plan B, but both within the same context?) but it's what seems to work and it is what you requested.

Yes, and by capturing the actual stopping condition as a separate Java interface, we can later implement other possible stopping conditions.

> The downhill move now picks the 'most downhill' move. It's a bit of a deviation from what the algorithm paper says, but for now it seems sort-of sensible in the absence of other good options. That being said, I'm a bit worried about there being 'false bottoms' where a single big downhill move ending up in a local minimum, where a smaller alternative downhill move would open up for more downhill moves later which would make the end result be lower cost.

That's an excellent observation. I wasn't thinking of this when I replied to your question related to the three options for implementing the downhill moves. Looking again at the paper now, and considering what you say here, I now think that the first of your three options is actually the right choice. Please implement that one (and sorry for wasting your time with asking for the third option before!)

(I have teaching now and will get back to the rest of your questions/comments later)

hartig:
Now to the rest of your comments/questions:

> The only thing left to do before I'd consider this algorithm to be run-ready is to get random starting points. [...] should I also make the amount of random jumps into a constructor value so that an arbitrary amount of random jumps isn't hard-coded into the algorithm?

Yes please.

> Regardless, I think that I should possibly look into finding an efficient way to randomize a physical plan once prototypes of these algorithms have all been finished - unless this is accomplished by using the yet-to-be-finished randomizer that I started working on in order to get my fingers warm with the project.

Great idea. It didn't cross my mind that we may use that random query plan generator in your other PR as a basis for creating the random starting points for the iterative-improvement algorithm. Let's do that!

> Additionally, how do I resolve the big red 'Changes requested'? From what I can see, those changes are marked outdated.

This can be done only by the person who has created the corresponding code review (me, in this case). Once I do another code review now that the code has changed, the result of that new code review will override the result from my previous review.

hartig (changes requested):
I just wanted to create a new code review, but then I saw that there are still lots of comments from my previous review that have not been addressed yet. So, I have added only two new comments for now.

Let me know when the code is ready to be reviewed again.

SHSenpai:
As part of reading the TerminatedByNumberOfGenerations I've discovered the PhysicalPlanWithCost datatype. I wonder if it would be cheaper to use this than to store the costs in a separate list. I've been thinking a bit about how to best associate a physical plan with a *future* cost too.

I created a stopping condition interface and a class which implements it. I decided to keep track of the number of iterations in the stopping condition class in order to minimize needless computation. If a future stopping condition does not require keeping track of the number of generations (such as a condition based on real time passed!) then there's no need to do so. This saves a few machine instructions every outer loop compared to making the algorithm keep track of its generation. Additionally, the readyToStop() can and does run without parameters, which should also save a few instructions every time it's called compared to having to push parameters from the algorithm when the method is called.

Since we will be using the Randomized implementation in order to create random starting points, I've decided that it would just be wasted work to implement a stop-gap solution in the meantime as it would ultimately just be replaced by the call to Randomized. Speaking of it, it doesn't exist in this branch. Should we merge this branch with the Randomized branch and consider work on the Randomized plan generator to be part of creating this algorithm?

Regardless, the next thing I intend to do is replace the 'largest downward move' picking process with a 'random downward move' picking process. Which class/method is the most efficient in Java for selecting a random object out of a container? Also, does Java have pointers/references? I'm thinking about creating a container and filling it with all neighbours which constitute a downward move, and then picking a random one out of those. This container will not have to remember what they cost since all lower costs are equally acceptable. In C++, this is where I would have it be a container of pointers/references in order to reduce data duplication.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,135 @@
...
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizer;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.CostModel;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationContext;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RewritingRule;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RuleApplication;
+
+public class IterativeImprovementBasedQueryOptimizer implements QueryOptimizer {
+	
+	protected final QueryOptimizationContext context;
+	StoppingConditionForIterativeImprovement condition;
```
hartig, line 21:
```suggestion
	protected final StoppingConditionForIterativeImprovement condition;

```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,135 @@
...
+	
+	public IterativeImprovementBasedQueryOptimizer (final QueryOptimizationContext ctxt, StoppingConditionForIterativeImprovement x) {
+		assert ctxt != null;
```
hartig, line 24:
```suggestion
	public IterativeImprovementBasedQueryOptimizer (final QueryOptimizationContext ctxt, final StoppingConditionForIterativeImprovement x) {
		assert ctxt != null;
		assert x != null;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/StoppingConditionByNumberOfGenerations.java
```diff
@@ -0,0 +1,26 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
+
+public class StoppingConditionByNumberOfGenerations implements StoppingConditionForIterativeImprovement{
+	
+    protected final int generationThreshold;
+    protected int generations;
```
hartig, line 6: This should not be a member variable in this class because any object of this class should be stateless so that we can use just one such object, exactly like we are using only one `IterativeImprovementBasedQueryOptimizer` object. In other words, calling the `readyToStop` method should not change the state of this object because we may (re)use this object for multiple queries to be optimized (perhaps even in parallel).

Therefore, after removing this member variable here, you need to extend the signature of the `readyToStop` method (not only here but also in the interface) by adding a parameter via which the current generation counter can be passed:
```java
public interface StoppingConditionForIterativeImprovement {
	boolean readyToStop( int numOfCurrentGeneration );
}
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/StoppingConditionByNumberOfGenerations.java
```diff
@@ -0,0 +1,26 @@
...
+	public boolean readyToStop() {
+
+		// As the stopping condition is for the iterative improvement algorithm, we know that it it is checked once per loop.
+		// We can thus increment the current number of generations every time this is called.
+		generations += 1;
+		if(generations > generationThreshold) {
+			return true;
+		} else {
+			return false;
+		}
```
hartig, line 24: This part can be simplified ;-)
```suggestion
		return (generations > generationThreshold);
```

hartig (changes requested):
Here are my comments on the recent changes.

hartig:
> As part of reading the TerminatedByNumberOfGenerations I've discovered the PhysicalPlanWithCost datatype. I wonder if it would be cheaper to use this than to store the costs in a separate list.

Yes, that class has been added fairly recently.

Not sure whether it would be cheaper to use but certainly more convenient. So, I would suggest you use it. The only issue is that we are currently having another PR (#118) in which this class will be moved into a different package. Therefore, using it in your PR now might give us a merge conflict once we have merged the other PR. Consequently, I would say, ignore the class for now and simply maintain the costs in a separate list.

> I've been thinking a bit about how to best associate a physical plan with a future cost too.

Why would you want to do that? In fact, I don't think you have to deal with `CompletableFuture` objects at all in your code. As I had mentioned in [one of my earlier comments](https://github.com/LiUSemWeb/HeFQUIN/pull/115#discussion_r740470582), you can simply use the `getEstimates` function of the helper class `CostEstimationUtils` which you find in the package `se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils`. All you have to do is to collect all the physical plans for which you want a cost value; then, you pass them all in one batch to that function, and the function takes care of handling all the asynchronous part of the whole process of getting the cost values.

> I created a stopping condition interface and a class which implements it. [...]

Thanks! My recent code review gives you comments on this part.

> Since we will be using the Randomized implementation in order to create random starting points, I've decided that it would just be wasted work to implement a stop-gap solution in the meantime as it would ultimately just be replaced by the call to Randomized.

I agree!

> Speaking of it, it doesn't exist in this branch. Should we merge this branch with the Randomized branch and consider work on the Randomized plan generator to be part of creating this algorithm?

No. I suggest you finish the work in the other PR (#108) first. After that, you finish the work in this PR here but with the following limitation: instead of actually picking a random plan at each begin of the outer loop, you always pick the initial plan. Then, when everything else is ready in this PR and in the other one, we can merge both of them into the main branch, and you can create a new branch/PR in which you implement this missing bit of the functionality. Okay?

> Which class/method is the most efficient in Java for selecting a random object out of a container?

Remember the `Random` class you have used in the other PR? You can use that one as follows: Assuming the container is a list, simply call the `nextInt(int bound)` method where you use the size of the list for `bound`. The random integer that you get can then be used as an index into the list; i.e., you simply call the `get(int index)` function of the `List` interface.

> Also, does Java have pointers/references?

In a way, everything is just references in Java. Consider the following example:
```java
List<QueryOptimizationException> l = new ArrayList<>();
QueryOptimizationException e = new QueryOptimizationException("test");
l.add(e);
QueryOptimizationException e2 = l.get(0);
```
In this case, `e` and `e2` are references to the same `QueryOptimizationException` object. The object is *not* copied or anything when adding it to the list. In fact, even the list contains only a reference to that object.

> I'm thinking about creating a container and filling it with all neighbours which constitute a downward move, and then picking a random one out of those. This container will not have to remember what they cost since all lower costs are equally acceptable. 

Sounds good.

> In C++, this is where I would have it be a container of pointers/references in order to reduce data duplication.

In Java you don't have to worry about that.

SHSenpai:
@hartig The randomized object takes in a list of subplans, whereas the optimize function in the algorithm takes in a logical plan. How are these subplans generated, given a logical or physical plan?

hartig:
I assume that, by "the randomized object", you are referring to the `RandomizedJoinPlanOptimizerImpl` that you had in PR #108, and I also assume that the background of your question is related to the idea of using that `RandomizedJoinPlanOptimizerImpl` to produce the next random plan at the begin of each iteration step of the outer loop of II. The answer then is that you cannot use `RandomizedJoinPlanOptimizerImpl` directly. Instead, you may use a `SimpleJoinOrderingQueryOptimizer` that you initialize by passing a `RandomizedJoinPlanOptimizerImpl` to its constructor and, then, each time you need a new random (physical) plan for the given initial physical plan, you call its `optimizePlan(PhysicalPlan plan)` function with the initial physical plan. However, as we have talked, this is something to worry about once you have completed this PR here without actually creating random plans at the begin of each iteration step.

SHSenpai:
I'm thinking about how to make it so that the algorithm can handle any stopping condition and not just pre-defined ones. For this, one of the solutions could be creating an object which contains status about the algorithm such as current generation, time since start, and any other information which may be needed, and then pass this universal object along to any given StoppingConditionForIterativeImprovement.readyToStop function. However, the drawback of this would be that it would keep track of and send along lots of data which might not necessarily be used...

hartig:
Don't worry about it. The way how stopping conditions are captured at the moment is sufficient for now.

SHSenpai:
(Forgot to request a review the other day.)
This may require some additional polish in a place and a half, but I think that this might be good enough for a merge so that I can get to start using the Randomized class and perhaps also the PhysicalPlanWithCost.

hartig:
If you already know that this needs some more polishing, can you please do that first and request my review when you are done. That way, I may not have to take another look later.

SHSenpai:
@hartig I can't re-request a review while one is pending, so I'm pinging you instead.
I believe that everything should work well now, and given a few generations produce a pretty decent plan. Obviously we may not find the total minimum due to not having the random starting point, but that's beyond the scope of this PR.

I have two questions.
First: Would it be okay to start working on the second algorithm even without this one being done, when I have time while waiting for a review?
Second: Is there any testing I can do on my own to make sure that my code works in practice?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+		// Declaring other variables here, possibly giving the future some time to be realized with minimal waiting.
+		RewritingRule rwRule = null;
+		List<PhysicalPlan> neighbours;
+		
+		Double[] neighbourCosts;
+		List<Pair<PhysicalPlan,Double>> betterPlans = new ArrayList<>();
+		Pair<PhysicalPlan,Double> betterPlan = null;
+		Random rng = new Random();
+		
+		boolean improvementFound;
```
hartig, line 58: Also this one, it should be moved into the outer while loop, and there it should be assigned the value `false`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+		CompletableFuture<Double> futureCost = costModel.initiateCostEstimation(bestPlan);
+
+		// Declaring other variables here, possibly giving the future some time to be realized with minimal waiting.
+		RewritingRule rwRule = null;
+		List<PhysicalPlan> neighbours;
+		
+		Double[] neighbourCosts;
+		List<Pair<PhysicalPlan,Double>> betterPlans = new ArrayList<>();
+		Pair<PhysicalPlan,Double> betterPlan = null;
+		Random rng = new Random();
```
hartig, line 56: Rather than creating this variable as a local variable within this function, you can make it a member variable. Additionally, you can mark it as `final`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+		}
+		// bestCost = best possible cost out of everything that the algorithm has found.
+		Double initialCost = currentCost;
+		Double bestCost = currentCost;
+		
+		while(!condition.readyToStop(generation)) { // Currently only handles generation number as a stopping condition!
+			
+			// The randomized plan generator is to be used here. As a temporary measure, the initial plan is used.
+			currentPlan = initialPlan;
+			currentCost = initialCost;
```
hartig, line 79: Given that you (re)initialize these two variables here at the begin of every iteration step of the outer loop, you should actually also declare these variables only here and not earlier. That way, it becomes more clear that you are using these variables only within the scope of such an iteration step.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+			
+			while(true) {
```
hartig, line 81: No! You should have an actual condition here, based on the `improvementFound` variable.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+				for (int x = 0; x < neighbourCosts.length; x++) {
+					if(neighbourCosts[x] < currentCost) { // Since it's a list, iterating through the list once in the loop is probably better than using get, but I'm having this as a stop-gap measure until I've properly figured out Java lists.
+						improvementFound = true;
+						betterPlan = new Pair<PhysicalPlan, Double>(neighbours.get(x), neighbourCosts[x]);
+						betterPlans.add(betterPlan); // I haven't figured out how to do this properly for Java yet, but it will be solved by the PhysicalPlanWithCost anyway.
+					}
+				}
+				
+				if(!improvementFound) {
+					break; // break the loop if no improvement is found. This means that we have found a local minimum.
```
hartig, line 97: As mentioned above, instead of breaking out of the loop in this way, capture this condition in the while statement.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
hartig: My questions here are still open!

See also my earlier comment: https://github.com/LiUSemWeb/HeFQUIN/pull/115#issuecomment-951933055

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,98 @@
...
+	
+	/**
+	 * Currently created as a member method, but I might end up moving this to somewhere else if I find that it is needed in more places.
+	 * That obviously depends on whether Java supports importing methods outside of objects.
+	 * In C++, I would be passing references here - is something similar possible/appropriate in Java?
+	 * If significant performance is lost by making this a separate function, I'll be baking it into optimize. I've just been taught so far at LiU that making separate smaller functions is preferable so that the code becomes a bit more readable.
+	 * @return
+	 */
+	protected List<PhysicalPlan> getNeighbours(final PhysicalPlan initialPlan, final RewritingRule rewritingRule) {
+		List<PhysicalPlan> resultList = new ArrayList<PhysicalPlan>(); // Create an empty list. Just List<PhysicalPlan> didn't work so I had to look this up.
```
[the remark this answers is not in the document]
hartig:
```suggestion
		final List<PhysicalPlan> resultList = new ArrayList<PhysicalPlan>();
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+			
+			// The randomized plan generator is to be used here. As a temporary measure, the initial plan is used.
+			currentPlan = initialPlan;
+			currentCost = initialCost;
+			
+			while(true) {
+				// Given the one starting state, find a local minimum.
+				neighbours = getNeighbours(currentPlan,rwRule);
+				neighbourCosts = CostEstimationUtils.getEstimates(costModel, neighbours);
+				improvementFound = false;
```
hartig, line 85: I think you can remove this line (after having moved the declaration of `improvementFound` into the outer loop--see my other comment about this).

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+				
+				// Using a for-loop in order to have the index for which neighbouring plan to pick.
+				for (int x = 0; x < neighbourCosts.length; x++) {
+					if(neighbourCosts[x] < currentCost) { // Since it's a list, iterating through the list once in the loop is probably better than using get, but I'm having this as a stop-gap measure until I've properly figured out Java lists.
```
hartig, line 89: Regarding your comment: `neighbourCosts` is not a Java `List` but a plain array, and you are actually iterating over it here.
```suggestion
					if(neighbourCosts[x] < currentCost) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+						improvementFound = true;
+						betterPlan = new Pair<PhysicalPlan, Double>(neighbours.get(x), neighbourCosts[x]);
```
hartig, line 91: Since you are using the `betterPlan` variable only in the next line, and nowhere outside of this if-branch, you should not declare this variable somewhere else and then reuse it, but instead declare it directly here.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+					currentCost = betterPlan.object2;
+				}
+			}
+			
+			if(currentCost < bestCost) {
+				bestCost = currentCost;
+				bestPlan = currentPlan;
+			}
+			
+			generation += 1;
```
hartig, line 110:
```suggestion
			generation++;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/StoppingConditionByNumberOfGenerations.java
```diff
@@ -0,0 +1,17 @@
...
+    protected final int generationThreshold;
+
+    public StoppingConditionByNumberOfGenerations( final int generationThreshold ) {
+        this.generationThreshold = generationThreshold;
+    }
+
+	public boolean readyToStop(final int currentGeneration) {
+
+		// As the stopping condition is for the iterative improvement algorithm, we know that it it is checked once per loop.
+		// We can thus increment the current number of generations every time this is called.
```
hartig, line 14:
```suggestion
```

hartig:
In addition to my code review, here are the answers to your two other questions:
> First: Would it be okay to start working on the second algorithm even without this one being done, when I have time while waiting for a review?

Yes. Just create a separate new branch, but make sure you are creating it from the current version of the main branch and not from the branch of this PR here (i.e., first do `git checkout main` then `git pull` then `git checkout -b SimulatedAnnealing`).

> Second: Is there any testing I can do on my own to make sure that my code works in practice?

The best option would be to create a few unit tests. However, creating isolated unit test for your `IterativeImprovementBasedQueryOptimizer` is not so easy. I will have to think about how this can be done.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	public PhysicalPlan optimize( final PhysicalPlan initialPlan ) throws QueryOptimizationException {
+		// The best plan we have found so far. As we have only found one plan, it is the best one so far.
+		PhysicalPlan bestPlan = initialPlan;
+		PhysicalPlan currentPlan = bestPlan; // This variable will hold the plan which is currently being worked on.
+		
+		// Creating the completable future.
+		CostModel costModel = context.getCostModel();
+		CompletableFuture<Double> futureCost = costModel.initiateCostEstimation(bestPlan);
```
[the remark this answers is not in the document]
SHSenpai: getEstimates (at least the one I have in this branch) only takes a list and spits out an array. In this case I just want a single value for a single plan. Wouldn't it be a waste to make a function call as well as have the additional overhead of the two containers just to get. Addtionally, as far as I understand it we will lose parallelism by seeking to call the get() function immediately after making the completablefuture rather than processing other things in the meanwhile if we can.

Are there any performance upsides to using the getEstimates function here?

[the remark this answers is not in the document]
hartig: > getEstimates (at least the one I have in this branch) only takes a list and spits out an array. In this case I just want a single value for a single plan. Wouldn't it be a waste to make a function call as well as have the additional overhead of the two containers just to get.

Don't worry so much about such low-level performance optimization opportunities. The Java runtime is usually doing a reasonably good job with many of these things.

By the way, you do not have to create the list for getEstimates yourself. Instead, there is a version of this function that is declared as follows:
```
public static Double[] getEstimates( final CostModel costModel,
                                     final PhysicalPlan... plans )
```
The type `PhysicalPlan...` means that you can call the function with as many PhysicalPlan objects as you like, which will then be treated as an array. So, you can call
```
getEstimates(myCostModel, plan1, plan2, plan3);
```
but you can also simply call

```
getEstimates(myCostModel, plan1);
```

> Addtionally, as far as I understand it we will lose parallelism by seeking to call the get() function immediately after making the completablefuture rather than processing other things in the meanwhile if we can.

That's a good point in general, but these few things that you had in between wouldn't make much of a difference I would assume.

An alternative idea the may lead to an actual performance gain would be to unroll the outer loop by pulling out the first iteration step of that loop and do that step first, separate from the rest of the loop. Then, in that first step, you can request the cost of the initial plan together with requesting the cost of the neighbors of that plan. However, since that would make the code a bit more complicated, I would suggest you ignore this idea for the moment.

> Are there any performance upsides to using the getEstimates function here?

No. However, it will make the code significantly easier to read.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
SHSenpai: I've moved the declaration into the inner loop. It has to be set false at the start of each inner loop iteration, otherwise the inner loop will repeat infinitely provided an improvement can be found in the very first iteration of it.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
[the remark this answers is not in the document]
SHSenpai: Using
Double initialCost = CostEstimationUtils.getEstimates(costModel, initialPlan);
instead of my current implementation gives me an error in the IDE telling me that I need to make initialCost into an array or make Double the return type of getEstimates.

I'm making initialCost as an array with a single member, this should hopefully only lose negligible performance since an array only has a few bytes worth of overhead in C and probably something similar in Java.

SHSenpai:
Thanks for all the pointers when it comes to declarations of variables. I'm used to C programming where each declaration of a variable requires memory allocation and thus minimizing the number of declarations will have a great impact on performance.

Sorry about the miss with the RuleInstances, I seem to have forgotten about it for my previous commit.

My suggestion on testing would be something like having something which just generates (pseudo)-random plans and feeds it into the algorithm, and then checks that it's logically equivalent in some way.
After that, it would be to create some sort of benchmark similar to what the research paper uses to see how worth it it is to run the algorithm and how it compares against others.

I've prepared a sizable commit, and ended up getting some time left-over which I will use to start off with at least making up some concepts with the SA algorithm.
The only thing that feels iffy for me in the commit I'll be pushing is with the RuleInstances.
```
		RuleInstances rInst = new RuleInstances();
		RewritingRule[] rules =  (RewritingRule[]) rInst.ruleInstances.toArray();
```
The IDE wants me to make a cast here, is this right or am I doing something wrong? It seems strange to cast, because in my mind the return of the function should be a RewritingRule[] since RuleInstances.ruleInstances is a set<RewritingRule>.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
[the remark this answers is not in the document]
hartig: Yes, the return type of `getEstimates` is still an array, even if you pass only one `PhysicalPlan` to this function.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,115 @@
...
+			generation++;
+		}
+		
+		return bestPlan;
+	}
+	
+	protected List<PhysicalPlan> getNeighbours(final PhysicalPlan initialPlan) {
+		List<PhysicalPlan> resultList = new ArrayList<PhysicalPlan>(); // Create an empty list. Just List<PhysicalPlan> didn't work so I had to look this up.
+		
+		RuleInstances rInst = new RuleInstances();
```
hartig, line 100: Make `rInst` a member variable (like `rng`).

hartig: In fact, make the class `RuleApplicationsOfPlans` the member variable and then use that one in the following.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,115 @@
...
+		RewritingRule[] rules =  (RewritingRule[]) rInst.ruleInstances.toArray(); // The IDE wants me to have a cast here.
+
+		Set<RuleApplication> ruleApplications = Collections.emptySet();
+		
+		for ( final RewritingRule rr : rules) {
+			ruleApplications.addAll(rr.determineAllPossibleApplications(initialPlan));
+		}
```
hartig, line 107: Again, take a look at my earlier comment: https://github.com/LiUSemWeb/HeFQUIN/pull/115#issuecomment-951933055

You can replace all of this code here by simply calling `getRuleApplications` of `RuleApplicationsOfPlans`

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,115 @@
...
+	public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException {
+		return optimize( context.getLogicalToPhysicalPlanConverter().convert(initialPlan,false) );
+	}
+
+	public PhysicalPlan optimize( final PhysicalPlan initialPlan ) throws QueryOptimizationException {
+		// The best plan we have found so far. As we have only found one plan, it is the best one so far.
+		PhysicalPlan bestPlan = initialPlan;
+		
+		// Creating the completable future.
+		CostModel costModel = context.getCostModel(); //Created as a value and stored rather than just using the getCostModel, because the same model is used later too.
```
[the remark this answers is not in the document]
SHSenpai: Gotcha. I'll assume the cost in performance for this would be negligible. I know it always gives me the same CostModel, but I've been taught that each extra call like this is more expensive computing-wise than simply storing a variable once and then re-using it.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -0,0 +1,115 @@
...
+
+		Double[] initialCost = CostEstimationUtils.getEstimates(costModel, initialPlan);
+		
+		// generation = number of times the outer loop has run. Has to be declared here since it will increment each outer loop.
+		int generation = 0;
+		
+		// bestCost = best possible cost out of everything that the algorithm has found. Has to be declared here so that we have something to compare to in all the iterations of the outer loop without losing track of it between loops.
+		Double bestCost = initialCost[0];
```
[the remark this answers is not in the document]
SHSenpai: As I posted before, the IDE gives me an error if I do it like this. It tells me to either make it a Double[] or to change the getEstimates function to return a Double instead of a Double[].

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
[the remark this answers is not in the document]
hartig: In principle you are right. However, for an engine like ours, the difference would not be measurable I guess. In fact, I would even assume that the Java compiler can even figure out that the access to the underlying CostModel object can be inlined.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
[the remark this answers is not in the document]
hartig: Did you notice that `[0]` that I have added at the end of the line? With this, it should directly access the first element of the array returned by `getEstimates` and assign that element to the `Double`. This doesn't work in your IDE?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
hartig: To be a bit more precise here, I meant that the type of the member variable should be `RuleApplicationsOfPlans`. Something like the following:
```
   protected final RuleApplicationsOfPlans rules = new RuleApplicationsOfPlans( new RuleInstances() );
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
hartig: In other words, after creating the member variable mentioned in [my previous comment](https://github.com/LiUSemWeb/HeFQUIN/pull/115/files#r759408228), you should be able to simply do the following:
```
    final Set<RuleApplication> ruleApplications = rules.getRuleApplications(initialPlan);
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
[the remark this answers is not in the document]
SHSenpai: Oh, I didn't notice the [0], sorry! I didn't even know you could do something like that. That's neat, thanks!

SHSenpai:
This *should* be everything, including the RuleApplications as a member variable. I'm still surprised every time at the things Java does whenever you make comments. I don't know if you'll find issues here, but if you do I think it would be best to still merge and then post them in a new PR so that I can work on integrating the randomizer into the algorithm. I'm trying to read into PhysicalPlanFactory in order to get an idea of how the two could work together and whether the 'neighbours' in the IterativeImprovement would be a valid thing to insert as the 'subplans' for the randomizer.

hartig (approved):

hartig:
I have added an abstract base class to this PR, which should contain everything that you will also need for the SA algorithm.

Additionally, I have added a test suite with a unit test for your `IterativeImprovementBasedQueryOptimizer`. This unit test completes successfully, which shows that your implementation does what it is supposed to do (at least for the case that is tested). Feel free to extend this test suite in another PR with further unit tests that capture other relevant cases.

I will merge this PR now.

After that you can create a new branch for a PR that adds the remaining two things to your `IterativeImprovementBasedQueryOptimizer`; namely:
1. the use of `PhysicalPlanWithCost` (note that there is a helper class called `PhysicalPlanWithCostUtils` that provides useful functionality)
2. the random selection of parent plans in the outer loop.

Perhaps you even do this in two separate PRs, one after another (first 1, afterwards 2).

PR #116, opened 2021-10-26
added LPG2RDFConfiguration interface + default impl

scferrada:


hartig:
Now with PR #117 merge, you may come back to this one. For instance, the argument for the `mapNode` method may be of type `LPGNode`. Perhaps something else can be changed as well.

Also, remember [my comments above](https://github.com/LiUSemWeb/HeFQUIN/pull/116#issuecomment-951908100).

hartig:
Please also address points 1 and 3 in [my comment above](https://github.com/LiUSemWeb/HeFQUIN/pull/116#issuecomment-951908100).

scferrada:
Yep, As for comment 3, the mapping methods now return Jena Node objects. The only overhead I see when using the methods is then to have if statements checking if the nodes are IRI or Literal, but on the other hand, the Configurations have the freedom to choose whatever they need.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
```diff
@@ -1,4 +1,76 @@
...
+
+    /**
+     * Returns the original property name of the given mapped IRI
+     * @param iri the mapped IRI of a property
+     */
+    String unmapProperty(final String iri);
+
+    /**
+     * Returns the IRI Node of the property defined as u_label
+     * @return a String with the IRI
```
hartig, line 72:
```suggestion
     * @return a String with the IRI
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
```diff
@@ -1,4 +1,75 @@
...
+    /**
+     * Returns the original edge label for the given mapped label
+     * @param label the mapped IRI or literal of an edge label
+     */
+    String unmapEdgeLabel(final String label);
+
+    /**
+     * Returns an IRI Node for a given property
+     * @param property the name of the property to map
+     * @return a String with the IRI
```
hartig, line 60:
```suggestion
```

hartig:
```suggestion
```

hartig:
Looks good. I will merge now.

The next task will then be to create another PR with code for parsing a Cypher result string into a list of `TableRecord` objects.

Issue #71, opened 2021-08-10
Dynamic Programming algorithm for query planning


hartig:
The current implementation of the optimizer in `se.liu.ida.hefquin.engine.queryproc.impl.optimizer.dynamicProgramming` is not actually implementing the classical DP algorithm. It should be replaced by a correct implementation.

Besides the original description in the paper by Selinger et al., the are several descriptions in database textbooks. For instance, [Ramakrishnan and Gehrke (3rd ed)](https://github.com/pforpallav/school/blob/master/CPSC404/Ramakrishnan%20-%20Database%20Management%20Systems%203rd%20Edition.pdf) have such a description in Sec.15.4.2 (pages 496ff), and so do [Garcia-Molina et al. (2nd ed)](https://people.inf.elte.hu/kiss/DB/ullman_the_complete_book.pdf) in Sec.16.6.4 (pages 819ff).

In fact, there are two version of the algorithm, a simplified one that focuses only on left-deep plans and a more general one that can also considers bushy plans. Ramakrishnan and Gehrke cover only the first one, whereas Garcia-Molina et al. cover both.

Speaking of versions, there is also an extension of the DP algorithm that is called _iterative dynamic programming_. I will check out what this is. According to [Kossmann's survey on The State of the Art in Distributed Query Processing](https://doi.org/10.1145/371578.371598), _"this extended algorithm is adaptive and produces as good plans as basic dynamic programming for simple queries and “as good as possible plans” for complex queries for which dynamic programming is not viable."_

/cc @chengsijin0817

hartig:
[...]

PR #118, opened 2021-11-03
Dynamic programming algorithm for query planning

chengsijin0817:
Implement a dynamic programming algorithm for query planning to address issue #71 .


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/DynamicProgramming/DPBasedOptimizerHelper.java
```diff
@@ -0,0 +1,71 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.DynamicProgramming;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class DPBasedOptimizerHelper {
```
hartig, line 7: I don't think we need an extra class for that. Simply move these two helper methods into `DPBasedJoinPlanOptimizer`.

hartig (changes requested):
Here are some initial comments.

Generally, I wonder whether there is a better (more efficient) way to implement this. I mean, using this `optPlan` map seems highly inefficient to me, with all the possible subsets of plans as keys in this map. The reason why you need something like this is because you have implemented the algorithm in a bottom-up fashion. Can you try a second alternative implementation that is top-down; I believe that such an implementation would not need such a big map.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DataStructureForStoringPlansOfSubsets.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class DataStructureForStoringPlansOfSubsets {
+    protected Map< Integer, Map<List<PhysicalPlan>, PhysicalPlan> > map=new HashMap<>();
+
+    public void add( final List<PhysicalPlan> subsets, final PhysicalPlan plan ) {
+        final int size=subsets.size();
+        map.putIfAbsent( size, new HashMap<>() );
+        map.get(size).put(subsets, plan);
```
hartig, line 15: It is probably more efficient to do the `map.get(size)` right away and if the result is `null` you create the new hash map and put it into the main map.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DataStructureForStoringPlansOfSubsets.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public PhysicalPlan get( final List<PhysicalPlan> subsets ) {
+        return map.get(subsets.size()).getOrDefault( subsets,null );
```
hartig, line 19: `map.get(subsets.size())` may return `null`, in which case you get a `NullPointerException` when you try to call `getOrDefault`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DPBasedJoinPlanOptimizer.java
```diff
@@ -0,0 +1,164 @@
...
+            for ( int num = 2; num < subplans.size(); num ++ ){
+                // Get all subsets with size num.
+                final List<List<PhysicalPlan>> subsets = getSubSet(subplans, num);
+
+                for( final List<PhysicalPlan> plans : subsets ){
+                    final List<PhysicalPlan> candidatePlans = new ArrayList<>();
+
+                    // Split the current set of subplans into two subsets, and create candidate plans with join for each of the combinations.
+                    final List<Pair<List<PhysicalPlan>, List<PhysicalPlan>>> candidatePairs = splitIntoSubSets(plans);
+                    for ( Pair p: candidatePairs ){
```
hartig, line 63:
```suggestion
                    for ( final Pair p: candidatePairs ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DPBasedJoinPlanOptimizer.java
```diff
@@ -0,0 +1,164 @@
...
+                        final PhysicalPlan plan_left = optPlan.get((List<PhysicalPlan>) p.object1);
+                        final PhysicalPlan plan_right = optPlan.get((List<PhysicalPlan>) p.object2);
```
hartig, line 65: I believe that the two casts to `List<PhysicalPlan>` should not be necessary because the generic class `Pair` has the relevant type parameters. However, to make this work, you may have to replace `Pair` in the for-loop by `Pair<List<PhysicalPlan>, List<PhysicalPlan>>`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DPBasedJoinPlanOptimizer.java
```diff
@@ -0,0 +1,164 @@
...
+
+                        candidatePlans.add( PhysicalPlanFactory.createPlanWithJoin( plan_left,  plan_right) );
```
hartig, line 67: Before doing that, include a check to verify that `plan_left` is not `null` and `plan_right` is not `null`. In principle, they should not, but better verify to avoid weird bugs. If they are `null`, for whatever reason, throw an `IllegalStateException`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DPBasedJoinPlanOptimizer.java
```diff
@@ -0,0 +1,164 @@
...
+
+                        if ( IdentifyTypeOfRequestUsedForReq.isBGPRequest( plan_left.getRootOperator() ) ){
+                            final LogicalOpBGPAdd newRoot = LogicalOpUtils.createBGPAddLopFromReq((BGPRequest) plan_left.getRootOperator());
+
+                            candidatePlans.add( PhysicalPlanFactory.createPlanWithIndexNLJ( newRoot, plan_right ) );
+                            candidatePlans.add( PhysicalPlanFactory.createPlanWithBindJoinFILTER( newRoot, plan_right ) );
+                            candidatePlans.add( PhysicalPlanFactory.createPlanWithBindJoinUNION( newRoot, plan_right ) );
+                            candidatePlans.add( PhysicalPlanFactory.createPlanWithBindJoinVALUES( newRoot, plan_right ) );
```
hartig, line 75: This assumes that the fed.member of the BGP request is a SPARQL endpoint. While, currently, this is the only possibility, in the future we may also support other types of interfaces that support BGP requests. Therefore, it would be good to include a second condition in the if-statement to check that it is indeed a SPARQL endpoint.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DPBasedJoinPlanOptimizer.java
```diff
@@ -0,0 +1,164 @@
...
+
+    }
+
+    // This method returns all subsets (with the given size) of the given superset.
+    public static <T> List<List<T>> getSubSet( final List<T> superset, final int n ){
+        final List<List<T>> result = new ArrayList<>();
+        if( n==0 ){
+            result.add( new ArrayList<>() );
+            return result;
+        }
```
hartig, line 108: Actually, I would consider `n==0` to be an invalid argument (in fact, any `n<1`). You should throw an `IllegalArgumentException`.

Similarly, throw an `IllegalArgumentException` if `n>superset.size()`.

Additionally, if `n==superset.size()` you can immediately return `superset` without having to do anything else.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/simple/DPBasedJoinPlanOptimizer.java
```diff
@@ -0,0 +1,164 @@
...
+                    result.add(clone);
+                }
+            }
+        }
+
+        return result;
+    }
+
+    // Split a superset into two subsets. This method returns all possible combinations of subsets.
+    public static <T> List<Pair<List<T>, List<T>>> splitIntoSubSets( final List<T> superset ){
```
hartig, line 137: Just for the sake of safety, throw an `IllegalArgumentException` if `superset.size() < 2`.

src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/DynamicProgramming/DPBasedOptimizerHelperTest.java
```diff
@@ -0,0 +1,58 @@
...
+public class DPBasedOptimizerHelperTest extends EngineTestBase {
+    final List<Integer> list= List.of( new Integer[]{1, 2, 3, 4, 5} );
+
+    @Test
+    public void getSubsets_1()
+    {
+        final List<List<Integer>> subsets = DPBasedJoinPlanOptimizer.getSubSet( list, 3 );
+
+        assertEquals( 10, subsets.size() );
+        subsets.forEach( System.out::println );
```
hartig, line 23: Unit tests should better not have such side effects (printing to std.out). Instead, for the given case, it would be better if the unit test actually checks (automatically) that `subsets` contains all the expected subsets.

hartig (changes requested):
Here are my comments. Sorry for the delay!

hartig (approved):
Looks all good now! :-)

...I will merge this PR.

PR #120, opened 2021-11-18
Triple pattern translation

scferrada:
I still need to do a last thorough check to compare with the rules on the paper, but this should be very close.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
@@ -25,6 +25,13 @@ public interface LPG2RDFConfiguration {
     Node mapNode(final LPGNode node);
 
     /**
+     * Checks if the given node is the mapping of an
+     * LPG Node using this configuration
+     * @param n the Node to be tested
+     */
+    boolean mapsToNode(final Node n);
+
+    /**
      * Returns the corresponding LPGNode of a given IRI
      * @param iri the IRI to be unmapped
      */
@@ -37,16 +44,30 @@ public interface LPG2RDFConfiguration {
     Node mapNodeLabel(final String label);
 
     /**
+     * Checks if the given node is the mapping of a
+     * Node Label using this configuration
+     * @param n the Node to be tested
+     */
+    boolean mapsToLabel(final Node n);
+
+    /**
      * Returns the original node label for the given mapped label
      * @param label the mapped IRI or literal of a node label
      */
-    String unmapNodeLabel(final String label);
+    String unmapNodeLabel(final Node label);
 
     /**
      * Returns an IRI or Literal Node for a given edge label
      * @param label the edge label to map
      */
     Node mapEdgeLabel(final String label);
+
+    /**
+     * Checks if the given node is the mapping of an
+     * Edge Label using this configuration
+     * @param n the Node to be tested
+     */
+    boolean mapsToEdgeLabel(final Node n);
 
     /**
      * Returns the original edge label for the given mapped label
@@ -61,6 +82,13 @@ public interface LPG2RDFConfiguration {
     Node mapProperty(final String property);
 
     /**
+     * Checks if the given node is the mapping of a property using
+     * this configuration or not.
+     * @param n the Node to be tested
+     */
+    boolean mapsToProperty(final Node n);
+
+    /**
      * Returns the original property name of the given mapped IRI
      * @param iri the mapped IRI of a property
      */
@@ -71,13 +99,11 @@ public interface LPG2RDFConfiguration {
      */
     Node getLabel();
 
-    boolean mapsToProperty(final Node n);
-
+    /**
+     * Checks if the given node is equal to the
+     * Label IRI of this configuration or not.
+     * @param n the Node to be tested
+     */
     boolean isLabelIRI(final Node n);
 
-    boolean mapsToLabel(final Node n);
-
-    boolean mapsToRelationship(final Node n);
-
-    boolean mapsToNode(final Node n);
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
@@ -20,7 +20,7 @@ public interface LPG2RDFConfiguration {
 
     /**
      * Returns a Jena Node for the given LPGNode object.
-     * @param node the id of the node to be mapped
+     * @param node the LPGNode node to be mapped
      */
     Node mapNode(final LPGNode node);
 
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
@@ -71,4 +71,13 @@ public interface LPG2RDFConfiguration {
      */
     Node getLabel();
 
+    boolean mapsToProperty(final Node n);
+
+    boolean isLabelIRI(final Node n);
+
+    boolean mapsToLabel(final Node n);
+
+    boolean mapsToRelationship(final Node n);
+
+    boolean mapsToNode(final Node n);
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
@@ -19,20 +19,18 @@ import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
 public interface LPG2RDFConfiguration {
 
     /**
-     * Returns a Jena Node for the given LPGNode object.
-     * @param node the LPGNode node to be mapped
+     * Returns a URI or a blank node (in the form of a Jena {@link Node} object) for the given LPG node.
      */
     Node mapNode(final LPGNode node);
 
     /**
-     * Checks if the given node is the mapping of an
-     * LPG Node using this configuration
-     * @param n the Node to be tested
+     * Returns true if the given RDF term is in the image of the mapNode function and, thus, in the domain of the unmapNode function.
      */
     boolean mapsToNode(final Node n);
 
     /**
-     * Returns the corresponding LPGNode of a given Jena Node
+     * Returns the LPG node that corresponds to the given RDF term.
+     * This function is the inverse of mapNode. As such, it is defined only for the RDF terms that may be returned by the mapNode function and, thus, throws an {@link IllegalArgumentException} if any other RDF term is given to it.
      */
     LPGNode unmapNode(final Node node);
 
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
@@ -19,7 +19,7 @@ import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
 public interface LPG2RDFConfiguration {
 
     /**
-     * Returns an IRI Node for the given node identifier.
+     * Returns a Jena Node for the given LPGNode object.
      * @param node the id of the node to be mapped
      */
     Node mapNode(final LPGNode node);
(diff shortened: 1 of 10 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/SPARQLStar2CypherTranslator.java
```diff
@@ -0,0 +1,470 @@
...
+        }
+
+        public List<CypherVar> getEdgeVars(final Node n) {
+            if (edgeVars.containsKey(n.getName()))
+                return edgeVars.get(n.getName());
+            edgeCount++;
+            final CypherVar source = new CypherVar("src" + edgeCount);
+            final CypherVar edge = new CypherVar("edge" + edgeCount);
+            final CypherVar target = new CypherVar("tgt" + edgeCount);
+            final List<CypherVar> vars = new ArrayList<>();
```
hartig, line 449:
```suggestion
            final List<CypherVar> vars = new ArrayList<>(3);
```

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/SPARQLStar2CypherTranslatorTest.java
```diff
@@ -0,0 +1,333 @@
...
+                new CypherUnionQueryImpl(
+                    CypherQueryBuilder.newBuilder().match(new NodeMatchClause(v1))
+                            .condition(new PropertyValueCondition(v1, "name", "Quentin Tarantino"))
+                            .returns(new VariableReturnStatement(v1, ret1))
+                            .build(),
+                    CypherQueryBuilder.newBuilder()
+                            .match(new EdgeMatchClause(src1, edge1, tgt1))
+                            .condition(new PropertyValueCondition(edge1, "name", "Quentin Tarantino"))
+                            .returns(new TripleMapReturnStatement(src1, edge1, tgt1, ret1))
+                            .build()), translation);
```
hartig, line 63: Please move `translation` to a new line, at the same indentation level as the `new` on line 54. That way, it will be much easier to see what is supposed to be equal to what. (Same thing for the other unit tests below.)

hartig (changes requested):

hartig:
In addition to the comments in my review, I was also thinking whether the `translateTriplePattern` function should produce not only the Cypher query but also some data structure needed to interpret the result records after executing the Cypher query and converting these result records into solution mappings. However, I would say that you first complete this PR here, and after it is merged, you can think about extending the code to produce such a data structure.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/SPARQLStar2CypherTranslator.java
```diff
@@ -0,0 +1,13 @@
...
+import se.liu.ida.hefquin.engine.query.impl.TriplePatternImpl;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherQuery;
+
+public interface SPARQLStar2CypherTranslator {
+
+    /**
+     * Translates the given Triple Pattern into a Cypher query, using a given LPG2RDFConfiguration.
+     * If the Triple Pattern is incompatible with the configuration, this method returns null.
+     */
+    CypherQuery translateTriplePattern(final TriplePatternImpl tp, final LPG2RDFConfiguration conf);
```
hartig, line 12:
```suggestion
    CypherQuery translateTriplePattern(final TriplePattern tp, final LPG2RDFConfiguration conf);
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/DefaultConfiguration.java
```diff
@@ -31,8 +31,9 @@ public Node mapNodeLabel(final String label) {
     }
 
     @Override
-    public String unmapNodeLabel(String label) {
-        return label.replaceAll(NS + NODELABEL, "");
+    public String unmapNodeLabel(final Node label) {
+        final String text = label.isURI()? label.getURI() : label.getLiteralValue().toString();
+        return text.replaceAll(NS + NODELABEL, "");
```
hartig, line 36: I don't think that this is correct. This function should be the inverse of `mapNodeLabel`. In this sense, it shouldn't be able to handle `Node` objects that represent literals and, in fact, it shouldn't even be able to handle all URIs, but only those that can be produced by `mapNodeLabel`. Hence, it should throw an IllegalArgumentException for every other type of `Node` object (which should be exactly all the `Node` objects for which `mapsToLabel` returns `false`).

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -0,0 +1,474 @@
...
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.match.EdgeMatchClause;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.match.NodeMatchClause;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.returns.*;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.utils.CypherQueryBuilder;
+
+import java.util.*;
+
+public class SPARQLStar2CypherTranslatorImpl implements SPARQLStar2CypherTranslator {
+
+    public CypherQuery translateTriplePattern(final TriplePatternImpl tp, final LPG2RDFConfiguration conf) {
```
hartig, line 24:
```suggestion
    public CypherQuery translateTriplePattern(final TriplePattern tp, final LPG2RDFConfiguration conf) {
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/SPARQLStar2CypherTranslator.java
```diff
@@ -0,0 +1,13 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper;
+
+import se.liu.ida.hefquin.engine.query.impl.TriplePatternImpl;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherQuery;
+
+public interface SPARQLStar2CypherTranslator {
+
+    /**
+     * Translates the given Triple Pattern into a Cypher query, using a given LPG2RDFConfiguration.
+     * If the Triple Pattern is incompatible with the configuration, this method returns null.
```
hartig, line 10: I don't think it's only about being compatible with the configuration. The triple pattern may generally be of a shape for which the (configuration-specific) RDF-star view of the queried LPG is guaranteed to contain no matching triples.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -0,0 +1,474 @@
...
+    private static int getNumberOfVars(final Node s, final Node p, final Node o) {
+        int count = 0;
+        if (s.isVariable())
+            count++;
+        if (p.isVariable())
+            count++;
+        if (o.isVariable())
+            count++;
+        return count;
+    }
```
hartig, line 110: No need for this helper function here. Instead, simply use `getNumberOfVarOccurrences` that we have in `QueryPatternUtils`.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherQueryBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public CypherQueryBuilder add(final Object clause) {
+        if (clause instanceof MatchClause) {
+            this.addMatch((MatchClause) clause);
+        } else if (clause instanceof WhereCondition) {
+            this.addCondition((WhereCondition) clause);
+        } else if (clause instanceof ReturnStatement) {
+            this.addReturn((ReturnStatement) clause);
+        }
```
hartig, line 46: Add an `else`-branch here and throw an `IllegalArgumentException`!

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -0,0 +1,474 @@
...
+                                                    final LPG2RDFConfiguration configuration,
+                                                    final CypherVarGenerator gen,
+                                                    final Set<Node> certainNodes) {
+        final String property = configuration.unmapProperty(p.getURI());
+        final String literal = o.getLiteralValue().toString();
+        final CypherVar svar = gen.getVarFor(s);
+        final CypherMatchQuery q = new CypherQueryBuilder()
+                                    .addMatch(new NodeMatchClause(svar))
+                                    .addCondition(new PropertyValueCondition(svar, property, literal))
+                                    .addReturn(new VariableReturnStatement(svar, gen.getRetVar(s)))
```
hartig, line 122: Use the `add` method here; and similarly in all the other cases below.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -0,0 +1,474 @@
...
+        private final Map<String, List<CypherVar>> edgeVars = new HashMap<>();
+
+        private int varCount = 0;
+        private int edgeCount = 0;
+        private int anonCount = 0;
+        private int retVarCount = 0;
+
+        private final String varPrefix = "cpvar";
+        private final String retPrefix = "ret";
+        private final String anonPrefix = "a";
```
hartig, line 432: I think that these three can even be `static` (in addition to being `final`).

hartig (changes requested):

hartig:
In addition to the points in my review, I also notice that there are still some `unmap...` functions in the `LPG2RDFConfiguration` interface where the argument is of type `String` but should be `Node`.

Also, the JavaDoc comment in lines 22-23 is not entirely correct: i) the input is not a node id and ii) the output may be a blank node instead of an IRI.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
```diff
@@ -19,56 +19,90 @@
 public interface LPG2RDFConfiguration {
 
     /**
-     * Returns an IRI Node for the given node identifier.
-     * @param node the id of the node to be mapped
+     * Returns a Jena Node for the given LPGNode object.
+     * @param node the LPGNode node to be mapped
```
hartig, line 23:
```suggestion
     * Returns a URI or a blank node (in the form of a Jena {@link Node} object) for the given LPG node.
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
```diff
@@ -19,56 +19,90 @@
...
      */
     Node mapNode(final LPGNode node);
```
hartig, line 25:
```suggestion
    Node mapNode(LPGNode node);
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
```diff
@@ -19,56 +19,90 @@
...
 
     /**
-     * Returns the corresponding LPGNode of a given IRI
-     * @param iri the IRI to be unmapped
+     * Checks if the given node is the mapping of an
+     * LPG Node using this configuration
+     * @param n the Node to be tested
```
hartig, line 30:
```suggestion
     * Returns true if the given RDF term is in the image of the mapNode function and, thus, in the domain of the unmapNode function.
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/LPG2RDFConfiguration.java
```diff
@@ -19,56 +19,90 @@
...
      */
-    LPGNode unmapNode(final String iri);
+    boolean mapsToNode(final Node n);
 
     /**
-     * Returns an IRI or Literal Node for the given node label
-     * @param label the label to be mapped
+     * Returns the corresponding LPGNode of a given Jena Node
```
hartig, line 35:
```suggestion
     * Returns the LPG node that corresponds to the given RDF term.
     * This function is the inverse of mapNode. As such, it is defined only for the RDF terms that may be returned by the mapNode function and, thus, throws an {@link IllegalArgumentException} if any other RDF term is given to it.
```

hartig (approved):
Looks good now. I will merge it.

hartig:
As the next PR, extend the code of this one to produce not only the Cypher query but also some data structure needed to interpret the result records after executing the Cypher query in order to convert these result records into solution mappings.

PR #126, opened 2021-12-07
Apply different optimization algorithms in QueryOptimizer

chengsijin0817:
evolutionary algorithm, dynamic programming algorithm, greedy algorithm


hartig:
Instead of changing `QueryOptimizerImpl` to call other implementations of `QueryOptimizer`, it is better to change the implementation of `QueryOptimizer` created by `OpExecutorHeFQUIN`. In the following commit, I have done this in a flexible way (by introducing a `QueryOptimizerFactory` that can be accessed via from the context object).

See commit: https://github.com/LiUSemWeb/HeFQUIN/commit/1f9213cfdd9e3e8e5d596925fbc27c81fefdda99

I think that this commit makes this PR obsolete.

chengsijin0817:
> See commit: [1f9213c](https://github.com/LiUSemWeb/HeFQUIN/commit/1f9213cfdd9e3e8e5d596925fbc27c81fefdda99)
> 
> I think that this commit makes this PR obsolete.

Thanks, it makes more sense to implement QueryOptimizer by introducing QueryOptimizerFactory.
Therefore, I will close this pull request.

PR #127, opened 2021-12-07
Randomized Algorithms for Optimizing Large Join Queries

SHSenpai:
I've edited a bit in the base class, which I want to use for all three algorithms.
Because of this, I've decided to create this is as my common branch. To make things easier for you to know what is what and keep order, I will create branches for each of the algorithms _out of this branch_. I'm also going to start working on two-phase optimization (2PO) soon, and I'll likely look back and forth a lot between all of the algorithms as I want to possibly have the Iterative and Annealing algorithms have an optimize function which is static and ideal for use in the 2PO in a way that the 2PO becomes sufficiently usable with different settings (input values to the 2PO will include things such as how many generations II will run before SA takes over, etc).

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionByRelativeSubplanCount.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionByRelativeSubplanCount.java
@@ -1,0 +1,15 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
+
+public class EquilibriumConditionByRelativeSubplanCount implements EquilibriumConditionForSimulatedAnnealing {
+	
+    protected final int scalingFactor;
+
+    public EquilibriumConditionByRelativeSubplanCount( final int scalingFactor ) {
+        this.scalingFactor = scalingFactor;
+    }
+
+	public boolean isEquilibrium(final int currentGeneration, final int subplans) {
+		return (currentGeneration >= subplans * scalingFactor);
+	}
+
+}
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionByRelativeSubplanCount.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionByRelativeSubplanCount.java
@@ -1,12 +1,18 @@
 package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
 
-public class EquilibriumConditionByRelativeSubplanCount implements EquilibriumConditionForSimulatedAnnealing {
-	
-    protected final int scalingFactor;
+public class EquilibriumConditionByRelativeSubplanCount implements EquilibriumConditionForSimulatedAnnealing
+{	
+	public final static int DEFAULT_SCALING_FACTOR = 16;
 
-    public EquilibriumConditionByRelativeSubplanCount( final int scalingFactor ) {
-        this.scalingFactor = scalingFactor;
-    }
+	protected final int scalingFactor;
+
+	public EquilibriumConditionByRelativeSubplanCount( final int scalingFactor ) {
+		this.scalingFactor = scalingFactor;
+	}
+
+	public EquilibriumConditionByRelativeSubplanCount() {
+		this(DEFAULT_SCALING_FACTOR);
+	}
 
 	public boolean isEquilibrium(final int currentGeneration, final int subplans) {
 		return (currentGeneration >= subplans * scalingFactor);
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionForSimulatedAnnealing.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionForSimulatedAnnealing.java
@@ -1,0 +1,5 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
+
+public interface EquilibriumConditionForSimulatedAnnealing {
+	public boolean isEquilibrium(final int currentGeneration, final int subplanCount);
+}
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionForSimulatedAnnealing.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionForSimulatedAnnealing.java
@@ -1,5 +1,6 @@
 package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
 
-public interface EquilibriumConditionForSimulatedAnnealing {
-	public boolean isEquilibrium(final int currentGeneration, final int subplanCount);
+public interface EquilibriumConditionForSimulatedAnnealing
+{
+	boolean isEquilibrium(int currentGeneration, int subplanCount);
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
@@ -51,7 +51,7 @@ public class IterativeImprovementBasedQueryOptimizer extends RandomizedQueryOpti
 				}
 
 				if(improvementFound) { // if we have found at least one possible improvement, we want to make one of them into our new current plan.
-					currentPlan = getRandom(betterPlans); // Get a random object.
+					currentPlan = getRandomElement(betterPlans); // Get a random object.
 				}
 			}
 			while(improvementFound);
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
@@ -32,14 +32,16 @@ public class IterativeImprovementBasedQueryOptimizer extends RandomizedQueryOpti
 	
 	public PhysicalPlan optimize( final PhysicalPlan initialPlan ) throws QueryOptimizationException {
 		// The best plan and cost we have found so far. As we have only found one plan, it is the best one so far.
-		PhysicalPlanWithCost bestPlan = new PhysicalPlanWithCost( initialPlan, CostEstimationUtils.getEstimates( context.getCostModel(), initialPlan )[0]);
+		PhysicalPlanWithCost bestPlan = PhysicalPlanWithCostUtils.annotatePlanWithCost( context.getCostModel(), initialPlan );
+
 
 		// generation = number of times the outer loop has run. Has to be declared here since it will increment each outer loop.
 		int generation = 0;
 
 		while ( !condition.readyToStop(generation) ) { // Currently only handles generation number as a stopping condition!
 			// The randomized plan generator is to be used here. As a temporary measure, the initial plan is used.
-			PhysicalPlanWithCost currentPlan = new PhysicalPlanWithCost(initialPlan, CostEstimationUtils.getEstimates( context.getCostModel(), initialPlan )[0]); // This variable will hold the plan which is currently being worked on.
+			PhysicalPlanWithCost currentPlan = PhysicalPlanWithCostUtils.annotatePlanWithCost( context.getCostModel(), initialPlan ); // This variable will hold the plan which is currently being worked on.
+
 			
 			boolean improvementFound = false;
 
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
@@ -2,6 +2,8 @@ package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
 
 import java.util.ArrayList;
 import java.util.List;
+
+import se.liu.ida.hefquin.engine.queryplan.LogicalPlan;
 import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
 import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationContext;
@@ -24,6 +26,10 @@ public class IterativeImprovementBasedQueryOptimizer extends RandomizedQueryOpti
 	}
 
 	@Override
+	public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException {
+		return optimize( context.getLogicalToPhysicalPlanConverter().convert(initialPlan,false) );
+	}
+	
 	public PhysicalPlan optimize( final PhysicalPlan initialPlan ) throws QueryOptimizationException {
 		// The best plan and cost we have found so far. As we have only found one plan, it is the best one so far.
 		PhysicalPlanWithCost bestPlan = new PhysicalPlanWithCost( initialPlan, CostEstimationUtils.getEstimates( context.getCostModel(), initialPlan )[0]);
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
@@ -25,7 +25,7 @@ public class IterativeImprovementBasedQueryOptimizer extends RandomizedQueryOpti
 		condition = x;
 	}
 
-	@Override
+	// @Override  TODO: fix after merging to main!!
 	public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException {
 		return optimize( context.getLogicalToPhysicalPlanConverter().convert(initialPlan,false) );
 	}
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/RandomizedQueryOptimizerBase.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/RandomizedQueryOptimizerBase.java
@@ -10,7 +10,7 @@ import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
 import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
 import se.liu.ida.hefquin.engine.queryproc.QueryOptimizer;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationContext;
-import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.evolutionaryAlgorithm.RuleApplicationsOfPlans;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.PlanRewritingUtils;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RuleApplication;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RuleInstances;
 
@@ -18,7 +18,7 @@ public abstract class RandomizedQueryOptimizerBase implements QueryOptimizer
 {
 	protected final Random rng = new Random();
 
-	protected final RuleApplicationsOfPlans rules;
+	protected final PlanRewritingUtils rules;
 	protected final QueryOptimizationContext context;
 
 	protected RandomizedQueryOptimizerBase( final QueryOptimizationContext context,
@@ -26,7 +26,7 @@ public abstract class RandomizedQueryOptimizerBase implements QueryOptimizer
 		assert context != null;
 		this.context = context;
 
-		rules = new RuleApplicationsOfPlans(rewritingRules);
+		rules = new PlanRewritingUtils(rewritingRules);
 	}
 
 	@Override
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/RandomizedQueryOptimizerBase.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/RandomizedQueryOptimizerBase.java
@@ -29,12 +29,9 @@ public abstract class RandomizedQueryOptimizerBase implements QueryOptimizer
 		rules = new PlanRewritingUtils(rewritingRules);
 	}
 
-	@Override
-	public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException {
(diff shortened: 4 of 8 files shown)


SHSenpai:
> That makes it very hard for me to see what you have done here (at least if I look only in the Github Web interface, which is where I have to look to put review comments).

Gotcha. This branch contains all changes to SA, as well as the introduction of a generic function to the base class which I use in the base class.

The SA class is almost but not quite done. There's something in the paper which I simply do not understand, which I need to understand before I can finish the algorithm. The algorithm paper implies that 16*(number of joins in query) is a boolean and I don't understand what this entails.

> Simulated Annealing almost done, but I am at a loss for how to implement Equilibrium. It's supposed to be a boolean from what I can read in the paper, but the paper's example implementation defines it to be 16*(number or joins in query). (table 2.2, page 314 in the paper, page 3 in the pdf)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
```diff
@@ -0,0 +1,82 @@
...
+			
+				// Here we want a random neighbour, and its cost
+				// My first instinct to do something like
+				// final PhysicalPlanWithCost alternativePlan = (PhysicalPlan alternativePlanPlan = getRandom(getNeighbours(currentPlan.getPlan())), CostEstimationUtils.getEstimates( context.getCostModel(), alternativePlanPlan )[0])
+				// to make the code really compact.
+				// because it seems wasteful to create a local variable that's stored in optimize() just because we need to get the cost separately.
+				final PhysicalPlan temporaryPlan = getRandom(getNeighbours(currentPlan.getPlan()));
+				final PhysicalPlanWithCost alternativePlan = new PhysicalPlanWithCost(temporaryPlan, CostEstimationUtils.getEstimates( context.getCostModel(), temporaryPlan )[0]);
+				// Alternatively, a better option is possibly to create a new constructor for PhysicalPlanWithCost...
+				// ...that only takes a PhysicalPlan argument, and gets the cost itself.
```
hartig, line 47: No, don't create such a new constructor -- see [my earlier comment](https://github.com/LiUSemWeb/HeFQUIN/pull/134#issuecomment-991719903).

However, here's what you can do: Please add the following method to `PhysicalPlanWithCostUtils`:
```java
    public static PhysicalPlanWithCost annotatePlanWithCost( final CostModel cm, final PhysicalPlan plan ) throws QueryOptimizationException {
        final Double[] costs;
        try {
            costs = CostEstimationUtils.getEstimates(cm, plan);
        } catch ( final CostEstimationException e ) {
            throw new QueryOptimizationException( "Determining the cost for the plan caused an exception.", e.getCause() );
        }

        return new PhysicalPlanWithCost( plan, costs[0] );
    }
```
Once you have that, you can change the line here as follows:
```suggestion
				final PhysicalPlanWithCost alternativePlan = PhysicalPlanWithCostUtils.annotatePlanWithCost( context.getCostModel(), temporaryPlan );
```

hartig: In fact, you can then also use this new helper function in line 21 above.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
```diff
@@ -0,0 +1,82 @@
...
+				
+				double costDelta = currentPlan.getWeight() - alternativePlan.getWeight();
```
hartig, line 49:
```suggestion
				final double costDelta = currentPlan.getWeight() - alternativePlan.getWeight();
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class SimulatedAnnealing extends RandomizedQueryOptimizerBase {
+
+	public SimulatedAnnealing(	final QueryOptimizationContext context,
+								final RuleInstances rewritingRules) {
+		super(context, rewritingRules);
+	}
+
+	@Override
+	public PhysicalPlan optimize(PhysicalPlan initialPlan) throws QueryOptimizationException {
```
hartig, line 18:
```suggestion
	public PhysicalPlan optimize( final PhysicalPlan initialPlan ) throws QueryOptimizationException {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+		
+		return bestPlan.getPlan();
+	}
+	
+	protected boolean isEquilibrium() {
+		return false;
+	}
+	
+	protected boolean isFrozen(double temperature, int unchanged) {
```
hartig, line 77:
```suggestion
	protected boolean isFrozen( final double temperature, final int unchanged ) {
```

hartig:
> but I am at a loss for how to implement Equilibrium. It's supposed to be a boolean from what I can read in the paper, but the paper's example implementation defines it to be 16*(number or joins in query). (table 2.2, page 314 in the paper, page 3 in the pdf)

I agree that this aspect of the algorithm should have been stated more explicitly in the paper. However, my interpretation is as follows: The equilibrium is assumed to be reached after the inner loop has been run a fixed number of times, where this number is 16 * (the number of joins in the query). Hence, for queries with more joins, the inner loop runs more often.

The follow-up question then is: What do we consider as "the number of joins in the query"? The issue is that the paper considers only join plans but, in our case, there may also be other operators in the plans (i.e., not only joins). A possible solution would be to adapt the equilibrium condition to our setting as follows. Instead of fixing the number of inner-loop runs to be 16 * (the number of joins in the query), we fix it to be 16 * (the number of logical operators in the given logical plan). What do you think of this idea?

hartig (changes requested):
Besides my response to your question about the equilibrium, here is my review of the current code in this PR.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -7,7 +7,8 @@
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationContext;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RuleInstances;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.CostEstimationUtils;
-import se.liu.ida.hefquin.engine.utils.Pair;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.PhysicalPlanWithCost;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.PhysicalPlanWithCostUtils;
 
 public class IterativeImprovementBasedQueryOptimizer extends RandomizedQueryOptimizerBase
```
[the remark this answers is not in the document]
SHSenpai: The II branch was a branch out of this branch. I wanted to make a separate branch for the IterativeImprovement changes since you wanted the PR's to be more limited in scope, but I wanted to use something from a change I had made to the base class.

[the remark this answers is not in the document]
hartig: Okay.

SHSenpai:
> Instead of fixing the number of inner-loop runs to be 16 * (the number of joins in the query), we fix it to be 16 * (the number of logical operators in the given logical plan). What do you think of this idea?

This seems sound, at least for now when we do not have any other idea. Testing later on will provide the data we need to evaluate the performance of all of our settings.
I'm thinking of whether I should be implementing an equilibrium condition for the SA algorithm, similar to how the II algorithm has a stopping condition, and also how I should best pass the number of operators to the optimize method, as the optimize method itself only deals with the physical plan and not the logical plan. I don't know what difference it would make, but we could also count operations in the physical plan instead of counting them in the logical plan.

Regardless, I've looked into LogicalPlanUtils, LogicalPlanVisitor, and LogicalPlanWalker to see if there's any method implemented which counts operations - as if such a method already exists I wouldn't want to reimplement the wheel, but it doesn't _seem_ like there is one.
If there isn't such a method to count operations, then should I implement it as a member for the Randomized algorithm or should I add it to the LogicalPlanUtils or another package? I'm assuming that a logical plan and its subplans is a tree without cycles, so that I can visit a subplan to see its subplans, and so on, and in that way count the number of operators (are all operators binary?).

hartig:
> I'm thinking of whether I should be implementing an equilibrium condition for the SA algorithm, similar to how the II algorithm has a stopping condition, and also how I should best pass the number of operators to the optimize method, as the optimize method itself only deals with the physical plan and not the logical plan.

Yeah, good idea!

> I don't know what difference it would make, but we could also count operations in the physical plan instead of counting them in the logical plan.

It would not make much difference, assuming you would always be using only the initial physical plan. Other physical plans that are semantically equivalent to the initial one (i.e., "neighbors" and "neighbors' neighbors") may have different numbers of operators. Therefore, to have a constant, the initial physical plan would need to be used. However, getting this number based on the logical plan may be easier in our current implementation because of the visitor pattern that is already implemented for them, as you have found ...

> Regardless, I've looked into LogicalPlanUtils, LogicalPlanVisitor, and LogicalPlanWalker to see if there's any method implemented which counts operations - as if such a method already exists I wouldn't want to reimplement the wheel, but it doesn't seem like there is one.

Good thinking. Thanks for taking initiative!

You are right, we do not yet have such a counting functionality.

> If there isn't such a method to count operations, then should I implement it as a member for the Randomized algorithm or should I add it to the LogicalPlanUtils or another package?

LogicalPlanUtils would be the best place to put this. So, similar to the SourceAssignmentChecker that is already there, you add another class that implements LogicalPlanVisitor and that simply maintains a counter which it increases by one in all of its visit methods. Then, you add another method similar to `isSourceAssignment` and in that method you pick out the counter in the end from the visitor.

> I'm assuming that a logical plan and its subplans is a tree without cycles, so that I can visit a subplan to see its subplans, and so on, and in that way count the number of operators (are all operators binary?).

Exactly.

hartig (commented):
Just one comment on the code at this stage. Everything else looks good and you can continue as discussed in [my recent comment](https://github.com/LiUSemWeb/HeFQUIN/pull/127#issuecomment-993344654).

SHSenpai:
I've ran into a bit of a conundrum.
The equilibrium condition is based on a scaling factor, as in, the equilibrium is a generation count of x times the number of subplans in a logical plan.

The issue is where to store the amount of subplans.
If the number of subplans is to be found when the equilibrium condition is constructed and stored in the condition, then you essentially need a different condition for each plan, and therefore the same optimizer object cannot be used well for different plans. Therefore we must exclude this from consideration.
If the number of subplans is to be found when optimize is ran, we must figure this out when optimize(logicalplan) is ran and then pass it on to a new optimize(physicalplan, numberOfSubplans) method. This would preclude SimulatedAnnealing from implementing the base optimize(physicalplan) method found in RandomizedQueryOptimizerBase.
We could decide this in optimize(PhysicalPlan) by finding a way to get the number of subplans from a physical plan rather than a logical plan, but I do not know how efficient this would be.
We could remove the optimize(PhysicalPlan) method and add its functionality into optimize(LogicalPlan) since optimize(LogicalPlan) just calls optimize(PhysicalPlan) anyway, but since the optimize(PhysicalPlan) method is public I assume you want some means of working directly on a PhysicalPlan from the outside.

What do you think? @hartig

hartig:
Yes, SimulatedAnnealing should implement optimize(logicalplan) by calling a new optimize(physicalplan, numberOfSubplans) method. Hence, it turns out that implementing optimize(logicalplan) in the abstract base class is too generic and does not work.

SHSenpai:
There seems to be something done with the base class in main, I'll leave it to you to resolve the conflict since I'm not clued in to what you've been using the base class for elsewhere.

Since SimulatedAnnealing no longer implements optimize(physicalPlan) but instead has optimize(physicalPlan, numberOfSubplans), the base class no longer declares an abstract optimize(physicalPlan). It's commented out for now.

SA is ready now and II still succeeds with the test. I will create a new branch for 2PO as soon as you've approved the merge.

Overall it's been more work than initially expected as helper methods needed to be made and so on, but it's been really fun to learn the IDE and the language, and I'm looking forward to seeing what you have in store for me once 2PO is completed at the end of this year/start of the next. Merry Christmas, by the way!

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionByRelativeSubplanCount.java
```diff
@@ -0,0 +1,15 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
+
+public class EquilibriumConditionByRelativeSubplanCount implements EquilibriumConditionForSimulatedAnnealing {
+
```
hartig, line 4:
```suggestion
public class EquilibriumConditionByRelativeSubplanCount implements EquilibriumConditionForSimulatedAnnealing
{	
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionByRelativeSubplanCount.java
```diff
@@ -0,0 +1,15 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
+
+public class EquilibriumConditionByRelativeSubplanCount implements EquilibriumConditionForSimulatedAnnealing {
+	
+    protected final int scalingFactor;
+
+    public EquilibriumConditionByRelativeSubplanCount( final int scalingFactor ) {
+        this.scalingFactor = scalingFactor;
+    }
```
hartig, line 9:
```suggestion
	protected final int scalingFactor;

	public EquilibriumConditionByRelativeSubplanCount( final int scalingFactor ) {
		this.scalingFactor = scalingFactor;
	}
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/EquilibriumConditionForSimulatedAnnealing.java
```diff
@@ -0,0 +1,5 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
+
+public interface EquilibriumConditionForSimulatedAnnealing {
+	public boolean isEquilibrium(final int currentGeneration, final int subplanCount);
```
hartig, line 4:
```suggestion
public interface EquilibriumConditionForSimulatedAnnealing
{
	boolean isEquilibrium(int currentGeneration, int subplanCount);
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -23,56 +26,50 @@ public IterativeImprovementBasedQueryOptimizer( final StoppingConditionForIterat
...
+		return optimize( context.getLogicalToPhysicalPlanConverter().convert(initialPlan,false) );
+	}
+	
 	public PhysicalPlan optimize( final PhysicalPlan initialPlan ) throws QueryOptimizationException {
-		// The best plan we have found so far. As we have only found one plan, it is the best one so far.
-		PhysicalPlan bestPlan = initialPlan;
-
-		final Double initialCost = CostEstimationUtils.getEstimates( context.getCostModel(), initialPlan )[0];
+		// The best plan and cost we have found so far. As we have only found one plan, it is the best one so far.
+		PhysicalPlanWithCost bestPlan = new PhysicalPlanWithCost( initialPlan, CostEstimationUtils.getEstimates( context.getCostModel(), initialPlan )[0]);
```
hartig, line 35:
```suggestion
		PhysicalPlanWithCost bestPlan = PhysicalPlanWithCostUtils.annotatePlanWithCost( context.getCostModel(), initialPlan );

```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -23,56 +26,50 @@ public IterativeImprovementBasedQueryOptimizer( final StoppingConditionForIterat
...
 		int generation = 0;
 
-		// bestCost = best possible cost out of everything that the algorithm has found. Has to be declared here so that we have something to compare to in all the iterations of the outer loop without losing track of it between loops.
-		Double bestCost = initialCost;
-
 		while ( !condition.readyToStop(generation) ) { // Currently only handles generation number as a stopping condition!
 			// The randomized plan generator is to be used here. As a temporary measure, the initial plan is used.
-			PhysicalPlan currentPlan = initialPlan; // This variable will hold the plan which is currently being worked on.
-			Double currentCost = initialCost; // The cost of the current plan. For it to be a local minimum, none of its neighbours can have a lower cost
+			PhysicalPlanWithCost currentPlan = new PhysicalPlanWithCost(initialPlan, CostEstimationUtils.getEstimates( context.getCostModel(), initialPlan )[0]); // This variable will hold the plan which is currently being worked on.
```
hartig, line 42:
```suggestion
			PhysicalPlanWithCost currentPlan = PhysicalPlanWithCostUtils.annotatePlanWithCost( context.getCostModel(), initialPlan ); // This variable will hold the plan which is currently being worked on.

```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/RandomizedQueryOptimizerBase.java
```diff
@@ -29,12 +29,9 @@ protected RandomizedQueryOptimizerBase( final QueryOptimizationContext context,
 		rules = new PlanRewritingUtils(rewritingRules);
 	}
 
-	@Override
-	public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException {
-		return optimize( context.getLogicalToPhysicalPlanConverter().convert(initialPlan,false) );
-	}
+	abstract public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException;
```
hartig, line 33:
```suggestion
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/RandomizedQueryOptimizerBase.java
```diff
@@ -29,12 +29,9 @@ protected RandomizedQueryOptimizerBase( final QueryOptimizationContext context,
...
 
-	abstract public PhysicalPlan optimize( PhysicalPlan initialPlan ) throws QueryOptimizationException;
+	// abstract public PhysicalPlan optimize( PhysicalPlan initialPlan ) throws QueryOptimizationException;
```
hartig, line 35:
```suggestion
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/RandomizedQueryOptimizerBase.java
```diff
@@ -47,5 +44,13 @@ public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimi
 
 		return resultList;
 	}
-
+	
+	protected <T> T getRandomElement(List<T> L) {
```
hartig, line 48:
```suggestion
	protected <T> T getRandomElement( final List<T> L ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
```diff
@@ -0,0 +1,83 @@
...
+								final QueryOptimizationContext context,
+								final RuleInstances rewritingRules) {
+		super(context, rewritingRules);
+		condition = x;
+	}
+
+
+	@Override
+	public PhysicalPlan optimize( final LogicalPlan initialPlan ) throws QueryOptimizationException {
+		return optimize( context.getLogicalToPhysicalPlanConverter().convert(initialPlan,false), LogicalPlanUtils.countSubplans(initialPlan) );
```
hartig, line 27:
```suggestion
		final PhysicalPlan initialPP = context.getLogicalToPhysicalPlanConverter().convert(initialPlan,false);
		final int numberOfSubplans = LogicalPlanUtils.countSubplans(initialPlan);
		return optimize(initialPP, numberOfSubplans);
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
```diff
@@ -0,0 +1,74 @@
...
+	public SimulatedAnnealing(	final QueryOptimizationContext context,
+								final RuleInstances rewritingRules) {
+		super(context, rewritingRules);
+	}
+
+	@Override
+	public PhysicalPlan optimize( final PhysicalPlan initialPlan ) throws QueryOptimizationException {
+		
+		// The first plan, which is currently the best plan we know of.
+		PhysicalPlanWithCost bestPlan = new PhysicalPlanWithCost(initialPlan, CostEstimationUtils.getEstimates( context.getCostModel(), initialPlan )[0]);
```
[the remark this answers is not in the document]
hartig:
```suggestion
		PhysicalPlanWithCost bestPlan = PhysicalPlanWithCostUtils.annotatePlanWithCost( context.getCostModel(), initialPlan );
```

hartig (approved):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
```diff
@@ -23,56 +26,52 @@ public IterativeImprovementBasedQueryOptimizer( final StoppingConditionForIterat
 	}
 
 	@Override
```
hartig, line 28:
```suggestion
	// @Override  TODO: fix after merging to main!!
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
```diff
@@ -0,0 +1,85 @@
...
+
+	public SimulatedAnnealing(	final EquilibriumConditionForSimulatedAnnealing x,
+								final QueryOptimizationContext context,
+								final RuleInstances rewritingRules) {
+		super(context, rewritingRules);
+		condition = x;
+	}
+
+
+	@Override
```
hartig, line 25:
```suggestion
	// @Override TODO: fix after merging to main!!
```

hartig:
I have merged this PR now.

Two things are remaining to complete the work on implementing the paper:
1. Change the existing code of `IterativeImprovementBasedQueryOptimizer` to use the randomized join plan optimizer.
2. Implement the third algorithm (2PO).

I suggest you do 1 next. To this end, notice that `RandomizedJoinPlanOptimizerImpl` is only for subplans that consist solely of binary joins. Therefore, you need to create a `SimpleJoinOrderingQueryOptimizer` with a `RandomizedJoinPlanOptimizerImpl` as its join plan optimizer.

PR #128, opened 2021-12-07
Run query with result stream

chengsijin0817:


hartig:
Honestly, this PR doesn't make much sense to me. I assume that you do not actually want to use this new command line program (`RunQueryWithResultInStream`) as such, do you?

What is the purpose of `RunQueryWithResultInStream`? I mean, what do you want to use it for?

chengsijin0817:
> Honestly, this PR doesn't make much sense to me. I assume that you do not actually want to use this new command line program (`RunQueryWithResultInStream`) as such, do you?

I am using this class in the HeFQUIN-Endpoint project. 
[queryResult = new RunQueryWithResultInStream(args).run();](https://github.com/LiUSemWeb/HeFQUIN-Endpoint/blob/792ddd2b61d93cd8378372e7d41f60084b61c4d1/src/main/java/se/liu/ida/hefquin/hefquinwebservice/Controller.java#L72)

> 
> What is the purpose of `RunQueryWithResultInStream`? I mean, what do you want to use it for?

The class 'RunQueryWithResultInStream' allows other applications to easily access our query engine to execute queries and receive results by calling the method RunQueryWithResultInStream(args).run()

Another reason for adding this new class was: when using with RunQueryWithoutSrcSel, the engine exits whenever any query cannot be answered successfully due to some errors, which causes the webserver to stop as well. In this case, no further queries can be processed. In order to process query sequences without stopping the server, I use mainMethod() instead of mainRun() in class 'RunQueryWithResultInStream'. In this way, the exception can be handled outside of HeFQUIN, and I decided to print the error messages in the log files of HeFQUIN-endpoint.

hartig:
@chengsijin0817 Does the commit mentioned in my previous comment address the need that made you work on this PR? If yes, I think this PR has become obsolete, right?

chengsijin0817:
> @chengsijin0817 Does the commit mentioned in my previous comment address the need that made you work on this PR? If yes, I think this PR has become obsolete, right?

Yes, I have updated the implementation of HeFQUIN-Endpoint by using the new helper classes (see [here](https://github.com/LiUSemWeb/HeFQUIN-Endpoint/blob/c9410e9775239542896b91b4b1ae48b724f49310/src/main/java/se/liu/ida/hefquin/hefquinwebservice/Controller.java#L66)).
Therefore, this PR is obsolete. I will close this PR.

PR #130, opened 2021-12-08
Fix an issue in QueryPatternUtils

chengsijin0817:
When getting the number of elements from a SPARQL patterns: org.apache.jena.sparql.algebra.op.OpBGP is not captured by the condition ( queryPattern instanceof BGP), so update the condition to ( queryPattern.asJenaOp() instanceof OpBGP).

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
@@ -86,11 +86,11 @@ public class QueryPatternUtils
 	 * of {@link #getVariablesInPattern(BGP)}.
 	 */
 	public static Set<Var> getVariablesInPattern( final SPARQLGraphPattern queryPattern ) {
-		if ( queryPattern.asJenaOp() instanceof Triple) {
-			return getVariablesInPattern( new TriplePatternImpl( (Triple) queryPattern.asJenaOp()) );
+		if ( queryPattern instanceof TriplePattern ) {
+			return getVariablesInPattern( (TriplePattern) queryPattern );
 		}
-		if ( queryPattern.asJenaOp() instanceof OpBGP) {
-			return getVariablesInPattern( createBGP(((OpBGP) queryPattern.asJenaOp()).getPattern()) );
+		else if ( queryPattern.asJenaOp() instanceof OpBGP) {
+			return getVariablesInPattern( (OpBGP) queryPattern.asJenaOp() );
 		}
 		else {
 			throw new UnsupportedOperationException("Getting the variables from arbitrary SPARQL patterns is an open TODO (type of Jena Op in the current case: " + queryPattern.asJenaOp().getClass().getName() + ").");
@@ -98,8 +98,12 @@ public class QueryPatternUtils
 	}
 
 	public static Set<Var> getVariablesInPattern( final TriplePattern tp ) {
+		return getVariablesInPattern( tp.asJenaTriple() );
+	}
+
+	public static Set<Var> getVariablesInPattern( final Triple tp ) {
 		final Set<Var> result = new HashSet<>();
-		Vars.addVarsFromTriple( result, tp.asJenaTriple() );
+		Vars.addVarsFromTriple( result, tp );
 		return result;
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
```diff
@@ -170,18 +171,18 @@ public static int getNumberOfBNodeOccurrences( final BGP bgp ) {
...
+	 *
 	 * If the given pattern is a {@link TriplePattern}, this function returns
 	 * the result of {@link #getNumberOfVarOccurrences(TriplePattern)}.
 	 * Similarly, if the given pattern is a {@link BGP}, this function
 	 * returns the result of {@link #getNumberOfVarOccurrences(BGP)}.
 	 */
 	public static int getNumberOfVarOccurrences( final SPARQLGraphPattern queryPattern ) {
-		if ( queryPattern instanceof TriplePattern ) {
-			return getNumberOfVarOccurrences( (TriplePattern) queryPattern );
+		if ( queryPattern.asJenaOp() instanceof Triple) {
```
hartig, line 181: This will never be true. `queryPattern.asJenaOp()` returns a Jena `Op` object but Jena's `Triple` class does not implement `Op`. Probably you meant `OpTriple` instead?

Something else: I would also like to keep the original if-branch (i.e., `queryPattern instanceof TriplePattern`) because that fires immediately for `TriplePattern` objects. In fact, if you look at `TriplePatternImpl`, you will notice that `asJenaOp()` is not even implemented there.

src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
```diff
@@ -170,18 +171,18 @@ public static int getNumberOfBNodeOccurrences( final BGP bgp ) {
...
+			return getNumberOfVarOccurrences( new TriplePatternImpl( (Triple) queryPattern.asJenaOp()) );
```
hartig, line 182: No! You should not create a new `TriplePatternImpl` object only for the purpose of calling that method. Instead, add a version of this method that accepts a Jena `Triple` object (and, after doing so, the version that expects a `TriplePattern` object can internally call that new method).

src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
```diff
@@ -170,18 +171,18 @@ public static int getNumberOfBNodeOccurrences( final BGP bgp ) {
...
 		}
-		else if ( queryPattern instanceof BGP ) {
-			return getNumberOfVarOccurrences( (BGP) queryPattern );
+		else if ( queryPattern.asJenaOp() instanceof OpBGP) {
+			return getNumberOfVarOccurrences( createBGP(((OpBGP) queryPattern.asJenaOp()).getPattern()) );
```
hartig, line 185: Same comment here. Do not create a new `BGP` object only for the purpose of calling this method.

src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
```diff
@@ -192,18 +193,18 @@ else if ( queryPattern instanceof BGP ) {
...
 	public static int getNumberOfTermOccurrences( final SPARQLGraphPattern queryPattern ) {
-		if ( queryPattern instanceof TriplePattern ) {
-			return getNumberOfTermOccurrences( (TriplePattern) queryPattern );
+		if ( queryPattern.asJenaOp() instanceof Triple) {
+			return getNumberOfTermOccurrences( new TriplePatternImpl( (Triple) queryPattern.asJenaOp())  );
 		}
-		else if ( queryPattern instanceof BGP ) {
-			return getNumberOfTermOccurrences( (BGP) queryPattern );
+		else if ( queryPattern.asJenaOp() instanceof OpBGP) {
+			return getNumberOfTermOccurrences( createBGP(((OpBGP) queryPattern.asJenaOp()).getPattern()) );
```
hartig, line 207: Same comments as above.

hartig (changes requested):

PR #131, opened 2021-12-08
Fix an issue in Greedy Algorithm

chengsijin0817:
"subplans.remove (indexOfBestPlan);" throw an UnsupportedOperationException.

Hence, change 'subplans' from a static List to a LinkedList


hartig (changes requested):
I understand the issue. However, I don't think that this is a good fix because other implementations of `JoinPlanOptimizer` may run into the exact same problem. Therefore, instead of this fix, I propose you change the default implementation of the method in lines 19-21 in `JoinPlanOptimizer`:
```
default PhysicalPlan determineJoinPlan( PhysicalPlan[] subplans ) throws QueryOptimizationException {
	return determineJoinPlan( new ArrayList<>(Arrays.asList(subplans)) );
}
```

hartig (approved):

PR #134, opened 2021-12-09
Simulated Annealing Query Optimization Algorithm

SHSenpai:
Simulated Annealing almost done, but I am at a loss for how to implement Equilibrium. It's supposed to be a boolean from what I can read in the paper, but the paper's example implementation defines it to be 16*(number or joins in query). (table 2.2, page 314 in the paper, page 3 in the pdf)

I also have some thoughts (see comments in code) of a small 'issue' (whose consequences are just that a couple of code lines appear to be inefficient in its implementation) concerning the selection of a random neighbour and the creation of PhysicalPlanWithCost objects. Since my proposed solution is to introduce a new constructor for PhysicalPlanWithCost, your input is necessary. @hartig

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/SimulatedAnnealing.java
@@ -34,45 +34,44 @@ public class SimulatedAnnealing extends RandomizedQueryOptimizerBase {
 			
 			// TODO: Define what equilibrium entails on a practical level
 			// The paper's implementation counts equilibrium as 16*(number of joins in query).
-			// While not equilibrium do
+			while(!isEquilibrium()) {
 			
-			// Here we want a random neighbour, and its cost
-			// My first instinct to do something like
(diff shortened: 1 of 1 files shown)


SHSenpai:
To clarify the issue with PhysicalPlanWithCost: When creating the object, it requires me to have the cost of the plan when I create the PhysicalPlanWithCost, so I currently need to create a temporary plan variable, get its cost, and then put them together into PhysicalPlanWithCost. My proposed alternative is to add a constructor to PhysicalPlanWithCost which takes just a PhysicalPlan and gets its cost as part of the constructor.

hartig:
I have tried to take a look at the code in this PR but the PR seems to only contain code changes over some earlier implementation of the `SimulatedAnnealing` class. That makes it very hard for me to see what you have done here (at least if I look only in the Github Web interface, which is where I have to look to put review comments). Since the `SimulatedAnnealing` class is new, don't you have a PR for me that introduces this class (in the version that you consider ready for review)?

hartig:
> To clarify the issue with PhysicalPlanWithCost: When creating the object, it requires me to have the cost of the plan when I create the PhysicalPlanWithCost, so I currently need to create a temporary plan variable, get its cost, and then put them together into PhysicalPlanWithCost. My proposed alternative is to add a constructor to PhysicalPlanWithCost which takes just a PhysicalPlan and gets its cost as part of the constructor.

No. I would not want to have such a constructor; in particular because the process to get the cost may be time consuming. I see the `PhysicalPlanWithCost` class really just as a container without any functionality (except for the functionality to return the components of this container).

PR #135, opened 2021-12-09
Interative Improvement Query Optimization Algorithm

SHSenpai:
@hartig The test gets an error (which is different from it just failing the test, right?) when I try a basic use of the Randomizer.
How do I debug errors in the test?

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
@@ -6,8 +6,9 @@ import se.liu.ida.hefquin.engine.queryplan.PhysicalPlan;
 import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationContext;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RuleInstances;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.simple.RandomizedJoinPlanOptimizerImpl;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.CostEstimationUtils;
-import se.liu.ida.hefquin.engine.utils.Pair;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.PhysicalPlanWithCost;
 
 public class IterativeImprovementBasedQueryOptimizer extends RandomizedQueryOptimizerBase
 {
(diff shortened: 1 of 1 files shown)


hartig:
I notice that you want to do two things in this PR:
1. change the existing code to use `PhysicalPlanWithCost`
2. change the existing code to use the randomized join plan optimizer

I would prefer if you focus on only one of these two things first. Really, just do only one thing at a time! That way, the PR will be smaller, which makes it easier to develop (i.e., for you) and easier to review (i.e., for me). I suggest you start with using `PhysicalPlanWithCost`. Hence, for everything in this PR where you try to use `RandomizedJoinPlanOptimizerImpl`, just revert this back to what it looked like before and keep only the changes for `PhysicalPlanWithCost`. Once these changes are okay and merged, you can create another PR in which you then focus on using `RandomizedJoinPlanOptimizerImpl`.

Regarding the use of `PhysicalPlanWithCost`, notice that we have a helper class called `PhysicalPlanWithCostUtils`. This class contains a method called `annotatePlansWithCost`. Please use this method instead of, first, calling `CostEstimationUtils.getEstimates` and, then, creating the `PhysicalPlanWithCost` objects (`annotatePlansWithCost` takes care of these things for you).

To your questions:

> The test gets an error (which is different from it just failing the test, right?) [...]

Yes, an error in a test is different from a test failure. An error means that something unexpected has happened; usually, an uncaught exception that should not have occurred. In contrast, if a test fails, it means that the tested code has not produced the expected result and, hence, did not exhibit the expected behavior. The purpose of the unit tests is to check that the code does exhibit the expected behavior.

> How do I debug errors in the test?

If you right-click on the test method, the pop-up menu does not only contain the entry 'Run As' but also 'Debug As', which allows you to initiate a debug run of the unit test. Before starting such a debug run, you may want to place breakpoints in the code. To this end, open the source file and go to the code line for which you want the breakpoint. Next, in this subwindow/panel that contains the source code, right-click close to the left-hand-side border and the pop-up menu allows you to 'Toggle Breakpoint'. Alternatively, move the cursor to the relevant source code line and open the drop-down menu called 'Run' of the main menu; there you also find 'Toggle Breakpoint'. Now, after setting a breakpoint and starting the debug run, the run will break at the breakpoint and you can step through the code or resume its automatic execution until the next breakpoint.

SHSenpai:
> I would prefer if you focus on only one of these two things first. Really, just do only one thing at a time!
Gotcha. Sorry for trying to do multiple things, I was just worried about not being able to complete everything in time.

This is the change to use PhysicalPlanWithCost. It uses InitialPlan instead of Randomized (will make a new PR for that one). The test runs and succeeds. @hartig

hartig (approved):
Looks good. I will merge this PR now.

hartig:
I have merged the PR and delete the branch. You can create a new branch (from main!) to work on using the randomized join plan optimizer.

Some things to take into account for this next step:
* Notice that the randomized join plan optimizer (`RandomizedJoinPlanOptimizerImpl`) is really just for join plans (rather than for complete plans). You have to use it within `SimpleJoinOrderingQueryOptimizer` (i.e., pass a `RandomizedJoinPlanOptimizerImpl` to the constructor of `SimpleJoinOrderingQueryOptimizer`).
* Moreover, assuming you will call `optimizePlan( final PhysicalPlan plan )` of the `SimpleJoinOrderingQueryOptimizer`, notice that this function assumes that all the joins in the given `PhysicalPlan` are multi-way joins rather than binary joins. Hence, to satisfy this assumption, when translating the given logical plan to the initial physical plan (by calling `context.getLogicalToPhysicalPlanConverter().convert(...)`) you need to pass `true` as the second argument of this `convert` function.

Let me know if any of these two points is unclear.

PR #136, opened 2021-12-09
Cypher record translation

scferrada:
I just added three tests, to see if my testing idea is okay. If approved, I will add the rest of the tests for all other cases.

Tests are currently using a truncated version of the actual output of the Cypher query executed in Neo4j.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
@@ -13,73 +13,90 @@ import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.RecordEntry;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.TableRecord;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.Value;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.*;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherQuery;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherVar;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.utils.CypherUtils;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.utils.CypherVarGenerator;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
 public class Record2SolutionMappingTranslatorImpl implements Record2SolutionMappingTranslator {
+
     @Override
-    public List<SolutionMapping> translateRecord(final TableRecord record, final LPG2RDFConfiguration conf) {
+    public List<SolutionMapping> translateRecord(final TableRecord record, final LPG2RDFConfiguration conf,
+                                                 final CypherQuery query, final CypherVarGenerator generator) {
         final List<BindingBuilder> builders = new ArrayList<>();
         final BindingBuilder baseBuilder = Binding.builder();
         for (final RecordEntry entry : record.getRecordEntries()){
             final CypherVar var = entry.getName();
             final Value value = entry.getValue();
+            final Var mappingVar = Var.alloc(generator.getReverseRetVarName(var));
             if (value instanceof LPGNodeValue) {
                 final Node nodeMapping  = conf.mapNode(((LPGNodeValue) value).getNode());
                 if (builders.isEmpty())
-                    baseBuilder.add(Var.alloc(var.getName()), nodeMapping);
+                    baseBuilder.add(mappingVar, nodeMapping);
                 else {
-                    addToAllBuilders(builders, Var.alloc(var.getName()), nodeMapping);
+                    addToAllBuilders(builders, mappingVar, nodeMapping);
                 }
             } else if (value instanceof LPGEdgeValue) {
                 final Node edgeMapping = conf.mapEdgeLabel(((LPGEdgeValue) value).getEdge().getLabel());
                 if (builders.isEmpty())
-                    baseBuilder.add(Var.alloc(var.getName()), edgeMapping);
+                    baseBuilder.add(mappingVar, edgeMapping);
                 else {
-                    addToAllBuilders(builders, Var.alloc(var.getName()), edgeMapping);
+                    addToAllBuilders(builders, mappingVar, edgeMapping);
                 }
             } else if (value instanceof MapValue) {
                 final Map<String, Object> mapValue = ((MapValue) value).getMap();
                 final Node triple = NodeFactory.createTripleNode(
                         conf.mapNode((LPGNode) mapValue.get("source")),
-                        conf.mapEdgeLabel(((LPGEdge)mapValue.get("edge")).getLabel()),
+                        conf.mapEdgeLabel(((LPGEdge) mapValue.get("edge")).getLabel()),
                         conf.mapNode((LPGNode) mapValue.get("target"))
                 );
                 if (builders.isEmpty())
-                    baseBuilder.add(Var.alloc(var.getName()), triple);
+                    baseBuilder.add(mappingVar, triple);
                 else {
-                    addToAllBuilders(builders, Var.alloc(var.getName()), triple);
+                    addToAllBuilders(builders, mappingVar, triple);
                 }
             } else if (value instanceof ListValue) {
                 final List<Object> values = ((ListValue) value).getList();
+                final boolean isPropertyList = CypherUtils.isPropertyColumn(query, var);
                 if (builders.isEmpty()) {
                     for (final Object val : values) {
-                        Node node = getPropertyOrLiteral(val, conf);
+                        Node node;
+                        if (isPropertyList) {
+                            node = conf.mapProperty(val.toString());
+                        } else {
+                            node = NodeFactory.createLiteral(val.toString());
+                        }
                         final BindingBuilder builder = Binding.builder();
                         builder.addAll(baseBuilder.build());
-                        //we still need to distinguish properties from values here
-                        builder.add(Var.alloc(var.getName()), NodeFactory.createLiteral(val.toString()));
+                        builder.add(mappingVar, node);
                         builders.add(builder);
                     }
                 } else {
                     for (final Object val : values) {
-                        addToAllBuilders(builders, Var.alloc(var.getName()), NodeFactory.createLiteral(val.toString()));
+                        Node node;
+                        if (isPropertyList) {
+                            node = conf.mapProperty(val.toString());
+                        } else {
+                            node = NodeFactory.createLiteral(val.toString());
+                        }
+                        addToAllBuilders(builders, mappingVar, node);
                     }
                 }
             } else if (value instanceof LiteralValue) {
                 final String literal = ((LiteralValue) value).getValue().toString();
-                Node node = null;
+                Node node;
                 if (literal.equals("label"))
                     node = conf.getLabel();
                 else
                     node = NodeFactory.createLiteral(literal);
                 if (builders.isEmpty()) {
-                    baseBuilder.add(Var.alloc(var.getName()), node);
+                    baseBuilder.add(mappingVar, node);
                 } else {
-                    addToAllBuilders(builders, Var.alloc(var.getName()), node);
+                    addToAllBuilders(builders, mappingVar, node);
                 }
             }
         }
@@ -91,10 +108,6 @@ public class Record2SolutionMappingTranslatorImpl implements Record2SolutionMapp
             mappings.add(new SolutionMappingImpl(builder.build()));
         }
         return mappings;
-    }
-
-    private Node getPropertyOrLiteral(final Object value, final LPG2RDFConfiguration conf) {
-        return null;
     }
 
     private void addToAllBuilders(final List<BindingBuilder> builders, final Var var, final Node node) {
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
@@ -110,6 +110,17 @@ public class Record2SolutionMappingTranslatorImpl implements Record2SolutionMapp
         return mappings;
     }
 
+    @Override
+    public List<SolutionMapping> translateRecords(final List<TableRecord> records,
+                                                  final LPG2RDFConfiguration conf, final CypherQuery query,
+                                                  final CypherVarGenerator generator) {
+        List<SolutionMapping> results = new ArrayList<>();
+        for (final TableRecord record : records) {
+            results.addAll(translateRecord(record, conf, query, generator));
+        }
+        return results;
+    }
+
     private void addToAllBuilders(final List<BindingBuilder> builders, final Var var, final Node node) {
         for (final BindingBuilder builder : builders)
             builder.add(var, node);
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
@@ -1,0 +1,99 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import com.github.jsonldjava.core.RDFDataset;
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.Binding;
+import org.apache.jena.sparql.engine.binding.BindingBuilder;
+import org.apache.jena.sparql.engine.binding.BindingFactory;
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.data.impl.SolutionMappingImpl;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.LPG2RDFConfiguration;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.Record2SolutionMappingTranslator;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.RecordEntry;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.TableRecord;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.Value;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.*;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherVar;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class Record2SolutionMappingTranslatorImpl implements Record2SolutionMappingTranslator {
+    @Override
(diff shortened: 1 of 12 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherUtils.java
```diff
@@ -50,4 +51,28 @@
         return records;
     }
 
+    public static boolean isPropertyColumn(final CypherQuery query, final CypherVar colName) {
```
hartig, line 54: Please add a JavaDoc comment with a brief sentence about the purpose/functionality of this method.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherUtils.java
```diff
@@ -50,4 +51,28 @@
...
+        if (query instanceof CypherMatchQuery) {
+            return isPropertyColumnPriv((CypherMatchQuery) query, colName);
+        }
+        else if (query instanceof CypherUnionQuery) {
+            for (final CypherMatchQuery q : ((CypherUnionQuery) query).getUnion()) {
+                if (isPropertyColumnPriv(q, colName)) {
+                    return true;
+                }
+            }
+        }
```
hartig, line 64: Please add an else-branch that throws an IllegalArgumentException.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherUtils.java
```diff
@@ -50,4 +51,28 @@
...
+        return false;
+    }
+
+    protected static boolean isPropertyColumnPriv(CypherMatchQuery query, CypherVar colName) {
```
hartig, line 68:
```suggestion
    protected static boolean isPropertyColumnPriv( final CypherMatchQuery query, final CypherVar colName ) {
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherVarGenerator.java
```diff
@@ -0,0 +1,74 @@
...
+        if (var != null)
+            return var;
+        retVarCount++;
+        var = new CypherVar(retPrefix + retVarCount);
+        retVars.put(n.getName(), var);
+        return var;
+    }
+
+    public String getReverseRetVarName(final CypherVar var) {
+        for (final Map.Entry<String, CypherVar> entry : retVars.entrySet()) {
```
hartig, line 67: Instead of iterating over all entries, add another Map to the class that you can use as a reverse lookup index.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherVarGenerator.java
```diff
@@ -0,0 +1,74 @@
...
+        CypherVar var = retVars.get(n.getName());
+        if (var != null)
+            return var;
+        retVarCount++;
+        var = new CypherVar(retPrefix + retVarCount);
+        retVars.put(n.getName(), var);
+        return var;
+    }
+
+    public String getReverseRetVarName(final CypherVar var) {
```
hartig, line 66: Why don't you return `Node` objects here?

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -16,14 +16,26 @@
...
+    protected final CypherVarGenerator generator;
+
+    public SPARQLStar2CypherTranslatorImpl() {
+        this.generator = new CypherVarGenerator();
+    }
+
     @Override
     public CypherQuery translateTriplePattern(final TriplePattern tp, final LPG2RDFConfiguration conf) {
-        return translateTriplePattern(tp, conf, new CypherVarGenerator(), new HashSet<>());
+        return translateTriplePattern(tp, conf, this.generator, new HashSet<>());
```
hartig, line 33: I would prefer this class to remain stateless. The new implementation with the `generator` member variable is problematic because, for every translation, you need to create a new object of this class because you need a fresh `CypherVarGenerator`. Hence, this prevents you from being able to reuse an object of this class for many translations, which is not what I would assume from looking just at the implemented interface (`SPARQLStar2CypherTranslator`).

Here is how you can avoid this issue: revert this class back to how it was before and, then, change the return type of this method (in the interface) to be `Pair<CypherQuery,CypherVarGenerator>`. That way, you can create a new `CypherVarGenerator` for every translation (exactly as it was done before). In fact, you may even change the return type to be `Pair<CypherQuery, Map<CypherVar,Node>>` where this `Map<CypherVar,Node>` is populated inside the `CypherVarGenerator` and, in the end, can be obtained from the `CypherVarGenerator` and returned here.

This `Map<CypherVar,Node>` can then be passed to the methods of `Record2SolutionMappingTranslator` (instead of passing the whole `CypherVarGenerator`).

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,128 @@
...
+                if (literal.equals("label"))
+                    node = conf.getLabel();
+                else
+                    node = NodeFactory.createLiteral(literal);
+                if (builders.isEmpty()) {
+                    baseBuilder.add(mappingVar, node);
+                } else {
+                    addToAllBuilders(builders, mappingVar, node);
+                }
+            }
```
hartig, line 101: There is no else-branch here. That's not good. Never assume that, only because the else-if branches cover all the cases that currently exist, there won't be any problem. Instead, always assume that further cases may be added in the future and, therefore, we need to put safeguards into place that give us explicit errors if we forget to cover these future cases. There are two ways to do this: First, you convert the last else-if into an else branch; that way, we would see an exception when cast `value` to a `LiteralValue`. Second, add an else branch in which you throw an IllegalArgumentException.

As a general rule: always make sure you have an else branch after else-if branches.

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/Record2SolutionMappingTranslationTest.java
```diff
@@ -0,0 +1,332 @@
...
+    final LPGNode node14 = new LPGNode("14", null, null);
+    final LPGNode node22 = new LPGNode("22", null, null);
+
+    @Test
+    public void translateVarPropertyLiteralTest() throws JsonProcessingException, Neo4JException {
+        final LPG2RDFConfiguration conf = new DefaultConfiguration();
+        final Triple t = new Triple(Var.alloc("s"), conf.mapProperty("name"),
+                NodeFactory.createLiteral("Lana Wachowski"));
+        final SPARQLStar2CypherTranslator translator = new SPARQLStar2CypherTranslatorImpl();
+        final CypherQuery translation = translator.translateTriplePattern(new TriplePatternImpl(t), conf);
```
hartig, line 47: I would prefer if, for this tests of `Record2SolutionMappingTranslation`, you do not rely on the functionality of the `SPARQLStar2CypherTranslator`. So, remove all of this stuff and, instead, simply populate a `CypherVarGenerator` for this test. Or, even better, populate a `Map<CypherVar,Node>` without even using a `CypherVarGenerator`.

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/Record2SolutionMappingTranslationTest.java
```diff
@@ -0,0 +1,332 @@
...
+                    "\"meta\":[{\"id\":6,\"type\":\"node\",\"deleted\":false}]}]}]," +
+                "\"errors\":[]}";
+        final List<TableRecord> records = CypherUtils.parse(neo4jResponse);
+        final List<SolutionMapping> solMaps = new Record2SolutionMappingTranslatorImpl()
+                .translateRecords(records, conf, translation, generator);
+        final BindingBuilder builder = Binding.builder();
+        builder.add(Var.alloc("s"), conf.mapNode(node6));
+        final SolutionMapping m1 = new SolutionMappingImpl(builder.build());
+        assertEquals(solMaps.size(), 1);
+        assertEquals(solMaps.get(0), m1);
```
hartig, line 60:
```suggestion
        assertEquals(1, solMaps.size());
        assertEquals(m1, solMaps.get(0));
```
The first argument is always the expected value and the second argument is the value to be tested!

hartig (changes requested):

scferrada:
I applied the suggestions, and just modified the first three tests. Upon aproval, I will finish the rest of the tests. I might also remove unnecesary data from the JSON test data, since we don't need all the properties and such.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+                    node = conf.getLabel();
+                else
+                    node = NodeFactory.createLiteral(literal);
+                if (builders.isEmpty()) {
+                    baseBuilder.add(mappingVar, node);
+                } else {
+                    addToAllBuilders(builders, mappingVar, node);
+                }
+            } else {
+                throw new IllegalArgumentException("Unsupported implementation of Value");
```
hartig, line 101:
```suggestion
                throw new IllegalArgumentException("Unsupported implementation of value (" + value.getClass().getName() + ")");
```
Including the name of the unexpected type helps if this issue actually occurs ;-)

hartig (commented):
I have only two minor suggestions. Looks good now. Please continue.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public List<SolutionMapping> translateRecord(final TableRecord record, final LPG2RDFConfiguration conf,
+                                                 final CypherQuery query, final Map<CypherVar, Node> varMap) {
+        final List<Pair<CypherVar, ListValue>> lists = new ArrayList<>();
+        SolutionMapping base = translateRecordsPriv(record, conf, query, varMap, lists);
+        return manageLists(base, lists, conf, query, varMap);
+    }
+
+    private List<SolutionMapping> manageLists(final SolutionMapping base, final List<Pair<CypherVar, ListValue>> lists,
```
hartig, line 36:
```suggestion
    protected List<SolutionMapping> manageLists(final SolutionMapping base, final List<Pair<CypherVar, ListValue>> lists,
```

hartig: Additionally, please include a comment that summarizes the purpose of this method.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+                throw new IllegalStateException("Record List Values should be the same size");
+        }
+        for (int i = 0; i < N; i++) {
+            if (builders.size() < (i + 1)) {
+                final BindingBuilder builder = Binding.builder();
+                builder.addAll(base.asJenaBinding());
+                builders.add(builder);
+            }
+            for (Pair<CypherVar, ListValue> pair : lists) {
+                Node node;
```
hartig, line 54:
```suggestion
                final Node node;
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+                if (CypherUtils.isPropertyColumn(query, pair.object1)) {
+                    node = conf.mapProperty(pair.object2.getList().get(i).toString());
+                } else {
+                    node = NodeFactory.createLiteral(pair.object2.getList().get(i).toString());
+                }
+                builders.get(i).add(Var.alloc(varMap.get(pair.object1)), node);
+            }
+        }
+        List<SolutionMapping> results = new ArrayList<>(builders.size());
```
hartig, line 63:
```suggestion
        final List<SolutionMapping> results = new ArrayList<>(builders.size());
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        for (final BindingBuilder builder : builders) {
+            results.add(new SolutionMappingImpl(builder.build()));
+        }
+        return results;
+    }
+
+    private SolutionMapping translateRecordsPriv(final TableRecord record, final LPG2RDFConfiguration conf,
```
hartig, line 70:
```suggestion
    protected SolutionMapping translateRecordsPriv(final TableRecord record, final LPG2RDFConfiguration conf,
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        return new SolutionMappingImpl(baseBuilder.build());
+    }
+
+    @Override
+    public List<SolutionMapping> translateRecords(final List<TableRecord> records,
+                                                  final LPG2RDFConfiguration conf, final CypherQuery query,
+                                                  final Map<CypherVar, Node> varMap) {
+        List<SolutionMapping> results = new ArrayList<>();
```
hartig, line 118:
```suggestion
        final List<SolutionMapping> results = new ArrayList<>();
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        for (final TableRecord record : records) {
+            results.addAll(translateRecord(record, conf, query, varMap));
+        }
+        return results;
+    }
+
+    private void addToAllBuilders(final List<BindingBuilder> builders, final Var var, final Node node) {
```
hartig, line 125:
```suggestion
    protected void addToAllBuilders(final List<BindingBuilder> builders, final Var var, final Node node) {
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherUtils.java
```diff
@@ -50,4 +58,87 @@
...
+                    return true;
+                }
+            }
+        } else {
+            throw new IllegalArgumentException("Unsupported implementation of Cypher Query (" + query.getClass().getName() +")");
+        }
+        return false;
+    }
+
+    protected static boolean isPropertyColumnPriv( final CypherMatchQuery query, final CypherVar colName ) {
```
hartig, line 82:
```suggestion
            return isPropertyColumn( (CypherMatchQuery) query, colName );
        }
        else if (query instanceof CypherUnionQuery) {
            for (final CypherMatchQuery q : ((CypherUnionQuery) query).getUnion()) {
                if (isPropertyColumn(q, colName)) {
                    return true;
                }
            }
        }
        else {
            throw new IllegalArgumentException("Unsupported implementation of Cypher Query (" + query.getClass().getName() +")");
        }
        return false;
    }

    public static boolean isPropertyColumn( final CypherMatchQuery query, final CypherVar colName ) {
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherUtils.java
```diff
@@ -50,4 +58,87 @@
...
+                    return true;
+                }
+            }
+        } else {
+            throw new IllegalArgumentException("Unsupported implementation of Cypher Query (" + query.getClass().getName() +")");
+        }
+        return false;
+    }
+
+    private static boolean isLabelColumnPriv(final CypherMatchQuery query, final CypherVar colName) {
```
hartig, line 109:
```suggestion
            return isLabelColumn( (CypherMatchQuery) query, colName );
        }
        else if (query instanceof CypherUnionQuery) {
            for (final CypherMatchQuery q : ((CypherUnionQuery) query).getUnion()) {
                if (isLabelColumn(q, colName)) {
                    return true;
                }
            }
        }
        else {
            throw new IllegalArgumentException("Unsupported implementation of Cypher Query (" + query.getClass().getName() +")");
        }
        return false;
    }

    public static boolean isLabelColumn(final CypherMatchQuery query, final CypherVar colName) {
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/CypherUtils.java
```diff
@@ -50,4 +58,87 @@
...
+                    return true;
+                }
+            }
+        } else {
+            throw new IllegalArgumentException("Unsupported implementation of Cypher Query (" + query.getClass().getName() +")");
+        }
+        return false;
+    }
+
+    private static boolean isRelationshipTypeColumnPriv(final CypherMatchQuery query, final CypherVar colName) {
```
hartig, line 135:
```suggestion
            return isRelationshipTypeColumn( (CypherMatchQuery) query, colName );
        }
        else if (query instanceof CypherUnionQuery) {
            for (final CypherMatchQuery q : ((CypherUnionQuery) query).getUnion()) {
                if (isRelationshipTypeColumn(q, colName)) {
                    return true;
                }
            }
        }
        else {
            throw new IllegalArgumentException("Unsupported implementation of Cypher Query (" + query.getClass().getName() +")");
        }
        return false;
    }

    public static boolean isRelationshipTypeColumn(final CypherMatchQuery query, final CypherVar colName) {
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class Record2SolutionMappingTranslatorImpl implements Record2SolutionMappingTranslator {
+
+    @Override
+    public List<SolutionMapping> translateRecord(final TableRecord record, final LPG2RDFConfiguration conf,
+                                                 final CypherQuery query, final Map<CypherVar, Node> varMap) {
+        final List<Pair<CypherVar, ListValue>> lists = new ArrayList<>();
+        SolutionMapping base = translateRecordsPriv(record, conf, query, varMap, lists);
```
hartig, line 32:
```suggestion
        final SolutionMapping base = translateRecordsPriv(record, conf, query, varMap, lists);
```

hartig (changes requested):
Just a few minor things.

scferrada:
Hi, the requests were addressed. There is still the issue I commented by mail, where a query returns irrelevant results with empty lists. But I think we need to discuss that separately.

On the next PR I will be adding the translation of BGPs to Cypher.

hartig (approved):

hartig:
Thanks for the updates. I have merged this PR.

The issue discussed in the email can be addressed separately (i.e., in a separate PR).

PR #137, opened 2021-12-10
Fix issues in query pattern utils

chengsijin0817:
Support getting the number of elements (RDF terms/ variables) from SPARQL patterns (contain AND and FILTER)

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
@@ -10,6 +10,7 @@ import org.apache.jena.graph.Triple;
 import org.apache.jena.sparql.algebra.OpAsQuery;
 import org.apache.jena.sparql.algebra.OpVars;
 import org.apache.jena.sparql.algebra.op.OpBGP;
+import org.apache.jena.sparql.algebra.op.OpJoin;
 import org.apache.jena.sparql.core.BasicPattern;
 import org.apache.jena.sparql.core.PathBlock;
 import org.apache.jena.sparql.core.TriplePath;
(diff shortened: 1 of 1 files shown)


src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
```diff
@@ -179,10 +208,56 @@ public static int getNumberOfVarOccurrences( final OpBGP bgp ) {
...
+	public static int getNumberOfVarOccurrences( final Op op ) {
+		if ( op instanceof OpBGP ) {
+			return getNumberOfVarOccurrences( (OpBGP) op);
+		}
+		else if ( op instanceof OpJoin ) {
+			return getNumberOfVarOccurrences( (OpJoin) op );
+		}
+		else if ( op instanceof OpFilter )  {
+			// TODO
+			return 0;
```
hartig, line 227: What is the "TODO" supposed to mean? Does it mean that this case still needs to be implemented? If so, please throw an UnsupportedOperationException instead of returning a number.

src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
```diff
@@ -179,10 +208,56 @@ public static int getNumberOfVarOccurrences( final OpBGP bgp ) {
...
+	public static int getNumberOfTermOccurrences( final Op op ) {
+		if ( op instanceof OpBGP ) {
+			return getNumberOfTermOccurrences( (OpBGP) op);
+		}
+		else if ( op instanceof OpJoin ) {
+			return getNumberOfTermOccurrences( (OpJoin) op );
+		}
+		else if ( op instanceof OpFilter )  {
+			// TODO
+			return 0;
```
hartig, line 254: Same question here: what's it with this TODO?

hartig (commented):
I am generally fine with this PR, except for one question that I have.

src/main/java/se/liu/ida/hefquin/engine/query/impl/QueryPatternUtils.java
chengsijin0817: I marked "TODO" here, to indicate that this number is still needed to be determined.

Reason: I was unsure whether we need to take into account the number of elements (variables and terms) in a FILTER expression in the cost model.
Returning 0 means that the variables and RDF terms in a filter expression are not considered in the total number of shipping data.

hartig: I see. The question then is, what would you say? More precisely, from the perspective of the current cost model, does it make sense to include the number of elements (variables and terms) of a FILTER in the count for the corresponding cost dimensions?

hartig (approved):

PR #138, opened 2021-12-16
Add an enhanced physical plan printer

chengsijin0817:
Print physical plans with graph pattern and federation member of each operator.


hartig (changes requested):
Instead of creating a new class, can you please add these extensions directly into the existing `PhysicalPlanPrinter` class. We don't need two such classes.

Additionally, instead of interleaving the `visit` methods with the helper methods that they use, please move all the helper methods after all the `visit` methods.   ...or, perhaps, now that I think about it. I would say that we shouldn't have these helper methods here at all. Instead, each physical operator (and also each logical operator) should override the `toString()` method and that's where this helper code should be placed. Please do that.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanUtils.java
```diff
@@ -55,4 +58,43 @@ public void visit( final LogicalOpUnion op ) {
...
 
+	public static void printStringOfFm( final StringBuilder builder, final FederationMember fm ) {
+
+		final DataRetrievalInterface intFace = fm.getInterface();
+		if ( intFace instanceof SPARQLEndpointInterface){
+			builder.append( ((SPARQLEndpointInterface) intFace).getURL() );
+		}
+		else {
+			builder.append( " Print the Federation Member in the type of interface: " + intFace.getClass().getName() + "is an open TODO" );
+		}
```
hartig, line 69: In addition to the URL, the string that is produced here should also contain an indication of the type of fed.member/interface. For instance, possible example strings could be:

"SPARQL endpoint at http://dbpedia.org/sparql"

"TPF server at TODO"

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanUtils.java
```diff
@@ -55,4 +58,43 @@ public void visit( final LogicalOpUnion op ) {
...
+
+	}
+
+	public static void printTriplesOfGraphPattern(final Op op, final StringBuilder builder ) {
```
hartig, line 73: Did you check whether Jena already has something for printing `Op` objects? If so, I would prefer to simply (re)use that one.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpBGPAdd.java
```diff
@@ -45,4 +48,18 @@ public void visit( final LogicalPlanVisitor visitor ) {
 		visitor.visit(this);
 	}
 
+	@Override
+	public String toString(){
+		final StringBuilder builder = new StringBuilder();
+
+		builder.append("> bgpAdd ");
+		builder.append("( ");
+		builder.append( bgp.getTriplePatterns() );
```
hartig, line 57: Does this work as expected? I mean, `getTriplePatterns` returns a `Set`. What does the `toString()` method of `Set` do?

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpRequest.java
```diff
@@ -44,4 +55,36 @@ public void visit( final LogicalPlanVisitor visitor ) {
...
+			else if ( pattern instanceof BGP) {
+				builder.append( ((BGP) pattern).getTriplePatterns() );
+			}
+			else {
+				LogicalPlanUtils.printTriplesOfGraphPattern( pattern.asJenaOp(), builder );
+			}
+		}
+		else {
+			throw new UnsupportedOperationException( "Print graph pattern of the Request: " + req.getClass().getName()+" is an open TODO." );
+		}
```
hartig, line 80: Perhaps we can simply rely on the `toString()` method of the classes that implement `DataRetrievalRequest`?

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpTPAdd.java
```diff
@@ -45,4 +48,18 @@ public void visit( final LogicalPlanVisitor visitor ) {
 		visitor.visit(this);
 	}
 
+	@Override
+	public String toString(){
+		final StringBuilder builder = new StringBuilder();
+
+		builder.append("> tpAdd ");
+		builder.append("( ");
+		builder.append( tp.asJenaTriple() );
```
hartig, line 57:
```suggestion
		builder.append( tp.asJenaTriple().toString() );
```
Can you please use this as a general rule and always call the `toString()` method explicitly.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoin.java
```diff
@@ -58,4 +59,17 @@ public void visit(final PhysicalPlanVisitor visitor) {
         visitor.visit(this);
     }
 
+    @Override
+    public String toString() {
+        final StringBuilder builder = new StringBuilder();
+
+        if ( lop instanceof LogicalOpTPAdd)  {
+            builder.append( "> bindJoin");
+            return ( (LogicalOpTPAdd) lop ).toString();
```
hartig, line 68: That does not work as expected. You are not using the builder in the return statement.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoin.java
```diff
@@ -58,4 +59,17 @@ public void visit(final PhysicalPlanVisitor visitor) {
...
+        }
+        else {
+            throw new IllegalArgumentException("Unsupported type of operator: " + lop.getClass().getName() );
+        }
```
hartig, line 72: Instead of having these if-else part with this exception, why don't you simply use `lop.toString()`, independent of the type of `lop`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoin.java
hartig: The same issue repeats in the other physical operators as well.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoin.java
hartig: The same issue repeats in the other physical operators as well.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/query/impl/SPARQLQueryImpl.java
```diff
@@ -39,4 +39,9 @@ public Query asJenaQuery() {
 		return jenaQuery;
 	}
 
+	@Override
+	public String toString(){
+		return this.asJenaQuery().getQueryPattern().toString();
```
hartig, line 44:
```suggestion
		return this.asJenaQuery().toString();
```
`getQueryPattern()` returns only the pattern of the WHERE clause.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanUtils.java
```diff
@@ -55,4 +54,15 @@ public void visit( final LogicalOpUnion op ) {
...
+
+		final DataRetrievalInterface intFace = fm.getInterface();
+		if ( intFace instanceof SPARQLEndpointInterface){
+			return "SPARQL endpoint at " + ((SPARQLEndpointInterface) intFace).getURL();
+		}
+		else {
+			return intFace.getClass().getName() + " server at TODO";
+		}
+	}
+
```
hartig, line 67: I don't think this is needed. Instead, also for the `FederationMember` objects we can rely on their `toString()` method.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/UnaryLogicalOp.java
```diff
@@ -4,5 +4,5 @@
 
 public interface UnaryLogicalOp extends LogicalOperator
 {
-	
+    String toString();
```
hartig, line 7: Shouldn't this method declaration already be in the `LogicalOperator` super-interface?

In fact, I wonder why you need to declare these methods at all. Every class in Java implicitly extends `java.Object` and, thus, has a `toString()`method anyways, and this method may be overridden in subclasses.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpBGPAdd.java
```diff
@@ -45,4 +46,14 @@ public void visit( final LogicalPlanVisitor visitor ) {
 		visitor.visit(this);
 	}
 
+	@Override
+	public String toString(){
+
+		return "> bgpAdd ( "
+				+ bgp.toString()
+				+ ", "
+				+ LogicalPlanUtils.printStringOfFm( fm )
```
hartig, line 55: As mentioned above, also for the `FederationMember` objects we extend each class that implements `FederationMember` with an overridden `toString()` method and, then, rely on that method here.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpRequest.java
```diff
@@ -44,4 +45,14 @@ public void visit( final LogicalPlanVisitor visitor ) {
 		visitor.visit(this);
 	}
 
+	@Override
+	public String toString(){
+
+		return " > req ( " +
+				"{ " + req.toString() + " }"
+				+ ", "
+				+ LogicalPlanUtils.printStringOfFm( fm )
```
hartig, line 54: same comment here

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpTPAdd.java
```diff
@@ -45,4 +46,14 @@ public void visit( final LogicalPlanVisitor visitor ) {
 		visitor.visit(this);
 	}
 
+	@Override
+	public String toString(){
+
+		return "> tpAdd ( "
+				+ tp.toString()
+				+ ", "
+				+ LogicalPlanUtils.printStringOfFm( fm )
```
hartig, line 55: same comment here

hartig (changes requested):

hartig (approved):

PR #139, opened 2021-12-18
Introduced LogicalPlanCounter to count the subplans in a logical plan.

SHSenpai:
A logical plan counter to count all the subplans within a logical plan.


SHSenpai:
Torn between whether to keep the LogicalCounterVisitor subclass as a child of LogicalWalkerVisitor to avoid code duplication or whether to make it separate in order to make the walk method protected.

I'm also don't quite get what the 'visit' function does. I see it declared but I don't know where it is implemented.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanUtils.java
```diff
@@ -20,6 +22,55 @@ static public boolean isSourceAssignment( final LogicalPlan plan ) {
...
 		return v.wasSourceAssignment();
 	}
+	
+	static public int countSubplans ( final LogicalPlan plan ) {
+		final LogicalPlanCounter c = new LogicalPlanCounter();
+		LogicalPlanWalker.walk(plan, null, c);
+		return c.getSubplanCount();
+	}
+	
+	static public class LogicalPlanCounter extends LogicalPlanVisitorBase {
```
hartig, line 32:
```suggestion
	static public class LogicalPlanCounter implements LogicalPlanVisitor {
```
Better this way to make sure you don't miss overriding any of the `visit` methods.

hartig (approved):
Looks good. I have just added one change suggestion which I will commit right away, and then I will merge this PR.

PR #141, opened 2021-12-18
Add cost of selected plan to statistics

chengsijin0817:


hartig (approved):
Okay (with the idea that I will add an "experiment-runs" flag to avoid calling the cost model just for the purpose of getting the cost for the stats, as is what currently happens in the case of SimpleJoinOrderingQueryOptimizer).

hartig:
I have implemented the aforementioned idea of an "experiment-runs" flag. See the following commit: https://github.com/LiUSemWeb/HeFQUIN/commit/b76aed8e1be894ad74c1c27280d5401aed4186da

PR #142, opened 2021-12-21
Filter empty lists

scferrada:
Fixes the case where a query would return empty lists if no property had a given value.


hartig (approved):
Looks good. Thanks!

PR #143, opened 2021-12-24
Two Phase Optimization

SHSenpai:
Issue: 2PO has to run II first and then SA, but SA requires a logical plan for the size estimation and not a physical plan - which is the only thing that II returns.
Do you think it may be the best to make a walker and plan counter for a physical plan after all? @hartig


hartig:
No, I don't think we need a physical subplan counter for this. Instead, you can simply count the subplans of the logical plan, then convert the logical plan to an initial physical plan, pass that to II, let it run for a fixed number of iterations (e.g., 10 as in the paper), and, finally, pass the resulting physical plan as initial state to SA, together with the subplan count. All of these steps can be initiated from within the `optimize(LogicalPlan)` function of the class you implement for 2PO.

SHSenpai:
Oh, so the number of subplans won't change from the start of II to the start of SA? Then yes, this implementation will be quite straightforward.

SHSenpai:
How does the QueryOptimizationStats work by the way? Each algorithm now yields one. My idea would be to want to 'combine' the stats from II and SA to have a single stats object in 2PO so that it can fit the return type.

hartig:
> How does the QueryOptimizationStats work by the way? Each algorithm now yields one. My idea would be to want to 'combine' the stats from II and SA to have a single stats object in 2PO so that it can fit the return type.

Don't worry about it at the moment. Just create a `QueryOptimizationStatsImpl` at the end of 2PO in the same way as the other optimizers do.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/ConditionsForTwoPhaseOptimization.java
```diff
@@ -0,0 +1,18 @@
+package se.liu.ida.hefquin.engine.queryproc.impl.optimizer.randomized;
+
+public class ConditionsForTwoPhaseOptimization {
```
hartig, line 3: I don't think you need an explicit class for that. Instead, simply make the two conditions two separate arguments of the constructor of `TwoPhaseQueryOptimizer`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/TwoPhaseQueryOptimizer.java
```diff
@@ -0,0 +1,34 @@
...
+import se.liu.ida.hefquin.engine.queryplan.logical.LogicalPlanUtils;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationException;
+import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationStats;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationContext;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RuleInstances;
+import se.liu.ida.hefquin.engine.utils.Pair;
+
+public class TwoPhaseQueryOptimizer extends RandomizedQueryOptimizerBase {
+	
+	protected final ConditionsForTwoPhaseOptimization conditions;
```
hartig, line 14: Not needed.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/TwoPhaseQueryOptimizer.java
```diff
@@ -0,0 +1,34 @@
...
+	                           final RuleInstances rewritingRules) {
+		super(context, rewritingRules);
+		assert x != null;
+		conditions = x;
+		optimizer1 = new IterativeImprovementBasedQueryOptimizer(conditions.getStopping(), context, rewritingRules);
+		optimizer2 = new SimulatedAnnealing(conditions.getEquilibrium(), context, rewritingRules);
+	}
+
+	@Override
+	public Pair<PhysicalPlan, QueryOptimizationStats> optimize(LogicalPlan initialPlan)
```
hartig, line 29:
```suggestion
	public Pair<PhysicalPlan, QueryOptimizationStats> optimize( final LogicalPlan initialPlan )
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/TwoPhaseQueryOptimizer.java
```diff
@@ -0,0 +1,34 @@
...
+			throws QueryOptimizationException {
+		return optimizer2.optimize(optimizer1.optimize(initialPlan).object1,LogicalPlanUtils.countSubplans(initialPlan)); // Should work, but currently only gets the stats from the SA stage.
```
hartig, line 31: For the sake of making the code more readable, can you please avoid nesting these function calls. Instead, assign the intermediate return objects to some local variables and use these variables for the next call.

hartig: Another issue: Notice that, according to the paper, the initial temperature of the SA phase within 2PO is different from the initial temperature when using SA just by itself. To capture this aspect correctly, I suggest that you extend the `SimulatedAnnealing` class with another `public optimize` function that takes the initial temperature as an additional parameter.

hartig (changes requested):
There are a few things that need to be changed--see my review comments.

SHSenpai:
All requested changes have been carried out.

Question: does Java support method headers that take in parameters in an arbitrary order? I was thinking a bit about which ordering of the parameters made the most sense, but since all of the datatypes are different it would be user-friendly to make the function's parameters order-independent if Java allows for such a thing.

hartig (approved):

hartig:
> Question: does Java support method headers that take in parameters in an arbitrary order?

No.

hartig:
This PR looks good now. I have merged it.

What remains to complete the work on implementing the algorithms from the paper is to use the randomized join plan optimizer in `IterativeImprovementBasedQueryOptimizer`. To this end, remember the things I mentioned in my earlier comment: https://github.com/LiUSemWeb/HeFQUIN/pull/135#issuecomment-992926997

Issue #60, opened 2021-06-30
Add possibility to request cardinality estimates for TPF and brTPF


hartig:
This issue is the TPF and brTPF specific version of #58. That is, the implementation for #58 has extended `FederationAccessManager` with a `performCardinalityRequest` method for SPARQL endpoints, and the purpose of this issue here is to do the same for TPF servers and brTPF servers. An implementation idea is to simply do the given TPF/brTPF request and, then, extract the cardinality estimate from the metadata of the response to this request.

hartig:
Half of this issue is addressed by commit https://github.com/LiUSemWeb/HeFQUIN/commit/1b6b4c80bdf9ce0bba662e5b0711a4f98efc8409 -- namely, the extension of FederationAccessManager with method to issue cardinality requests to TPF/brTPF servers. What remains is to implement the extraction of the cardinality estimate from the metadata of TPF responses.

hartig:
After completing the implementation for this issue (i.e., after implementing the extraction of the cardinality estimate from the metadata of TPF/brTPF responses), for this feature to actually work, we also need to address #17 and #18.

Issue #18, opened 2021-05-03
implementation of BRTPFRequestProcessor needed


hartig:
This may be an extension of the implementation of TPFRequestProcessor. Hence, #17 should be addressed first.

PR #144, opened 2022-01-01
Implementation of TPFRequestProcessor

### LINKED ISSUE #17: implementation of TPFRequestProcessor needed
This will be a bit of work. For inspiration, one may take a look at how Jena's 'RDFConnectionRemote` works internally.

hartig:
closes #17


hartig:
closes #18 as well, and thus also closes #60

hartig:
@chengsijin0817 Here comes the code to perform TPF requests and brTPF requests, including some unit tests which show that it works :-)     ...the only thing that is missing is correct support for paging; in fact, when implementing this PR, I noticed that the way I was assuming how paging would be supported (namely, based on page numbers) does not work. I have created another issue about it and will work on that one next.

PR #153, opened 2022-01-18
Using Randomized for IterativeImprovement

SHSenpai:
I'm trying out using a SimpleJoinOrderingQueryOptimizer with a Randomized as its join-plan-optimizer as per your instructions, but I'm getting an error in the test due to some operation returning null. Since I don't quite understand how the joins work on a low-level I'm having trouble scratching my brain as to what is causing this error.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/randomized/IterativeImprovementBasedQueryOptimizer.java
@@ -10,6 +10,8 @@ import se.liu.ida.hefquin.engine.queryproc.QueryOptimizationStats;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationContext;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.QueryOptimizationStatsImpl;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.rewriting.RuleInstances;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.simple.RandomizedJoinPlanOptimizerImpl;
+import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.simple.SimpleJoinOrderingQueryOptimizer;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.PhysicalPlanWithCost;
 import se.liu.ida.hefquin.engine.queryproc.impl.optimizer.utils.PhysicalPlanWithCostUtils;
 import se.liu.ida.hefquin.engine.utils.Pair;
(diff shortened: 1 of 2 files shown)


hartig:
The cause of this issue is that `DummyPlanForTest` is not an actual physical plan. In particular, it does not have a root operator but, instead, its `getRootOperator()` function simply returns `null` and that's what is picked up in the `hasMultiwayJoinAsRoot` function of `SimpleJoinOrderingQueryOptimizer`.

Another (separate) issue that I observe is that the `optimizePlan(PhysicalPlan)` function of `SimpleJoinOrderingQueryOptimizer` expects a `PhysicalPlan` in which all joins are captured using multi-way join operators (rather than binary ones). Therefore, when your `optimize(LogicalPlan)` function of your `IterativeImprovementBasedQueryOptimizer` calls the `convert` function of the LogicalToPhysicalPlanConverter, you need to pass `true` as the second argument. Note that I had already mentioned this earlier in the following comment that I have been pointing to at the ends of your various PRs: https://github.com/LiUSemWeb/HeFQUIN/pull/135#issuecomment-992926997

SHSenpai:
Sorry! I didn't see your comment in PR 135, it must've gone unnoticed by me due to the branch deletion after the merge. My bad!

@hartig Is there any documentation or reference which I could read up to get a better understanding of how the plans work? I assume that the root operator of any plan isn't supposed to be null, so if I learn more about the workings of HeFQUIN and its plans I would probably be able to make my own tests and have them be 'realistic' when it comes to the plans. It would save you time as well since it would make me more independent.

hartig:
There was no such documentation or reference, other than the source code, of course (which actually contains quite a bit of JavaDoc comments in most of the relevant interfaces). So, I took some hours today to start writing up something that should be helpful for you, and for future students/collaborators as well:

https://github.com/LiUSemWeb/HeFQUIN/wiki/Query-Processing-in-HeFQUIN

Let me know if anything is unclear in this description.

hartig:
No, this understanding is incorrect. Subplans that do not have children (i.e., the leaf nodes of the tree) still have an operator. These operators are nullary operators (not "null operators"!). A nullary function/operator is an operator that can produce output without consuming any input obtained from some other operator; think of a function that has zero arguments. The only type of nullary operator that we have in the system is the request operator.

You find links to all my publications at http://olafhartig.de/publications.html
Two publications that are particularly relevant for HeFQUIN (besides the FedQPL paper that you have looked at) are:

* Sijin Cheng and Olaf Hartig: OPT+: A Monotonic Alternative to OPTIONAL in SPARQL. In Journal of Web Engineering 18(1-3), 2019. http://olafhartig.de/files/ChengHartig_OptPlus_2019.pdf
* Olaf Hartig and Carlos Buil-Aranda: Bindings-Restricted Triple Pattern Fragments. Proceedings of the 15th International Conference on Ontologies, Databases, and Applications of Semantics (ODBASE), 2016. http://arxiv.org/pdf/1608.08148

SHSenpai:
Thanks. It seems like I was overthinking things and I ended up making it harder for myself than it needed to be. Sorry that it took so long for me to figure it out!

I believe everything should be ready now. The IterativeImprovement algorithm now uses Randomized and the test is updated so that it runs properly, where I have created a dummy operator class and changed your dummy plan class to use it so that the operator isn't null and can be cast.

What do you have for me going forward? It might be that something in these algorithms still need polishing, but I can do that while being blocked. My workflow is more like a few long bursts of constant work which lasts for a whole prenoon/afternoon/day rather than short iterative things.

hartig (approved):

hartig:
Thanks for completing this PR. It looked good now and I have merged it.

> What do you have for me going forward?

I didn't have a chance to think about it. I need to do that but I am hopelessly overbooked at the moment, and next week I want (and need!) to take vacation. Nonetheless, I will try to come up with something beforehand.

PR #154, opened 2022-01-23
Fix an issue with "costOfSelectedPlan'

chengsijin0817:
Fix an issue with "costOfSelectedPlan': 
1. The updated Stats (including costOfSelectedPlan) should be returned by the method 'processQuery'; 
2. As 'costOfSelectedPlan' has been collected in QueryProcStats, remove 'costOfSelectedPlan' from the evolutionary algorithm's 'QueryOptimizationStats' to maintain consistency between different algorithms.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/QueryProcessorImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/QueryProcessorImpl.java
@@ -66,7 +66,7 @@ public class QueryProcessorImpl implements QueryProcessor
 
 		final long t4 = System.currentTimeMillis();
 
-		final StatsImpl myStats = new QueryProcStatsImpl( t4-t1, t2-t1, t3-t2, t4-t3, qepAndStats.object2, execStats );
+		final QueryProcStatsImpl myStats = new QueryProcStatsImpl( t4-t1, t2-t1, t3-t2, t4-t3, qepAndStats.object2, execStats );
 
 		// stats that may be expensive to collect should be collected only when running experiments
 		if ( ctxt.isExperimentRun() ) {
@@ -75,7 +75,7 @@ public class QueryProcessorImpl implements QueryProcessor
 			myStats.put("costOfSelectedPlan", costOfSelectedPlan);
 		}
 
-		return new QueryProcStatsImpl( t4-t1, t2-t1, t3-t2, t4-t3, qepAndStats.object2, execStats );
+		return myStats;
 	}
 
 }


hartig:
Honestly, I don't like this idea to go in from the "outside" and change the stats of a subcomponent in retrospect.

What is the problem with having the cost of the executed plan as a statistics entry on the level of the query processor?

chengsijin0817:
No problem, it is another solution.
 If having "costOfSelectedPlan" as a statistical entry on the level of the query processor, then a function "getCostOfSelectedPlan" and the corresponding implementation needs to be added to the relevant classes for retrieving "costOfSelectedPlan":
QueryProcStats, QueryProcStatsImpl

hartig:
I would not explicitly add such a getCostOfSelectedPlan function because this stats entry is added only in case of experiment runs. Note that you do not need such a function to access the entry. Instead, you can also access it via the method of the Stats interface.

chengsijin0817:
> 

It makes sense. I have accessed the cost value via the Stats interface. ;-)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/EvolutionaryAlgorithmQueryOptimizer.java
```diff
@@ -59,7 +59,6 @@ public EvolutionaryAlgorithmQueryOptimizer( final QueryOptimizationContext ctxt,
 
         final QueryOptimizationStatsImpl myStats = new QueryOptimizationStatsImpl();
         myStats.put( "numberOfGenerations", previousGenerations.size() + 1 );
-        myStats.put( "costOfSelectedPlan", bestPlan.getWeight() );
```
hartig, line 62: I think you don't need to delete this. It doesn't hurt to have this stats entry here too, and here it is always added, not only during experiment runs.

hartig (changes requested):
Just two small things.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/QueryProcessorImpl.java
```diff
@@ -75,7 +75,7 @@ public QueryProcStats processQuery( final Query query, final QueryResultSink res
 			myStats.put("costOfSelectedPlan", costOfSelectedPlan);
 		}
 
-		return new QueryProcStatsImpl( t4-t1, t2-t1, t3-t2, t4-t3, qepAndStats.object2, execStats );
+		return (QueryProcStats) myStats;
```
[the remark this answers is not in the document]
chengsijin0817: Thank you. I have updated the type of variable myStats.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/EvolutionaryAlgorithmQueryOptimizer.java
chengsijin0817: You're right, it is not only added during experiment runs. Sorry I didn't think of this.

PR #158, opened 2022-02-02
Consider tpAdd and bgpAdd …

chengsijin0817:
…when converting the logical plan with multiwayJoins to a physical plan.


hartig (approved):
Look good. Thanks!

PR #160, opened 2022-02-07
SPARQL's FILTER

SHSenpai:
Stuck on trying to understand Nodevalue and Expr. More in a comment.

I'm trying to understand Nodevalue and the Eval function through reading the reference material here:
https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/sparql/expr/NodeValue.html#eval(org.apache.jena.sparql.engine.binding.Binding,org.apache.jena.sparql.function.FunctionEnv)
https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/sparql/expr/ExprNode.html
https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/sparql/expr/Expr.html#eval(org.apache.jena.sparql.engine.binding.Binding,org.apache.jena.sparql.function.FunctionEnv)

as well as looking at the files Expr.class, ExprUtils.class, and NodeValue.class.

I can't figure out whether it guarantees a return of NodeValue TRUE or NodeValue False from looking at the reference material alone. Expr.class is just an interface without any implemented code for the eval-function, and all of the references just says that it will "Evaluate this expression against the binding" which I do not fully understand the implications of.

Do you have any material that I can read in order to gain a better understanding of how this works?
@hartig


scferrada:
Hi Simon,

Maybe in the meantime you can see this package of Jena, that contains all the implementations of `Expr`. An expression can be, for instance an addition `?x + 1`, a function call `STARTSWITH(?s, 'Prefix')` or a conjunction of logical expresions `?x AND ?y`.

https://github.com/apache/jena/tree/main/jena-arq/src/main/java/org/apache/jena/sparql/expr

There are, of course, no guarantees as to what would a generic expression return when evaluated over a set of solution mappings, but for instance `E_IsIRI` should evaluate to true, false or unbound.

Cheers.

SHSenpai:
Thanks, will do!

SHSenpai:
Thanks both of you for your explanations. I have a few questions for going forward with writing the tests.

First of all, what does the `?`-sign mean in the context of the expressions `?x + 1` and `STARTSWITH(?s, 'Prefix')`? My first instinct was that they were characters, as in ?x + 1 = ?y, and that STARTSWITH(?s, 'Prefix') would return FALSE, but from the way the SPARQL documentation reads it seems like they could also be wildcard variables. A search on the Internet yielded first that they were ternary operators, and second that they were wildcards as one of my guesses were. I would assume that `?x` can be any integer, and that `?s` can be any string, but those may also just be names.

Second, I'd like to be better able to find things when using Eclipse. I've started making use of the 'Navigate' tab much more, but how do I know if there's more than one implementation for a given interface? For example, Navigate -> Open Implementation for `IntermediateResultBlock` gives me the class `GenericIntermediateResultBlockImpl`. How do I know whether there are more implementations, and if there are - how do I access them?

scferrada:
Yes, in the examples `?x` and `?s` are variables, which are bound to a set of values when the query is evaluated. Then, to evaluate an expression, the bound values are used.

For instance, if we have the following solution mappings:
1.- `(?x -> 1, ?s -> "Mr. Smith")`
2.- `(?x -> 2, ?s -> "Mrs. Smith")`

Then the expression `?x + 1` will evaluate to 2 using sol. map. 1, and to 3 using sol. map. 2. The expression `STARTSWITH(?s, 'Mr.')` evaluates to `true` with sol. map. 1, and to `false` with sol. map. 2.

hartig:
In addition to Sebastian's response about he variables, I refer you again to the following short video that should give you a basic idea of SPARQL:
https://www.youtube.com/watch?v=FvGndkpa4K0

To your other question ("how do I know if there's more than one implementation for a given interface?"), in Eclipse you can right-click on an interface name (both within the source code and in the project explorer on the left hand side) and then select "Open Type Hierarchy" in the popup menu.

SHSenpai:
Thanks to both of you for your answers. The video was helpful, I can't recall having had it elsewhere though but I've added it to my bookmarks folder of reference material for this work.

I've made some good progress this week in creating a test for seeing that the filter works, and I've found implementations for everything except for IntermediateResultElementSink, so I made a dummy to use in the test.
The current thing I'm looking into is how Bindings work. SolutionMappingForTests looks to be really convenient but it lacks support for asJenaBinding(), which is something which the filter makes use of.
If there's no more obstacles which turn up (such as something else breaking or the filter function not working as its supposed to) I'll hopefully be done with this pretty soon.

I have a question about unit testing in Java. I want to make sure that the 'iterable' I get out of the result block contains a certain variable. I've searched online for how to do this, and all results refer to something called 'hamcrest' which I cannot seem to import properly. What is the right way to make such an assertion?

hartig:
> The current thing I'm looking into is how Bindings work. SolutionMappingForTests looks to be really convenient but it lacks support for asJenaBinding(), which is something which the filter makes use of.

We have a helper class called `SolutionMappingUtils`. That one has functions for creating `SolutionMapping` objects. One of many places where they are used (just to give you a pointer to example usage) is `TestsForSolutionMappingsIterableWithFilter` which is a base class for some unit tests.

> I have a question about unit testing in Java. I want to make sure that the 'iterable' I get out of the result block contains a certain variable. I've searched online for how to do this, and all results refer to something called 'hamcrest' which I cannot seem to import properly. What is the right way to make such an assertion?

It is not that the `Iterable` as a whole contains variables. Instead, you have to check separately for every `SolutionMapping` of the `Iterable`.

For some of the more complex types of checks, you find some helper functions in `EngineTestBase` (see the functions called `assertHasNext`). The latter ones, together with the aforementioned `TestsForSolutionMappingsIterableWithFilter`, is used for instance in `SolutionMappingsIterableWithOneVarFilterTest`.

Let me know if you have further questions about this.

SHSenpai:
I've been trying to debug my test when I use it with the solution creator in SolutionMappingUtils. It seems to get further than the last one, but the debug isn't giving me much. (image 1)
image
I send this trace to console, giving me this: (image 2)
image
I then set a breakpoint for this exception... (image 3)
image
...but the breakpoint doesn't trigger during subsequent runs and it gives me the exact same as in images 1 and 2.

During a previous run the breakpoint worked, and it showed me something helpful where I saw that the eval function checked that the binding's "nv" was not null - but it was null. After editing the test it ceased stopping at the breakpoint though, even when I reverted the test to being the way it was when the breakpoint triggered correctly. I assume nv stands for "node variable" or "node value".

---

I also have a question regarding the solution mappings in SolutionMappingUtils, When doing like it was done in TestForSolutionMappingsIterableWithFilter, a variable is 'allocated' with a certain name, and then said to be put in a certain URI, as such:
```
final Var var1 = Var.alloc("v1");
final Var var2 = Var.alloc("v2");
final Node uri1 = NodeFactory.createURI("http://example.org/uri1");
final Node uri2 = NodeFactory.createURI("http://example.org/uri2");
final SolutionMapping sol1 = SolutionMappingUtils.createSolutionMapping(var1, uri1, var2, uri2);
```
I'm having a hard time understanding how the variable is actually set. If I understand it correctly, a variable named v1 is now bound to a place called uri1, but v1 doesn't have a value, which is what makes the exception `org.apache.jena.sparql.expr.VariableNotBoundException: Not bound: variable ?x`
occur.

---

As for
> It is not that the Iterable as a whole contains variables. Instead, you have to check separately for every SolutionMapping of the Iterable.

I currently do it like this:
`for(SolutionMapping solution : input.getSolutionMappings()) {
	//Check whether it satisfies the filter expression
	final NodeValue evaluationResult = ExprUtils.eval(filterExpression, solution.asJenaBinding());
	if(evaluationResult == NodeValue.TRUE) { // Temporary. Need to look more into the nodevalue and expr.
	//push it to the output sink.
		sink.send(solution);
	} else if (evaluationResult != NodeValue.FALSE ) {
		throw new ExecOpExecutionException("The result of the eval is neither TRUE nor FALSE!", null);
	}
}`
It is the eval method which gives me all of the errors, as far as I am able to see.

In this situation, I believe that one or more of these are happening:
1. I'm incorrectly writing the filter expression.
2. The variables that I'm putting in the solution has no value and thus cannot be compared.
3. The variables that I'm putting in the solution has the wrong datatype and cannot be compared as the filter's expression says.

I want to create a variable, I want that variable to be an integer/number with a value of 8. I want to make sure that this value passes the filter `x? < 10` and that a variable with a value of 12 is sent to the sink and doesn't show in the resultblock after the filter processes it.

hartig:
Regarding the `VariableNotBoundException`, I don't know what you are trying to do with the "Java Exception Breakpoint". I have never used this feature. Independent of that, it seems a bit strange to me that the stack/failure trace for this exception does not show where the exception was thrown. I tried on my computer and it is the same. However, let's ignore that and focus on the essentials: the exception happens somewhere inside the call of `ExprUtils.eval` (currently line 36 in `ExecOpFilter`) and it indicates that the given expression cannot correctly be evaluated based on the given solution mapping (which is obvious because the expression refers to variable ?x but in the given solution mapping only the variables ?v1 and ?v2 are bound; hence, the error message of the exception). So, this exception is probably Jena's way of saying that the evaluation of the expression resulted in an error. Consequently, the current solution mapping needs to be filtered out in `ExecOpFilter`. To do this, simply put a try-catch block around the while code inside the for-loop inside the `process` function.

> I also have a question regarding the solution mappings in SolutionMappingUtils, [...] I'm having a hard time understanding how the variable is actually set. If I understand it correctly, a variable named v1 is now bound to a place called uri1, but v1 doesn't have a value, which is what makes the exception org.apache.jena.sparql.expr.VariableNotBoundException: Not bound: variable ?x occur.

To make the discussion less ambiguous let's distinguish query variables (i.e., variables that can be used in SPARQL queries) and program variables (i.e., variables that can be used in Java code). In the example, ?x and ?v1 are query variables, and `var1`, `uri1`, etc. are program variables.

Now, it is not correct to say that "a variable named v1 is now bound to a place called uri1." Instead, what happens in this piece of code is that a Java object representing a solution mapping is created such that, in this solution mapping, the query variable ?v1 is bound to a value, where this value happens to be the URI http://example.org/x1 (within the Java code, this URI is represented through a Java object of type `Node`). Additionally, the query variable ?v2 is bound to the URI http://example.org/x2 in the solution mapping. You can also say that the solution mapping has values for two (query) variables, namely for the variables ?v1 and ?v2.

Perhaps the concept of a solution mapping is also not so clear to you. Informally, you can think of solution mappings to be the rows in the result of a SPARQL query (or subquery).

> It is the eval method which gives me all of the errors, as far as I am able to see.

Yes.

> I want to create a variable, I want that variable to be an integer/number with a value of 8.

Assuming by "variable" you mean a query variable, such as ?x. In this case, your thinking in the latter sentence is a bit wrong. It is not that a (query) variable can have a value in general. Instead, there needs to be solution mapping in which a query variable is assigned to a value. You can create such a solution mapping (or, more precisely, a Java object that represent it) by doing:
```java
final Node value = ... // create a Node object that represents the desired value
final Var var = Var.alloc("x");
final SolutionMapping sm = SolutionMappingUtils.createSolutionMapping(var, value);
```
Now, the only question that remains about this piece of code is. What kind of `Node` object do I need to assign to the program variable `value` in order to represent the a value of 8?

In the context of RDF such values are called literals, which consist of two things: a string representing the value in a lexicographic form (e.g., the string "8") and a datatype that is captured by means of a URI. Assuming that your value of 8 is meant to be an integer, the URI for this datatype is http://www.w3.org/2001/XMLSchema#integer (sometimes you may also see xsd:integer whenever it is clear what the prefix xsd refers to).

Now, to create a `Node`object that represent the literal with this datatype and the value 8, you can do:
```java
final Node value = NodeFactory.createLiteral("8", XSDDatatype.XSDinteger);
```

> I want to make sure that this value passes the filter x? < 10 

I hope that this part should be clear now.

> [...] and that a variable with a value of 12 is sent to the sink and doesn't show in the resultblock after the filter processes it.

If a solution mapping bind the (query) variable ?x to the value 12, then this solution should **_not_**  be sent to the sink because it does not pass the filter ?x < 10

> GitHub doesn't show my code correctly indented with spaces and line breaks, so I changed it for a picture.

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks

SHSenpai:
Thanks. This has cleared up a lot.

Question: Does the next()-method of Iterator only return the next value, or does it return the next value and also advance the iterator by one step?

I've implemented a non-error-causing test.
x = 8.
The filter expression is "x? < 10".
The evaluation of x against the filter expression is true.
Therefore, x is sent to the sink.
Should x then show up in the result block after the filtering?

x = 12.
The filter expression is "x? < 10".
the evaluation of x against the filter expression is falsse.
Therefore, x is not sent to the sink.
Should x then show up in the result block after the filtering?

Question: Should the filter at any time remove the x = 8 from the result block and retain x = 12 in it, or vice versa, or not?
Question: Are the variables in the sink ever used for anything, or is it just a rubbish heap where they go never to be seen again?

hartig:
> Question: Does the next()-method of Iterator only return the next value, or does it return the next value and also advance the iterator by one step?

Yes, it also advances the iterator to the next element (if any).

> x = 8. [...] x is sent to the sink. [...] Should x then show up in the result block after the filtering?

I don't think that you are sending x to the sink but the you are sending the solution mapping to the sink.

In this sense, it is also not x that shows up in the result block (at least not directly) but it is the solution mapping that contains x.

> x = 12. [...] the evaluation of x against the filter expression is falsse. Therefore, x is not sent to the sink. Should x then show up in the result block after the filtering?

If the solution mapping is not sent to the sink (which is the correct behavior in this case), then it should also not show up in the result block.

> Question: Should the filter at any time remove the x = 8 from the result block and retain x = 12 in it, or vice versa, or not?

I don't understand the question.
Which result block are you referring to--the one that is given as input to the `process`function?
What do you mean by "remove"? ...and what do you mean by "at any time"?

> Question: Are the variables in the sink ever used for anything, or is it just a rubbish heap where they go never to be seen again?

Again, you are not putting variables into the sink but solution mappings.

The execution engine uses the sinks to pass solution mappings from one operator to the next operator in the query execution plan. So, every operator writes/sends the solution mapping that it produces as output to its sink. That sink is connected as input to the next operator which can then consume the solution mappings from that sink and use them to produce its own output solution mappings from them, and these output solution mappings are then sent to the output sink of that next operator. So, the `IntermediateResultBlock` that is `input` to the `process` function of the `ExecOpFilter` is essentially set up as a wrapper around the sink that the previous operator in the plan sends its output solution mappings to.

In the unit tests there is only one operator. So, here you can look into the sink to which this operator writes and check that your implementation of that operator does indeed send the correct output solution mappings (and does not send solution mappings that are not correct output for that operator).

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpFilterTest.java
```diff
@@ -1,17 +1,78 @@
...
+import se.liu.ida.hefquin.engine.queryplan.executable.IntermediateResultElementSink;
+import se.liu.ida.hefquin.engine.queryplan.executable.impl.GenericIntermediateResultBlockImpl;
+import se.liu.ida.hefquin.engine.queryplan.executable.impl.pullbased.TestUtils;
+
 public class ExecOpFilterTest
 {
 	@Test
-	public void test1_RenameTheseTestMethodsToSomethingThatDescribesTheirPurpose() {
+	public void filter_Numbers() {
+		final FilterTestSink sink = new FilterTestSink();
```
hartig, line 28: Instead of creating your own `FilterTestSink`, simply use a `MaterializingIntermediateResultElementSink` here. That one materializes all the solution mappings that are sent to it into an internal data structure to which you can get access via the `getMaterializedIntermediateResult()` function of this class.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpFilterTest.java
```diff
@@ -1,17 +1,78 @@
...
+		final GenericIntermediateResultBlockImpl resultBlock = new GenericIntermediateResultBlockImpl();
+		final Expr lessThan10 = ExprUtils.parse("?x < 10");
+		
+		final Node value8 = NodeFactory.createLiteral("8", XSDDatatype.XSDinteger);
+		final Node value12 = NodeFactory.createLiteral("8", XSDDatatype.XSDinteger);
+		final Var x8 = Var.alloc("x");
+		final Var x12 = Var.alloc("x");
```
hartig, line 35: There is no need to create two java objects that represent the same query variable. Simply do:
```java
final Var x = Var.alloc("x");
```
... and then use that program variable `x` wherever you are currently using (or planning to use) `x8` and `x12`.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpFilterTest.java
```diff
@@ -1,17 +1,78 @@
...
+		resultBlock.add(sol12);
 		
+		final ExecOpFilter filterLessThan10 = new ExecOpFilter(lessThan10);
+		try {
+			filterLessThan10.process(resultBlock, sink, TestUtils.createExecContextForTests());
+		} catch (ExecOpExecutionException e) {
+			e.printStackTrace();
+		}
+
+		final Iterator<SolutionMapping> it = resultBlock.getSolutionMappings().iterator();
```
hartig, line 50: That's incorrect. You are asking for an iterator over the result block that you use as input to the test. Instead, you should ask for an iterator over the output, which will have been written to the sink. So, if you use a `MaterializingIntermediateResultElementSink` as I suggest above, then here you now need to do: 
```java
final Iterator<SolutionMapping> it = sink.getMaterializedIntermediateResult().iterator();
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpFilter.java
```diff
@@ -27,7 +29,24 @@ public int preferredInputBlockSize() {
 	public void process( final IntermediateResultBlock input,
 	                     final IntermediateResultElementSink sink,
 	                     final ExecutionContext execCxt ) throws ExecOpExecutionException {
-		// TODO Auto-generated method stub
+		
+		
+		// For every solution mapping in the input...
+		for(SolutionMapping solution : input.getSolutionMappings()) {
```
hartig, line 35:
```suggestion
		for( final SolutionMapping solution : input.getSolutionMappings() ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpFilter.java
```diff
@@ -27,7 +29,24 @@ public int preferredInputBlockSize() {
...
+			//Check whether it satisfies the filter expression
+			
+			try {
+				final NodeValue evaluationResult = ExprUtils.eval(filterExpression, solution.asJenaBinding());
+				if(evaluationResult == NodeValue.TRUE) {
+					sink.send(solution);
+				} else if (evaluationResult != NodeValue.FALSE ) {
+					throw new ExecOpExecutionException("The result of the eval is neither TRUE nor FALSE!", null);
+				}
+			} catch (VariableNotBoundException e) {
```
hartig, line 45:
```suggestion
			} catch ( final VariableNotBoundException e ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpFilter.java
```diff
@@ -27,7 +29,24 @@ public int preferredInputBlockSize() {
...
+				throw new ExecOpExecutionException("A variable name in the binding does not match the variable names in the filter.", null);
```
hartig, line 46: You should not throw an exception here. The `VariableNotBoundException` that triggers this catch block indicates that evaluating the filter expression based on the current solution mapping results in an error, which means that the current solution mapping does not satisfy the filter condition and, thus, is not in the output of this operator (i.e., should not be sent to the sink). Hence, instead of throwing an exception here, simply do not do anything in this catch block, which will then cause the program flow to move on to the next solution mapping from the input.

Having said that, instead of leaving this catch block completely empty, it may be helpful to include a comment about why the catch block does not do anything. So, here is my change suggestion for this source code line:
```suggestion
				// The current solution mapping does not satisfy the filter condition (because
				// evaluating the filter expression based on this solution mapping resulted in
				// this error). Therefore this solution mapping is not in the output of this operator
				// and, thus, must not be sent to the sink. Hence, we do not have to do anything
				// here in this catch block.
```

hartig (changes requested):

SHSenpai:
Thanks for the clarifications. Previous programming tasks I have done have always had 'sink' to mean 'garbage' similar to /dev/null in Linux, so I got a bit confused about its meaning in this project.

I'm currently working on getting testing done for this, but I'm running into an issue when it comes to creating date variables (see lines 65, 93, and 110 in ExecOpFilterTest).

scferrada:
> I'm currently working on getting testing done for this, but I'm running into an issue when it comes to creating date variables (see lines 65, 93, and 110 in ExecOpFilterTest).
 In lines 93 and onwards, dates should be surrounded by quotes, otherwise you're subtracting integers:
```
assertHasNext( it, "2020-10-20", x);
assertHasNext( it, "2021-02-01", x);
assertHasNext( it, "2020-01-01", x);
```

SHSenpai:
image
image
I've tried doing that, but it seems to make no difference (except that I have to make a string function - something seems to be wrong with my XSDDateType function). I get these errors, and there's not much of a trace to them. I have no idea where it's getting 1976 from.

scferrada:
I think the error is quite clear. The Iterator contains a `Node` object with a literal value "2020-10-20"^^<http://www.w3.org/2001/XMLSchema#date>, and you are trying to compare that with the result of the subtraction 2020-10-20. The method `getLiteralValue` that you use in `assertHasNext` returns an object, not an int.

hartig:
Instead of `getLiteralValue`, use `getLiteralLexicalForm`, which gives you a `String` representation of the value of the literal. And, then, as Sebastian says, you need to compare that string to the expected string (string and not integer).

hartig:
@SHSenpai Can you please wrap this up so that we can move on.

If you don't manage to fix the unit test with the date filter, then drop it. It doesn't actually add anything over the other unit test because it tests the `process` method of `ExecOpFilter` in exactly the same way. I mean, we don't need to test here whether the underlying Jena code works for all sorts of filter expressions; Jena has its own unit tests. So, one more unit test that we would actually need (that tests another code path within the `process` method of `ExecOpFilter`) is to have an input solution mapping in which the variable mention in the filter condition is not bound. It should be quick to add that by simply making a copy of `filter_Numbers()` and then adjust this copy accordingly.

hartig:
Good catch about the date in the filter expression. I have fixed this issue for you. The thing is that the date in the filter expression must also be represented as a date-typed literal. An example for how this looks like in an actual SPARQL query can be found at https://codyburleson.com/blog/sparql-examples-date-range 

This PR is ready to be merged now. However, I see that there is a merge conflict because the main branch has advanced in the meantime. I will see how I can resolve this conflict...

hartig:
... merge conflict resolved. I will merge this PR now...

hartig:
...merged!

Next tasks for you @SHSenpai : Add a corresponding logical operator and a corresponding physical operator. Begin with the logical one (because for implementing the physical one, you will need the logical one).

The class for the logical operator should be called `LogicalOpFilter`, it should go into the Java package `se.liu.ida.hefquin.engine.queryplan.logical.impl`, and it should implement the `UnaryLogicalOp` interface.

As an example of another such logical operator, take a look at `LogicalOpTPAdd`. You can essentially copy this class and replace its two member variables by a member variable of type `Expr` (i.e., exactly as you have it in the new executable operator in this PR here).

When you do this, you will notice that you also have to extend the `LogicalPlanVisitor` interface with an additional function for the new logical operator, and then you will also have to extend all classes that implement this interface (because they need to provide this additional function). The following list lists all these classes and what the added function should do:
* LogicalPlanCounter: do the same as for LogicalOpTPAdd
* LogicalPlanPrinterAfterVisitor: do the same as for LogicalOpTPAdd
* LogicalPlanPrinterBeforeVisitor: do the same as for LogicalOpTPAdd
* LogicalPlanVisitorBase: do the same as for LogicalOpTPAdd

Let me know if you have questions about this task.

PR #161, opened 2022-02-16
First commit VocabularyMappings with questions

JulietteVV:


src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ModelFactory;
+import org.apache.jena.riot.RDFDataMgr;
+
+import se.liu.ida.hefquin.engine.data.VocabularyMapping;
+
+public class VocabularyMappingImpl implements VocabularyMapping{
+	
+	Model vocabularyMapping;
```
hartig, line 11:
```suggestion
	protected final Model vocabularyMapping;
```

hartig: Note that the previous comment is a change suggestion which you can accept and apply directly here in the PR. Doing so creates a new commit for your branch. Therefore, after accepting such suggestions, before you start implementing more changes on your computer, you need to do a `git pull` in your local branch in order to receive these new commits into your local branch.

src/main/java/se/liu/ida/hefquin/engine/federation/FederationMember.java
```diff
@@ -1,8 +1,12 @@
 package se.liu.ida.hefquin.engine.federation;
 
+import se.liu.ida.hefquin.engine.data.VocabularyMapping;
 import se.liu.ida.hefquin.engine.federation.access.DataRetrievalInterface;
 
 public interface FederationMember
 {
 	DataRetrievalInterface getInterface();
+	
+	VocabularyMapping getVocabularyMapping();
```
hartig, line 10:
```suggestion
	/**
	 * For federation members for which sub-queries need to be rewritten
	 * based on a particular vocabulary mapping, this function return that
	 * mapping. For the federation members for which the sub-queries can
	 * be used as is, this function returns <code>null</code>.
	 */
	VocabularyMapping getVocabularyMapping();
```

hartig: Another directly-applicable change. It just adds some documentation of this new function.

src/main/java/se/liu/ida/hefquin/engine/federation/SPARQLEndpoint.java
```diff
@@ -1,9 +1,13 @@
...
+import se.liu.ida.hefquin.engine.data.VocabularyMapping;
 import se.liu.ida.hefquin.engine.federation.access.SPARQLEndpointInterface;
 
 public interface SPARQLEndpoint extends FederationMember
 {
 	@Override
 	SPARQLEndpointInterface getInterface();
+	
+	@Override
+	VocabularyMapping getVocabularyMapping();
```
hartig, line 12: That's not necessary and should be removed again. Sub-interfaces inherit all the functions of their parent interface.

(Now, you may wonder why we repeat the `getInterface()` function here in the sub-interface because that one is also already declared in the parent interface. The reason is that, for this sub-interface, we restrict the return type of that function.)

src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
```diff
@@ -173,6 +186,11 @@ else if (req.getQueryPattern() instanceof TriplePattern) {
 			}
 			return new SolMapsResponseImpl( result, this, req, new Date() );
 		}
+		
+		@Override
+		public VocabularyMapping getVocabularyMapping() {
+			return vMapping;
```
hartig, line 192:
```suggestion
			return null;
```

hartig (changes requested):
Juliette, I have added several comments for the different parts of your PR, including responses to the questions that you had included in the code.

/cc @scferrada

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,23 @@
...
+	public VocabularyMappingImpl() {
+		vocabularyMapping = ModelFactory.createDefaultModel();
+	}
+	
+	
+	// Source: https://jena.apache.org/documentation/io/rdf-input.html
+	public VocabularyMappingImpl(final String rdfFile) {
+		vocabularyMapping = RDFDataMgr.loadModel(rdfFile); //.nt file for N-Triple
+	}
+
```
scferrada, line 22: Maybe for testing it would be best to have a constructor that receives a collection of Triples and loads them into the model:

```
	public VocabularyMappingImpl(final Collection<Triple> triples { ... }
```

scferrada (changes requested):
Hi, I just had an extra comment.

src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
```diff
@@ -0,0 +1,25 @@
...
+	 * triple pattern does not have any effect), then the result of this
+	 * function is a singleton set that simply contains the given triple
+	 * pattern.
+	 */
+	
+	/**
+	 * TODO: Is TripplePattern enough? Which data structure to use?
+	 * Unions and intersections also need to be represented
+	 * Would Graph be a fitting? Both Jena union and intersection are graphs
+	 */
```
hartig, line 22: I assume that this question is about the return type and not the argument of this function.

If so, when I proposed this return type to be a set of triple patterns, I was assuming that the physical operator(s) that obtain this set will consider the triple patterns in this set to form a union. In other words, the physical operator may take these resulting triple patterns and issue a request for each of them, one after another; and, then, the solution mappings produced from the responses to these requests can all be output of the operator (which is, essentially, making a union of the responses).

However, now you mention intersection. I am not entirely sure where intersection may come into play. May understanding so far was that your approach uses a sequence of three functions (translateSubject, translatePredicate, and translateObject) where:
1. translateSubject may produce a set of triple patterns (to be considered as a union) 
2. then, for each of these triple patterns, translatePredicate is invoked, which also may produce a set of triple patterns for each input triple pattern---however, all of these triple patterns together (i.e., for all input triple patterns coming from translateSubject) can also still be considered as a single union
3. then the same happens again with translateObject

So, in the end, there is still only one set of triple patterns that can be considered as a union. Is that how it works or did I overlook something and things are more complicated?

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,116 @@
...
+	@Override
+	//TODO: Which data structure to use? Unions and intersections also need to be represented
+	public Set<TriplePattern> translateTriple(TriplePattern tp) {
+		Set<TriplePattern> results = translateSubject(tp);
+		return results;
+	}
+	
+	private Set<TriplePattern> translateSubject(TriplePattern tp){
+		Set<TriplePattern> results = Collections.emptySet();
```
hartig, line 61:
```suggestion
	protected Set<TriplePattern> translateSubject(final TriplePattern tp){
		final Set<TriplePattern> results = Collections.emptySet();
```

hartig: Comment about my suggested change: For internal helper functions such as this one, unless there is a good reason for making it private, it is better to make it only protected. This makes it possible if we later want to implement a subclass that extends some of the functionality (by overriding some of these internal helper functions).

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,116 @@
...
+		if (tp.getSubject().isVariable()) {
```
hartig, line 62: This does not work. Our `TriplePattern` interface does not have a function called `getSubject()`. You have to do:
```suggestion
		if ( tp.asJenaTriple().getSubject().isVariable() ) {
```

...ah, now I see that you have imported the from `TriplePattern`. The one that you need to use is the one that we have in the HeFQUIN code, in the Java package `se.liu.ida.hefquin.engine.query`.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,116 @@
...
+		
+		/* OR ?
+		 * Selector selector = new SimpleSelector(tp.getSubject(), null, null);
+		 * Iterator i = vocabularyMapping.listStatements(selector);
+		 */
+		
+		Iterator<TriplePattern> i = mappings.iterator();
+		while (i.hasNext()) {
+			TriplePattern m = i.next();
+			if (m.getPredicate().toString() == "owl:sameAs") {
```
hartig, line 79: This won't work. Notice that 'owl:sameAs' is just a short form for the full URI 'http://www.w3.org/2002/07/owl#sameAs' and can be used only in contexts in which the prefix 'owl' is declared, which is not the case here. So, you need to use the full URI for the comparison.

Having said that, the Jena API has helper classes with static member variable for the most common vocabularies, including the OWL vocabulary and the RDFS vocabulary. You find them in the Java package `org.apache.jena.vocabulary` of Jena. For instance, the one of 'owl:sameAs' is `OWL.sameAs`. Using these ones helps to avoid making errors with wrongly types full URIs.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,116 @@
...
+				TriplePattern translation = new TriplePattern(m.getObject(), tp.getPredicate(), tp.getObject());
+				results.add(translation);
+			}
+		}
+		return results;
+	}
+	
+	private Set<TriplePattern> getMappings(TriplePattern tp){
```
hartig, line 87: Are you sure that the return type of this function should be a set of triple patterns? I would assume that this should be a set of triples instead.

src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
```diff
@@ -143,8 +144,10 @@ public FederationMemberBaseForTest( final Graph data ) {
 		
 	}
 
+
```
hartig, line 147:
```suggestion
```

src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
```diff
@@ -143,8 +144,10 @@ public FederationMemberBaseForTest( final Graph data ) {
 		
 	}
 
+	
 	protected static class SPARQLEndpointForTest extends FederationMemberBaseForTest implements SPARQLEndpoint
 	{
+
```
hartig, line 150:
```suggestion
```

hartig (changes requested):
Here is my feedback on your recent changes.

src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
hartig: Something is wrong with the example. `owl:union` and `owl:intersectionOf` are not for properties but for classes, right?

Do you have any other examples where you would get a join (AND) of triple patterns?
Also, can it even be a join of something more complex than triple patterns (e.g., unions or other joins)?

hartig: @JulietteVV No matter what your answers to my questions in my previous comment are, I see now that the return type of this function should not be `Set<TriplePattern>`. Even if it is only about making the meaning of this set more explicit, the return type should be an actual representation of a union of triple patterns (for the cases in which such a union is the intended result). Therefore, I propose that you change the return type to the general interface `SPARQLGraphPattern` (in `se.liu.ida.hefquin.engine.query`). Then, there are a number of relevant sub-interfaces that you can use for the actual results, depending on what the result is:
* If the result is just a single triple pattern, return a `TriplePattern` object.
* If the result is a join of triple patterns, return a `BGP` object.
* If the result is a union of triple patterns (or, more generally, a union of some graph patterns), return a `SPARQLUnionPattern` object.

The corresponding classes that implement these interfaces and that you can use can be found in the package `se.liu.ida.hefquin.engine.query.impl`.

Regarding the `SPARQLUnionPattern` interface, I have just added this interface (plus an implementation) to your branch for this purpose. Hence, you need to do a `git pull` to get it in your local branch.

Let me know if you have any questions.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
scferrada: Just to chime in, when comparing Strings, you need to use the equals function instead of the `==` operator. Something ike this:
```
m.getPredicate().toString().equals(OWL.sameAs.toString())
```

hartig: Good point Sebastian!

In fact, when comparing Java objects for equivalence, it is always a good idea to use 'equals'.  == is useful mostly for comparing atomic values.

scferrada (commented):
I had one more small comment, on top of the equals comment I made on Olaf's review.

I'm starting to think that this PR is getting out of scope, since the idea was only to have the basic model for Vocabulary Mappings. Now with the translation algorithms, several issues have appeared, and unit tests need to be added. 

@hartig do you think that we can push the translations of triple patterns to the next PR?

src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
```diff
@@ -1,8 +1,25 @@
...
+		
+	Set<TriplePattern> translateTriple( TriplePattern tp );
```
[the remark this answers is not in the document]
JulietteVV: Should I remove the implementation of the translation function for now, and create a new branch and PR for it once the VocabularyMapping looks good?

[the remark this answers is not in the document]
hartig: Good point Sebastian!

Juliette, yes, remove the implementation of the translation function from this PR. More precisely, save it somewhere else so that you can use it for the next PR, and in this PR simply `return null` in this function.

src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
```diff
@@ -0,0 +1,20 @@
...
+
+	/**
+	 * Applies this vocabulary mapping to the given triple pattern and
+	 * returns the resulting triple patterns. If this mapping is not relevant
+	 * for the given triple pattern (i.e., applying the mapping to the given
+	 * triple pattern does not have any effect), then the result of this
+	 * function is a singleton set that simply contains the given triple
+	 * pattern.
+	 */
+	Set<SPARQLGraphPattern> translateTriplePattern( TriplePattern tp );
```
hartig, line 19: The idea was to return just a `SPARQLGraphPattern` and not a set of them. Here is the suggested change:
```suggestion
import se.liu.ida.hefquin.engine.query.SPARQLGraphPattern;
import se.liu.ida.hefquin.engine.query.TriplePattern;


public interface VocabularyMapping {

	/**
	 * Applies this vocabulary mapping to the given triple pattern and
	 * returns the resulting combination of triple patterns, which may be
	 * a union of triple patterns (captured as a {@link SPARQLUnionPattern})
	 * or just a single triple pattern (captured as a {@link TriplePattern}).
	 * If this mapping is not relevant for the given triple pattern (i.e.,
	 * applying the mapping to the given triple pattern does not have any
	 * effect), then the result of this function is simply the given triple
	 * pattern itself.
	 */
	SPARQLGraphPattern translateTriplePattern( TriplePattern tp );
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,59 @@
...
+	public VocabularyMappingImpl(final String rdfFile) {
+		vocabularyMapping = RDFDataMgr.loadModel(rdfFile); //.nt file for N-Triple
+	}
+	
+	public VocabularyMappingImpl(final Set<Triple> tps) {
+		vocabularyMapping = ModelFactory.createDefaultModel();
+		Iterator<Triple> i = tps.iterator();
+		while(i.hasNext()) {
+			Triple tp = i.next();
+			vocabularyMapping.asStatement(tp.asJenaTriple());
```
hartig, line 34: The names of the program variables that you are using here (`tp`and `tps`) seem to indicate that this is about triple patterns rather than triples. To avoid any confusion I suggest to rename them (additionally, mark all local variables as `final` if you are not intending to change them):
```suggestion
	public VocabularyMappingImpl(final Set<Triple> triples) {
		vocabularyMapping = ModelFactory.createDefaultModel();
		final Iterator<Triple> i = triples.iterator();
		while(i.hasNext()) {
			final Triple t = i.next();
			vocabularyMapping.asStatement(t.asJenaTriple());
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,59 @@
...
+		}
+	}
+
+	@Override
+	public Set<SPARQLGraphPattern> translateTriplePattern(final TriplePattern tp) {
```
hartig, line 39:
```suggestion
	public SPARQLGraphPattern translateTriplePattern(final TriplePattern tp) {
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,59 @@
...
+		return null;
+	}
+	
+	protected Set<TriplePattern> translateSubject(final TriplePattern tp){
+		return null;
+	}
+	
+	protected Set<SPARQLGraphPattern> translateObject(final TriplePattern tp){
```
hartig, line 47:
```suggestion
	protected SPARQLGraphPattern translateObject(final TriplePattern tp){
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,59 @@
...
+		return null;
+	}
+	
+	protected Set<SPARQLGraphPattern> translatePredicate(final TriplePattern tp){
```
hartig, line 51:
```suggestion
	protected SPARQLGraphPattern translatePredicate(final TriplePattern tp){
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,59 @@
...
+		return null;
+	}
+	
+	protected Set<Triple> getMappings(TriplePattern tp){
```
hartig, line 55:
```suggestion
	protected Set<Triple> getMappings(final TriplePattern tp){
```

hartig (changes requested):
A few more things to consider (all given as change suggestions that you can directly apply here in the PR), and then the PR will be ready to be merged.

src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
```diff
@@ -142,7 +143,7 @@ public FederationMemberBaseForTest( final Graph data ) {
 		}
 		
 	}
-
+
```
hartig, line 146:
```suggestion

```

src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
```diff
@@ -156,7 +157,7 @@ public FederationMemberBaseForTest( final Graph data ) {
 		public SPARQLEndpointForTest( final String ifaceURL, final Graph data ) {
 			super(data);
 			iface = new SPARQLEndpointInterfaceImpl(ifaceURL);
-		}
+		}
```
hartig, line 160:
```suggestion
		}
```

hartig:
... besides the change suggestions in my code review, I see that the automatic build process for this PR is still failing due to compilation errors. You can see the build log here:
https://github.com/LiUSemWeb/HeFQUIN/runs/5285052923?check_suite_focus=true

Please remove these errors as well.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
scferrada, line 34: Are you sure that this adds the statement into the model? The javadoc of the method `asStatement` is a bit cryptic, but it seems to me that it takes a Triple and transforms it into a Statement, but I'm not sure that it also adds the Statement object to the Model. Please, check this, if it doesn't work, then maybe you need to do something like:

```
vocabularyMapping.add(vocabularyMapping.asStatement(t.asJEnaTriple());
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -0,0 +1,59 @@
...
+			vocabularyMapping.asStatement(tp.asJenaTriple());
+		}
+	}
+
+	@Override
+	public Set<SPARQLGraphPattern> translateTriplePattern(final TriplePattern tp) {
+		return null;
+	}
+	
+	protected Set<TriplePattern> translateSubject(final TriplePattern tp){
```
scferrada, line 43: `translateSubject` should also return `SPARQLGraphPattern`?

scferrada (commented):
A few more comments/questions.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
hartig: I think that depends on how Juliette will want to implement the overall `translateTriplePattern` function. For that, it might in fact be a bit more convenient to simply receive a set of triple patterns from this function here.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
hartig: Another even easier option: instead of creating a `Model` initially, create a Jena `Graph` object (there is a `GraphFactory` helper class to do that), add the triples into this `Graph`, and finally create the `Model` based on that `Graph`.

In fact, it may not be necessary to have a `Model` at all but, instead, simply have it only as a `Graph` object.

JulietteVV: If the graph isn't made into a Model in the constructor, should an extra function be added to create a model out of it, to be able to query the model later? Or is it okay to create a model out of it directly in the getMappings function?

JulietteVV:
> 

I managed to fix the errors that I could see which were related to my changes. I still got build errors in eclipse in the test files, something about functional interfaces. I don't know if these are related to my changes or how to fix them?

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
JulietteVV: I think I missed that find can be used on the graph directly, and with that a Model would not be necessary right?

scferrada: Right.

src/main/java/se/liu/ida/hefquin/engine/federation/SPARQLEndpoint.java
```diff
@@ -6,4 +6,5 @@
 {
 	@Override
 	SPARQLEndpointInterface getInterface();
+
```
hartig, line 9:
```suggestion
```

hartig (approved):
From my perspective, this PR is now ready.

@scferrada would you approve it as well? If so, then I can merge it.

scferrada (approved):

scferrada:
Yes, please merge it.

@JulietteVV I think next PR should be about the translation method then.

hartig:
Perfect. I have merged the PR into the main branch and deleted the corresponding branch on the server

@JulietteVV just in case you don't know: What you can/should do on your local computer now is to switch to the main branch:
```
git checkout main
```
Then, pull the updates for that branch from the server
```
git pull
```
You may also remove your local copy of the `VocabularyMappings` branch:
```
git branch -d VocabularyMappings
```
And finally you can create a new branch locally and change into that one for the next part (implementation of the translation method). Just make sure that you are in the up-to-date main branch when you create the new branch.

PR #162, opened 2022-02-23
Translate triple pattern

JulietteVV:


hartig:
I see that now you are using the Jena `Element` classes internally, and then create a `SPARQLGraphPatternImpl`. Unfortunately, that was not exactly my idea, and I realize now that I should have been more clear about it in the related comments that I wrote earlier. While the idea was to use our `SPARQLGraphPattern` as return type, I was assuming that the actual types of objects that you return are the sub-interfaces such as `SPARQLUnionPattern`, `BGP`, and our `TriplePattern`. Each of these sub-interfaces has its own, more specific implementing class which is not using Jena's `Element`. Can you please change your code to remove everything about the Jena `Element` classes and `SPARQLGraphPatternImpl`, and instead use the interface `SPARQLGraphPattern` really only as return type, together with the relevant sub-interfaces `SPARQLUnionPattern`, `BGP`, and our `TriplePattern`. I am really really sorry to cause you this extra work!

hartig (changes requested):
Here are some comments already for the `translateTriplePattern` function. I have a meeting now and will get back to the rest later. You can already start working on my comments in the meantime.

hartig:
@JulietteVV I have added a new interface called `SPARQLGroupPattern` (see commit: https://github.com/LiUSemWeb/HeFQUIN/pull/162/commits/41cb66eb75f873a1a55c0b713ad7bf5a502db3eb) and changed the code block where a BGP is the input to the predicate-related translation such that this code block uses this new interface if necessary (see commit: https://github.com/LiUSemWeb/HeFQUIN/pull/162/commits/8ab44b661b94c5de5caba69e23ec6189ec28c1a1).

Additionally, I have added a public (!) helper function called `addSubPattern` to the class `SPARQLUnionPatternImpl` (see commit: https://github.com/LiUSemWeb/HeFQUIN/pull/162/commits/68e3221c290d25c5f88e8dd43e55055b717d5e6a). With this you can simplify your code a bit more. That is, you do not anymore have to first create an initially empty `List<SPARQLGraphPattern>`, then populate this list, and finally pass this list to the constructor of `SPARQLUnionPatternImpl`. Instead, you can now simply create an initially empty `SPARQLUnionPatternImpl` (do: `new SPARQLUnionPatternImpl()`) and then populate this union pattern directly by using the new function.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -154,39 +153,68 @@ protected SPARQLGraphPattern translateObject(final TriplePattern tp){
...
-				 * results.add(union);
+				/*
+				 * Blank node used to help create intersections and unions
+				 * (a equals b)
+				 * (b unionof c)
+				 * (b unionof d)
 				 */
+				if (m.asJenaTriple().getObject().isBlank()) {
+					//Union or intersection
+					Node blankP = NodeFactory.createVariable("p");
```
scferrada, line 175: If you intend to create blank nodes, you should use `NodeFactory.createBlankNode("p")`

scferrada (commented):
Good work Juliette. Now I think you should add some unit tests to this PR to check that the process works as expected.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -61,20 +61,19 @@ public SPARQLGraphPattern translateTriplePattern(final TriplePattern tp) {
 		List<SPARQLGraphPattern> predicateTranslation = Collections.<SPARQLGraphPattern>emptyList();
 		for(final SPARQLGraphPattern j : objectTranslation) {
 			
-			if(j instanceof SPARQLUnionPattern) { 
-				List<SPARQLGraphPattern> unionList = Collections.<SPARQLGraphPattern>emptyList();
+			if(j instanceof SPARQLUnionPattern) {
+				SPARQLUnionPatternImpl union = new SPARQLUnionPatternImpl();
```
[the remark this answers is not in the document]
JulietteVV: I found that for SPARQLUnionPattern specifically, I hade to use the Impl version to be able to call the addSupPattern function. Should I access this function another way?

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
JulietteVV: I did not mean to create blank nodes here, but I changed the names of the nodes since I realised that they were a bit confusing.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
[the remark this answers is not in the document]
hartig: Just some clarification: While Sebastian is right in general, in this specific case it is indeed correct to use `SPARQLUnionPatternImpl` rather than `SPARQLUnionPattern`, because `addSubPattern` is not declared in the interface but only in the implementing class. See my earlier comment: https://github.com/LiUSemWeb/HeFQUIN/pull/162#issuecomment-1048895270

Now you may ask why I did not add the `addSubPattern` function also to the interface. The reason is that I would generally like these interfaces to represent immutable objects.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -154,39 +153,68 @@ protected SPARQLGraphPattern translateObject(final TriplePattern tp){
...
+				/*
+				 * Blank node used to help create intersections and unions
+				 * (a equals b)
+				 * (b unionof c)
+				 * (b unionof d)
 				 */
+				if (m.asJenaTriple().getObject().isBlank()) {
+					//Union or intersection
+					Node blankP = NodeFactory.createVariable("p");
+					Node blankO = NodeFactory.createVariable("o");
```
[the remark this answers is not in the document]
hartig: I want to second Sebastian's comment!!!
Making these program variables `final` is really a safety measure for ourselves when programming.

JulietteVV:
> Good work Juliette. Now I think you should add some unit tests to this PR to check that the process works as expected.

I'm not very familiar with "formal" testing, such as with JUnit. What would be the first step in making some unit tests?

hartig:
Just one addition to Sebastián's description: once you have this class, you can run its tests as follows. Right-click on the class name, go to 'Run as' in the pop-up menu, and select 'as JUnit Test'.

JulietteVV:
I struggled a little bit with the assertEqual when it came to the TriplePattern Sets, to test if the VocabularyMappings where created correctly. But the translation test is working for this case now. 

I had issues with adding to sets and lists. I had to change all empty list to new ArrayList and empty set to new HashSet. I also couldn't use the union addSubPattern. I changed in the places that where needed for the tests to pass, should I change it everywhere or did I make a mistake when trying to create/add to sets and list orignially?

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+			return translation;
+		} else {
+			return predicateTranslation.get(0);
+		}
 	}
 	
 	protected Set<TriplePattern> translateSubject(final TriplePattern tp){
-		return null;
+		Set<TriplePattern> results = new HashSet<TriplePattern>();
+		org.apache.jena.graph.Triple t = tp.asJenaTriple();
```
hartig, line 134:
```suggestion
		final Set<TriplePattern> results = new HashSet<>();
		final org.apache.jena.graph.Triple t = tp.asJenaTriple();
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+			} else {
+				throw new IllegalArgumentException(m.asJenaTriple().getPredicate().getURI());
+			}
+		}
+		return results;
 	}
 	
 	protected SPARQLGraphPattern translateObject(final TriplePattern tp){
-		return null;
+		org.apache.jena.graph.Triple t = tp.asJenaTriple();
```
hartig, line 154:
```suggestion
		final org.apache.jena.graph.Triple t = tp.asJenaTriple();
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+		final Node p = NodeFactory.createVariable("p");
+		final Node o = NodeFactory.createVariable("o");
+		final TriplePattern tpQuery = new TriplePatternImpl(t.getObject(), p, o);		
+		List<SPARQLGraphPattern> resultsList = new ArrayList<SPARQLGraphPattern>();
+		for (final Triple m : getMappings(tpQuery)) {
+			String predicate = m.asJenaTriple().getPredicate().getURI();
+			
+			if(predicate.equals(OWL.sameAs.getURI())) {
+				final TriplePattern translation = new TriplePatternImpl(t.getSubject(), t.getPredicate(), m.asJenaTriple().getObject());
+				resultsList.add(translation);
```
hartig, line 173: Related to my comment above, it seems that you are indeed aiming to using owl:sameAs mappings for the objects.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+				/*
+				 * Blank node used to help create intersections and unions
+				 * (a equals b)
+				 * (b unionof c)
+				 * (b unionof d)
+				 */
+				if (m.asJenaTriple().getObject().isBlank()) {
+					//Union or intersection
+					final Node newP = NodeFactory.createVariable("p");
+					final Node newO = NodeFactory.createVariable("o");
```
hartig, line 185: No need for this. You can reuse the same `Node` objects that you have created above.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+					final TriplePattern blankQuery = new TriplePatternImpl(m.asJenaTriple().getObject(), newP, newO);
+					String newPredicate = "";
+					List<Node> objects = new ArrayList<Node>();
+					for(final Triple i : getMappings(blankQuery)) {
+						String iPredicate = i.asJenaTriple().getPredicate().getURI();
```
hartig, line 190:
```suggestion
						final String iPredicate = i.asJenaTriple().getPredicate().getURI();
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+						}
+						objects.add(i.asJenaTriple().getObject());
+					}
+					
+					if (newPredicate.equals(OWL.unionOf.getURI())) {
+						//SPARQLUnionPatternImpl union = new SPARQLUnionPatternImpl();
+						List<SPARQLGraphPattern> unionList = new ArrayList<SPARQLGraphPattern>();
+						for(Node j : objects){
+							final TriplePattern translation = new TriplePatternImpl(t.getSubject(), t.getPredicate(), j);
+							//union.addSubPattern(translation)
```
hartig, line 206: Why exactly did that didn't work?? Of course, you need a semicolon here but, other than that, I don't see a reason why you cannot use this one directly.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+						resultsList.add(union);
+						
+					} else if (newPredicate.equals(OWL.intersectionOf.getURI())) {
+						 Set<TriplePattern> intersectionList = new HashSet<TriplePattern>();;
+						 for(Node j : objects){
+						  	final TriplePattern translation = new TriplePatternImpl(t.getSubject(), t.getPredicate(), j);
+						  	intersectionList.add(translation);
+						 }
+						 final BGP intersection = new BGPImpl(intersectionList);
+						 resultsList.add(intersection);
```
hartig, line 219: To simplify this part you may also add a public `addTriplePattern` function to the `BGPImpl` class (in the same way as I did with `addSubPattern` in `SPARQLUnionPatternImpl`

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+import se.liu.ida.hefquin.engine.query.impl.SPARQLUnionPatternImpl;
+import se.liu.ida.hefquin.engine.query.impl.TriplePatternImpl;
+
+public class VocabularyMappingTest {
+
+	@Test
+	public void VocabularyMappingConstructorTest() {
+		final Set<Triple> testTriples = CreateTestTriples();
+		final List<Triple> triplesList = new ArrayList<Triple>();
+		triplesList.addAll(testTriples);
```
scferrada, line 31: You can add all the elements of another collection directly in the constructor. I'm not sure though why do you need to have both a set and a list.
```suggestion
		final List<Triple> triplesList = new ArrayList<Triple>(testTriples);
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		final VocabularyMapping vm = new VocabularyMappingImpl(testTriples);
+		
+		final Node s = NodeFactory.createVariable("s");
+		final Node p = NodeFactory.createVariable("p");
+		final Node o = NodeFactory.createVariable("o");
+		final org.apache.jena.graph.Triple testQuery = new TriplePatternImpl(s, p, o).asJenaTriple();
+		List<Triple> QueryResults = new ArrayList<Triple>();
```
scferrada, line 38:
```suggestion
		final List<Triple> queryResults = new ArrayList<>();
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		final Iterator<org.apache.jena.graph.Triple> i = vm.getVocabularyMappingAsGraph().find(testQuery);
+		while(i.hasNext()) {
+			final Triple t = new TripleImpl(i.next());
+			QueryResults.add(t);
```
scferrada, line 42:
```suggestion
			queryResults.add(t);
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		}
+		//assertEquals(QueryResults, triplesList);
```
scferrada, line 44: Why did this not work? Did the test fail? You may need to implement the `equals` method in `Triple` for this `assertEquals` to work correctly.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+import se.liu.ida.hefquin.engine.query.SPARQLGraphPattern;
+import se.liu.ida.hefquin.engine.query.SPARQLUnionPattern;
+import se.liu.ida.hefquin.engine.query.TriplePattern;
+import se.liu.ida.hefquin.engine.query.impl.SPARQLUnionPatternImpl;
+import se.liu.ida.hefquin.engine.query.impl.TriplePatternImpl;
+
+public class VocabularyMappingTest {
+
+	@Test
+	public void VocabularyMappingConstructorTest() {
```
scferrada, line 28: In general, I'm not sure about the need for this test, I'm leaving comments anyways because they might be useful for future development.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		final SPARQLUnionPattern correct = new SPARQLUnionPatternImpl(union);
+		
+		//System.out.print(subPatterns.toString() + "\n");
+		//System.out.print(union.toString() + "\n");
+
+		assertEquals(union, subPatterns);
+	}
+	
+	public Set<Triple> CreateTestTriples(){
+		Set<Triple> testSet = new HashSet<Triple>();
```
scferrada, line 81:
```suggestion
		final Set<Triple> testSet = new HashSet<>();
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		
+		Node s = NodeFactory.createLiteral("s1");
```
scferrada, line 83:
```suggestion
		final Node s = NodeFactory.createLiteral("s1");
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		Node p = NodeFactory.createURI(OWL.sameAs.getURI());
```
scferrada, line 84: Isn't `OWL.sameAs` already a `Node`? Make this variable `final`

Edit, it is not, but it is a `Property`, so you can use `OWL.sameAs.asNode()`

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		Node o = NodeFactory.createLiteral("s2");
+		Triple t = new TripleImpl(s, p, o);
+		testSet.add(t);
+		
+		s = NodeFactory.createURI(RDF.type.getURI());
+		p = NodeFactory.createURI(OWL.inverseOf.getURI());
```
scferrada, line 90: Same here, use the `asNode` method

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+	public void TranslateTriplePatternTest() {
+		final Set<Triple> testTriples = CreateTestTriples();
+		final VocabularyMapping vm = new VocabularyMappingImpl(testTriples);
+		
+		final Node s = NodeFactory.createLiteral("s1");
+		final Node p = NodeFactory.createURI(RDF.type.getURI());
+		final Node o = NodeFactory.createLiteral("o1");
+		final TriplePattern testTp = new TriplePatternImpl(s, p, o);
+		final SPARQLGraphPattern translation = vm.translateTriplePattern(testTp);
+		List<SPARQLGraphPattern> subPatterns = new ArrayList<SPARQLGraphPattern>();
```
scferrada, line 57:
```suggestion
		final List<SPARQLGraphPattern> subPatterns = new ArrayList<>();
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		for(SPARQLGraphPattern gp : ((SPARQLUnionPattern)translation).getSubPatterns()) {
+			subPatterns.add(gp);
```
scferrada, line 59: Why can't you use the `Set` resulting from `getSubPatterns` directly? If you really need a list you can do it like this:

`final List<SPARQLGraphPattern> subPatterns = new ArrayList<>(translation.getSubPatterns());`

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		final Set<Triple> testTriples = CreateTestTriples();
+		final VocabularyMapping vm = new VocabularyMappingImpl(testTriples);
+		
+		final Node s = NodeFactory.createLiteral("s1");
+		final Node p = NodeFactory.createURI(RDF.type.getURI());
+		final Node o = NodeFactory.createLiteral("o1");
+		final TriplePattern testTp = new TriplePatternImpl(s, p, o);
+		final SPARQLGraphPattern translation = vm.translateTriplePattern(testTp);
+		List<SPARQLGraphPattern> subPatterns = new ArrayList<SPARQLGraphPattern>();
+		for(SPARQLGraphPattern gp : ((SPARQLUnionPattern)translation).getSubPatterns()) {
```
scferrada, line 58: Here instead of just casting, you can also check that `translation` is indeed of type `SPARQLUnionPattern` by using:

`assertTrue(translation instanceof  SPARQLUnionPattern)`

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		}
+		
+		
+		final Node sTranslated = NodeFactory.createLiteral("s2");
+		final Node pTranslated = NodeFactory.createLiteral("Not type");
+		final Node oTranslated1 = NodeFactory.createLiteral("o2");
+		final Node oTranslated2 = NodeFactory.createLiteral("o3");
+		final TriplePattern tp1 = new TriplePatternImpl(oTranslated1, pTranslated, sTranslated);
+		final TriplePattern tp2 = new TriplePatternImpl(oTranslated2, pTranslated, sTranslated);	
+		List<SPARQLGraphPattern> union = new ArrayList<SPARQLGraphPattern>();
```
scferrada, line 69:
```suggestion
		final List<SPARQLGraphPattern> union = new ArrayList<>();
```

Adding elements to a list or set is not the same as reassigning the variable, so you can make these collections `final` even if you add elements to them.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,113 @@
...
+		Node p = NodeFactory.createURI(OWL.sameAs.getURI());
+		Node o = NodeFactory.createLiteral("s2");
+		Triple t = new TripleImpl(s, p, o);
+		testSet.add(t);
+		
+		s = NodeFactory.createURI(RDF.type.getURI());
+		p = NodeFactory.createURI(OWL.inverseOf.getURI());
+		o = NodeFactory.createLiteral("Not type");
+		t = new TripleImpl(s, p, o);
+		testSet.add(t);
```
scferrada, line 93:
```suggestion
		testSet.add(new TripleImpl(s, p, o));
```
Repeat this for the following assignments of `t`.

scferrada (changes requested):
Hi Juliette, here are my comments about the unit tests.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -40,23 +54,293 @@ public VocabularyMappingImpl(final Set<Triple> triples) {
...
+		org.apache.jena.graph.Triple t = tp.asJenaTriple();
+		if(!t.getPredicate().isURI()) {
+			//System.out.print("Not URI : " + t.getPredicate().toString());
+			return tp;
+		} else {
+			if(!t.getPredicate().getURI().equals(RDF.type.getURI())) {
+				//System.out.print("Not RDF type : " + t.getPredicate().getURI());
+				return tp;
+			}
+		}
```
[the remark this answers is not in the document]
JulietteVV: It was mistake, I missed one of the cases in the translation

hartig:
Juliette, just a general word because I can imagine that all these comments may give you frustration and may make you think that you are not getting it right. That's not the intention! You are doing a good job and are making good progress! What you have done absolutely does go in the right direction. Our comments are really just in the details, and I can also see that you already have been learning from them.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
JulietteVV: I get an Unsupported Operation Exception when using this. It is the same error that I get when trying to add to lists/set created with Collections.EmptySet or EmptyList.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
JulietteVV: If I'm guessing, this comparison did not work because I was creating new triples from the query results. I changed this to sets of Jena Triples instead so that the results could be compared directly

scferrada: Just to expand a bit on the `assertEquals(Object expected, Object actual)` method. The only think it does is to check if both are not `null` and then call `expected.equals(actual)`. In the case of lists, the equals method is implemented in such a way that it iterates through both lists and checks element by element if they are equal. This is why I first suggested to implement the `equals` method in `Triple`. Since Jena Triples already have an implementation for `equals` it seems that now it works :)

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,105 @@
...
+		final TriplePattern tp2 = new TriplePatternImpl(oTranslated2, pTranslated, sTranslated);	
+		final List<SPARQLGraphPattern> union = new ArrayList<>();
+		union.add(tp1);
+		union.add(tp2);
+		final SPARQLUnionPattern correct = new SPARQLUnionPatternImpl(union);
+		
+		System.out.print(translation.toString() + "\n");
+		System.out.print(correct.toString() + "\n");
+
+		//assertEquals(correct, translation);
```
scferrada, line 74: For this to work correctly you need to implement the `equals` method in all the classes that implement `SPARQLGraphPattern`. Usually with the auto-generated method from eclipse it would suffice (right click on the class > source > generate hashCode() and equals() > Generate).

JulietteVV: After trying multiple solutions, I think my issue is that the triples in the translation and in the expected result don't always come in the same order. When I print in the equals function it seems like the order differes from run to run which causes the test to fail sometimes but pass other. Have I missed to add an equal() somewhere or how would this be fixed?

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
hartig: Ah! I see!! Try again now with the fix in commit https://github.com/LiUSemWeb/HeFQUIN/pull/162/commits/bec0d4d2e12d790648c18f928122edd7359e0839

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
scferrada: Right, in order for lists to be equal, they need to have the elements in the same order. If order is not important, you should consider then having `Set`s of triples instead, both for the return type of the translation method and the solution set that contains the desired output.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -48,58 +45,59 @@ public void VocabularyMappingConstructorTest() {
...
 		
 		final Node s = NodeFactory.createLiteral("s1");
 		final Node p = NodeFactory.createURI(RDF.type.getURI());
 		final Node o = NodeFactory.createLiteral("o1");
 		final TriplePattern testTp = new TriplePatternImpl(s, p, o);
 		final SPARQLGraphPattern translation = vm.translateTriplePattern(testTp);
+		assert(translation instanceof SPARQLUnionPattern);
+		final Set<org.apache.jena.graph.Triple> translatedTriples = new HashSet<>();
+		for(final SPARQLGraphPattern i : ((SPARQLUnionPattern) translation).getSubPatterns()) {
+			assert(i instanceof TriplePattern);
```
scferrada, line 60:
```suggestion
			assertTrue(i instanceof TriplePattern);
```
Please use the JUnit method to assert, also in line 57

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -48,58 +45,59 @@ public void VocabularyMappingConstructorTest() {
...
-		final Node oTranslated1 = NodeFactory.createLiteral("o2");
-		final Node oTranslated2 = NodeFactory.createLiteral("o3");
-		final TriplePattern tp1 = new TriplePatternImpl(oTranslated1, pTranslated, sTranslated);
-		final TriplePattern tp2 = new TriplePatternImpl(oTranslated2, pTranslated, sTranslated);	
-		final List<SPARQLGraphPattern> union = new ArrayList<>();
-		union.add(tp1);
-		union.add(tp2);
-		final SPARQLUnionPattern correct = new SPARQLUnionPatternImpl(union);
+		System.out.print(testData.object2);
+		System.out.print(translatedTriples);
```
scferrada, line 65: Remove these for the final version

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -48,58 +45,59 @@ public void VocabularyMappingConstructorTest() {
...
 	
-	public Set<Triple> CreateTestTriples(){
+	public Pair<Set<Triple>,Set<org.apache.jena.graph.Triple>> CreateTestTriples(){
 		final Set<Triple> testSet = new HashSet<>();
 		
 		Node s = NodeFactory.createLiteral("s1");
 		Node p = OWL.sameAs.asNode();
-		Node o = NodeFactory.createLiteral("s2");
-		testSet.add(new TripleImpl(s, p, o));
+		Node sRes = NodeFactory.createLiteral("s2");
```
scferrada, line 75: These (`sRes`, `pRes`, etc.) should be `final` if you are not reassigning them.

scferrada (commented):
Thanks Juliette, just some small comments now.

@hartig should we add more tests to this PR or may we leave them for the next one?

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -0,0 +1,116 @@
...
+		p = OWL.intersectionOf.asNode();
+		final Node o1Res = NodeFactory.createLiteral("o2");
+		testSet.add(new TripleImpl(s, p, o1Res));
+		
+		final Node o2Res = NodeFactory.createLiteral("o3");
+		testSet.add(new TripleImpl(s, p, o2Res));
+		
+		final Set<org.apache.jena.graph.Triple> expectedRes = new HashSet<>();
+		expectedRes.add(new TriplePatternImpl(o2Res, pRes, sRes).asJenaTriple());
+		expectedRes.add(new TriplePatternImpl(o1Res, pRes, sRes).asJenaTriple());
```
scferrada, line 109: Here you should call directly the Jena Triple constructor, instead of creating a `TriplePatternImpl` object just to get its inner Jena Triple
```suggestion
		expectedRes.add(new org.apache.jena.graph.Triple(o2Res, pRes, sRes));
		expectedRes.add(new org.apache.jena.graph.Triple(o1Res, pRes, sRes));
```

hartig:
> @hartig should we add more tests to this PR or may we leave them for the next one?

I think we can leave that for another PR.

@JulietteVV please let me know here when you are done with the remaining changes

JulietteVV:
> > @hartig should we add more tests to this PR or may we leave them for the next one?
> 
> I think we can leave that for another PR.
> 
> @JulietteVV please let me know here when you are done with the remaining changes

@hartig I think all changes are done now!

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -1,62 +1,326 @@
...
+				final SPARQLUnionPatternImpl union = new SPARQLUnionPatternImpl();
+				for(final SPARQLGraphPattern k: ((SPARQLUnionPattern) j).getSubPatterns()) {		
+					if(k instanceof TriplePattern) {
+						union.addSubPattern(translatePredicate((TriplePattern) k));
+					} else if (k instanceof BGP) {
+						for(final TriplePattern l : ((BGP) k).getTriplePatterns()) {
+// TODO: I think that this part here is incorrect, please double check
+							union.addSubPattern(translatePredicate(l));
+						}
+					} else {
```
hartig, line 71: @JulietteVV I have cleaned up the code a bit more. In this context, I have come across this part, which I think it is incorrect. Can you please double check it again. The issue that I mean here is that you are essentially turning a BGP into a union; in other words, the fact the `k` was a BGP gets completely lost.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
JulietteVV: Yes I see now that I forgot to change this, it should be fixed now!

hartig: Yes, that fixes it. Thanks!

hartig (approved):
This version is good now. I will merge it.

JulietteVV:
Should the next step be to create a function to translate a SolutionMapping back to the glocal vocabulary?

scferrada:
I'm not sure. Is that process already defined theoretically? I would prefer to continue with intensive testing, until we have one test for each of the translation cases defined.

JulietteVV:
> I'm not sure. Is that process already defined theoretically? I would prefer to continue with intensive testing, until we have one test for each of the translation cases defined.

I didn't adress translating the results back in the theoretical approach. The approach only discusses translating a subquery. 
I think for my research question, right now, it only touches on the translation of subqueries, but I guess I could make a seperate theoretical approach to translate the results back? 

Should I create a new branch for testing, and keep testing for now?

JulietteVV:
I don't really know, I haven't thought much about what translating the results back would look like. But I'm guessing it's not too different from the subquery translation so, should not take too long. 
But it feels like it would make sense to address it and discuss it, since it's part of the implementation. Maybe it's more important to spend time on that rather than performance impact?

scferrada:
I'm thinking the inverse translation is going to be even easier, probably each solution mapping will produce only one other solution mapping when translated. Could you please make a sketch of the translation of solution mappings? I'll check it when it's done :), probably it's best if you do this while waiting for feedback on the new test PR.

Thanks!

PR #164, opened 2022-03-04
Vocabulary mapping testing

JulietteVV:


scferrada:
Hi Juliette, is this ready for a review? because I see there is still only one test however in your commits you reference more tests, so maybe there is something missing?

JulietteVV:
> Hi Juliette, is this ready for a review? because I see there is still only one test however in your commits you reference more tests, so maybe there is something missing?

I put everything in the same test case, but i'll split it up into several ones!

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -219,6 +234,8 @@ protected SPARQLGraphPattern translateObject(final TriplePattern tp){
 								union.addSubPattern(translation);
 							}
 							resultsList.add(union);
+							for(SPARQLGraphPattern gp : union.getSubPatterns()) {
+							}
```
scferrada, line 238: Is there something missing here?

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -335,7 +352,8 @@ protected SPARQLGraphPattern translatePredicate(final TriplePattern tp){
 		final Set<Triple> mappings = new HashSet<>();
 		final Iterator<Triple> i = vocabularyMapping.find(s,p,o);
 		while(i.hasNext()) {
-			mappings.add(i.next());
+			Triple next = i.next();
```
scferrada, line 355:
```suggestion
			final Triple next = i.next();
```

scferrada: You can accept this change requests by clicking the button "Commit suggestion", then you can get the changes into your local branch with `git pull`. I recommend this, because it will mark the conversation as resolved and thus make the PR shorter to look at.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -74,6 +74,226 @@ public void TranslateTriplePatternTest() {
...
+		for (final SPARQLGraphPattern i : ((SPARQLUnionPatternImpl) translation).getSubPatterns()) {
+			assertTrue(i instanceof TriplePattern);
+			translationTriples.add(((TriplePattern) i).asJenaTriple());
+		}
+			
+		final Set<Triple> expectedResults = new HashSet<>();
+		expectedResults.add(new Triple(o1, p, o));
+		expectedResults.add(new Triple(o2, p, o));
+		
+		assertEquals(translationTriples, expectedResults);
```
scferrada, line 107: The expected results should go first. This is because in case the test fails, it will show you the differences correctly.

```suggestion
		assertEquals(expectedResults, translationTriples);
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -74,6 +74,226 @@ public void TranslateTriplePatternTest() {
...
+		p = OWL.unionOf.asNode();
+		final Node t8 = NodeFactory.createLiteral("p5");
+		testTriples.add(new Triple(t7, p, t8));		
+		final Node t9 = NodeFactory.createLiteral("p6");
+		testTriples.add(new Triple(t7, p, t9));
+
+		final VocabularyMapping vm = new VocabularyMappingImpl(testTriples);
+		
+		p = s;
+		s = NodeFactory.createLiteral("s");
```
scferrada, line 152: Is there really a need to reuse these variables like this? it seems wrong to me to assign s (subject) into p (predicate). Please just use fresh variables every time.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -74,6 +74,226 @@ public void TranslateTriplePatternTest() {
...
+		final Node o = NodeFactory.createLiteral("o");
+		final TriplePattern testTp = new TriplePatternImpl(s, p, o);
+		final SPARQLGraphPattern translation = vm.translateTriplePattern(testTp); 
+		final Set<Triple> translationTriples = new HashSet<>();
+		assertTrue(translation instanceof SPARQLUnionPatternImpl);
+		boolean oneUnion = false;
+		boolean oneIntersection = false;
+		for (final SPARQLGraphPattern i : ((SPARQLUnionPatternImpl) translation).getSubPatterns()) {
```
scferrada, line 160: I'm a bit skeptic with this way of testing. Firstly, it seems to me that it checks for "at least one" union and intersection, not exactly one. On the other hand, I think it should be enough to test the direct object equality between the result of the translation and the expected result: `assertEquals(expectedPattern, translation)`, for which you need to instanciate `expectedPattern` in such a way that it contains the correct structure. For instance, like this:

```
final SPARQLGraphPattern expectedPattern = new SPARQLUnionPatternImpl(
   new TriplePattern(...), new TriplePattern(...), ...
);
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -74,6 +74,226 @@ public void TranslateTriplePatternTest() {
...
+		final Node o = s;
+		s = NodeFactory.createLiteral("s");
+		p = RDF.type.asNode();
+		final TriplePattern testTp = new TriplePatternImpl(s, p, o);
+		final SPARQLGraphPattern translation = vm.translateTriplePattern(testTp); 
+		final Set<Triple> translationTriples = new HashSet<>();
+		assertTrue(translation instanceof SPARQLUnionPatternImpl);
+		boolean oneUnion = false;
+		boolean oneIntersection = false;
+		for (final SPARQLGraphPattern i : ((SPARQLUnionPatternImpl) translation).getSubPatterns()) {
```
scferrada, line 253: Same point as before here.

scferrada (changes requested):
Thanks Juliette, it looks much better now :). Here are some comments regarding the way the testing is done, and how I think would be best to do it. If you have any questions, please just ask. I also left one small question.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -57,13 +57,14 @@ public SPARQLGraphPattern translateTriplePattern(final TriplePattern tp) {
 		final List<SPARQLGraphPattern> predicateTranslation = new ArrayList<>();
 		for(final SPARQLGraphPattern j : objectTranslation) {
 			
-			if(j instanceof SPARQLUnionPattern) {
-				
+			if(j instanceof SPARQLUnionPatternImpl) {
```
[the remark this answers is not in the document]
JulietteVV: I changed this because I was having some issues, but it in the end this wasn't the problem so I'll change it back!

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
JulietteVV: I did this because I thought I needed to use the same node in order for the results to equate, but maybe it isn't the case? I will try to just make new ones

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
JulietteVV: I have tried comparing SPARQLUnionPatterns, my issue is that they do not count as equal even if the members are equal. I have tried fixing this in the equals functions but I do not get it to work. The only comparisons that I've managed to make work are the ones using Jena Triples.

JulietteVV: I did test so that more than one union/intersection will not be accepted.

scferrada: I think that it would be best to investigate why the equals between SPARQLUnionPatternImpl doesn't work correctly. A first step is to implement the `toString` method, so that jUnit shows where the differences between the object are.

JulietteVV: I have pushed a version of the SPARQLUnionPattern equals function, where if the prints are used you can see that the Set contain the same memebers. I made them to sets, as you said earlier that lists take order into consideration. But even then they are not considered equal. I will keep looking at the equals functions of other classes to see where it goes wrong but I've been a bit stuck on this

scferrada: Give me a sec and I'll add some code to help, please don't push anything more yet.

JulietteVV: It seems to me that the order changes, even after changing the order of the expected elements they are still wrong. 

When it comes to inner unions i think it depends on what the union contains, if an intersection is involved I don't think you can simply flaten it? 

image

scferrada: Why does the algorithm not return always things in the same order? is there a non-deterministic step somewhere?

As for the flattening, BGPs are not *exactly* intersections, are joins or even cartesian products. Anyhow, my thought is that:
Union(tp1, tp2, Union(BGP, tp3)) is the same as Union(tp1, tp2, BGP, tp3), right?

JulietteVV: The only step I can think of is the find function that is used to get the mappings. That might return the mappings in different orders, which in turn means that they are added in different orders to the union.

I guess to yes! Should I try to flatten out bthe results then and compare them?

hartig: No, don't sort the mappings in the production code in order to make the unit test work. If the order of the subpatterns in the returned union pattern is nondeterministic, then that's fine; it is still a correct behavior. Therefore, the code of the unit test needs to be extended to accept this correct behavior (any order of the expected subpatterns is okay and should pass the test) instead of adding an expensive but unnecessary ordering process into the runtime code.

src/main/java/se/liu/ida/hefquin/engine/query/impl/BGPImpl.java
```diff
@@ -30,7 +32,19 @@ public boolean equals( final Object o ) {
...
+		final Set<Triple> otpsTriple = new HashSet<>();
+		
+		for(TriplePattern i : tps) {
+			tpsTriple.add(i.asJenaTriple());
+		}
+		
+		for(TriplePattern j : otps) {
+			otpsTriple.add(j.asJenaTriple());
+		}
+
```
scferrada, line 46: Why not having sets of `TriplePattern`??

src/main/java/se/liu/ida/hefquin/engine/query/impl/BGPImpl.java
```diff
@@ -30,7 +32,19 @@ public boolean equals( final Object o ) {
...
+		}
+		
+		for(TriplePattern j : otps) {
+			otpsTriple.add(j.asJenaTriple());
+		}
+		
+		if ( tpsTriple.equals(otpsTriple) )
 			return true;
 		else if ( tps.size() != otps.size() )
 			return false;
```
scferrada, line 50: this can be simplified:
```suggestion
		return tpsTriple.equals(otpsTriple)
```

scferrada (changes requested):
Thanks Juliette, just small things left.

src/main/java/se/liu/ida/hefquin/engine/query/impl/BGPImpl.java
JulietteVV: I realised that when I compared sets the equals function of the objects in the sets where never called. So The TriplePatterns are not considered equal in the sets, but the jena triples are. I don't know if there's any way around this? Lists disn't work because of the order.

src/main/java/se/liu/ida/hefquin/engine/query/impl/SPARQLUnionPatternImpl.java
```diff
@@ -40,13 +43,15 @@ public void addSubPattern( final SPARQLGraphPattern p ) {
...
 	public boolean equals( final Object other ) {
+		
 		if ( this == other )
 			return true;
 
 		if ( other == null || !(other instanceof SPARQLUnionPattern) )
 			return false;
 
-		return ((SPARQLUnionPattern) other).getSubPatterns().equals(subPatterns);
+		return subPatterns.containsAll((Collection<?>) ((SPARQLUnionPattern) other).getSubPatterns());
```
hartig, line 53: That's incorrect for two reasons:
First, it verifies only that every subpattern of the `other` union pattern is among the subpatterns of `this` union pattern; hence, it would return `true` even if `this` union pattern contains additional subpatterns that are not in the `other` union pattern.
Second, it does not consider the order of the subpatterns (`containsAll` checks only for containment not for order).

What was wrong with the original comparison?
```
return ((SPARQLUnionPattern) other).getSubPatterns().equals(subPatterns);
```

src/main/java/se/liu/ida/hefquin/engine/query/impl/BGPImpl.java
hartig: > The TriplePatterns are not considered equal in the sets, but the jena triples are.

That's strange! The `equals` method in `TriplePatternImpl` is implemented correctly (by relying on `equals` of the underlying Jena `Triple` objects). Perhaps our assumptions about `equals` of Java collection types (such as `Set`) are too strong. Maybe they do not check containment based on the `equals` function of the contained objects but just based on the `==` operator?

scferrada:
I couldn't reply directly to Olaf's comment. The `equals` method of `HashSet` is implemented in `AbstractSet`, which checks that both objects have the same size and that one `containsAll` the elements of the other. The `containsAll` method calls for each element the `contains` method, which in turn checks for `equals`. So, yep, the `equals` method in `TriplePatternImpl` should be called there.

JulietteVV:
> I couldn't reply directly to Olaf's comment. The `equals` method of `HashSet` is implemented in `AbstractSet`, which checks that both objects have the same size and that one `containsAll` the elements of the other. The `containsAll` method calls for each element the `contains` method, which in turn checks for `equals`. So, yep, the `equals` method in `TriplePatternImpl` should be called there.

I have put print in all equals function I can find, TriplePatternImpl included. I have tried multiple different ways to compare the unions and the only way that I have found, which works is to use the contains all function, which as Olaf pointed out would not be correct. When I call the equals function between two sets there, nothing is printed from the other equals functions (BGP, TriplePattern etc) and the results are wrong. 

I don't really know what else to try, at this point i've tried all solutions I can think of multiple times.

scferrada:
So I see two options. If order of subpatterns is not relevant (though I think Olaf is saying it is), you can still use `containsAll` to check, as long as you also check that both sets have the same size.

But, since order seems to be relevant, I think you should simply use list equality, and in the test, instead of using `assertEquals`, you should `assertTrue` that they have the same size and contain the same elements.

Cheers.

JulietteVV:
> Hi again! I've just talked with Olaf, and this is what needs to be done:
> 
> * We need to consider order in the `equals` method of `SPARQLUnionPatternImpl`. This is because we have the method `getSubPattern(int i)`, and it would not be correct for two objects that are "equal" to have different behaviours.
> * This means that in the test, we cannot use `assertEquals`, because, for the purposes of query rewritting, the order of the subpatterns is not really relevant, thus comparing equality is not necessary
> * In the test, you need to instead add two `assertTrue` calls, the first checking that both expected and actual `SPARQLUnionPattern`s have the same number of subpatterns, and that those subpatterns are the same in any order (or, alternatively said, all the patterns of one are contained in the other).
> 
> Please let me know if you have any doubts or if I didn't make myself clear enough. Sorry again for making you change things around so much, but this should be the final attempt.

I think this last push should correspond to what you wrote! I ended up flatening out the nestled union as you suggested earlier as well.

scferrada (changes requested):
Just some details here to fix. Please also just delete all code that is commented and you no longer need.

Thanks!

src/main/java/se/liu/ida/hefquin/engine/query/impl/BGPImpl.java
```diff
@@ -30,12 +34,19 @@ public boolean equals( final Object o ) {
...
+		
+		for(TriplePattern i : tps) {
+			tpsTriple.add(i.asJenaTriple());
+		}
+		
+		for(TriplePattern j : otps) {
+			otpsTriple.add(j.asJenaTriple());
+		}
+		
+		return tpsTriple.equals(otpsTriple);
```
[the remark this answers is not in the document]
JulietteVV: Making this change the bgps are not considered equal even if they contain the same triple, I have tried different method but I only got it to be equal if the jena triples are added

scferrada:
Hi all, it seems that the `equals` method on `TriplePatternImpl` was not correct, so I changed it, as well as the one on `BGPImpl` and it now doesn't need to create any auxiliary list. I think we're ready to merge.

scferrada (approved):

JulietteVV:
> Hi all, it seems that the `equals` method on `TriplePatternImpl` was not correct, so I changed it, as well as the one on `BGPImpl` and it now doesn't need to create any auxiliary list. I think we're ready to merge.

Okay, thank you for your help!

hartig (changes requested):
Sorry to be a pain in the neck but there are four major things that need to be changed here.

First, I see that you are creating literals almost everywhere. These should be URIs instead. Real RDF triples cannot have literals in the subject position, and real OWL axioms are not about literals.

Second, do not reuse the same `Node` objects for both the RDF representation of the vocabulary mapping and the triple patterns to be checked. For instance, in the helper function `CreateTestTriples()` you have `Node` objects such as `s1Res` that you use for the triples in the vocabulary mapping and also for the expected triples. Don't do that. Create separate `Node` objects even if they represent the same URI. The reason why that is important is because, when the `VocabularyMapping` is actually used by the engine, then the triple patterns to which the mapping is applied are also created separately from the RDF description of the mapping; hence, also there, the triple patterns will have different `Node` objects.

Third, I notice that none of the triple patterns considered in these tests contains variables. There should also be some tests with triple patterns that have variables.

Fourth, your RDF representation of owl:unionOf and owl:intersectionOf is incorrect. As an example, consider the following snippet of Turtle (prefix declarations omitted):
```
ex:Person owl:unionOf ( ex:Child ex:Adult ) .
```
Notice that this is not the same as the following:
```
ex:Person owl:unionOf ex:Child .
ex:Person owl:unionOf ex:Adult .
```
The parenthesis in the first snippet is a Turtle-specific shorthand notation for RDF collections. See: https://www.w3.org/TR/turtle/#collections  
So, the line of Turtle in the first snippet actually represents five RDF triples! That is, without using this shorthand notation, the first snippet expands to the following:
```
ex:Person  owl:unionOf  _:b1 .
_:b1  rdf:first  ex:Child .
_:b1  rdf:rest  _:b2 .
_:b2  rdf:first  ex:Adult .
_:b2  rdf:rest  rdf:nil .
```
Sorry I didn't see that earlier!!

JulietteVV:
Hi! I think I made a mistake while pushing my latest changes. I had forgotten to pull the lastest changes, so I commited the changes, pulled and then pushed. But now the changes I made won't show up in the PR. Is there anyway to fix this?

hartig:
Juliette, regarding your comment that your changes don't show up in the PR, there are three new commits from you here. Is this what you wanted to push or do you still have something else?

JulietteVV:
@hartig The changes you mentioned should be made now!

hartig:
@JulietteVV I am looking at this PR again now. I have to say that I find the unit tests very hard to understand. It is just not easily visible what exactly the mapping is, and what exactly the expected output is. I tried to change that by writing the respective sets of RDF triples in Turtle rather than constructing them programmatically. See my recent commit. Unfortunately, now one of the tests fails. Can you please take a look why that is (first, please double check that I didn't make a mistake when creating my Turtle representation of your code).

One thing that I observe already after having created this Turtle representation is that something seems to be missing in the expected output in `TranslateTriplePatternTest`. Given that the input triple pattern is the RDF triple (ex:s1, rdf:type, ex:o1) and that the mapping says (ex:s1, owl:sameAs, ex:s2), I would expect the output triple patterns to include the RDF triple (ex:s2, rdf:type, ex:o1) but this triple is not among the ones that you create as `expectedResult`. Why is that?

JulietteVV:
> @JulietteVV I am looking at this PR again now. I have to say that I find the unit tests very hard to understand. It is just not easily visible what exactly the mapping is, and what exactly the expected output is. I tried to change that by writing the respective sets of RDF triples in Turtle rather than constructing them programmatically. See my recent commit. Unfortunately, now one of the tests fails. Can you please take a look why that is (first, please double check that I didn't make a mistake when creating my Turtle representation of your code).
> 
> One thing that I observe already after having created this Turtle representation is that something seems to be missing in the expected output in `TranslateTriplePatternTest`. Given that the input triple pattern is the RDF triple (ex:s1, rdf:type, ex:o1) and that the mapping says (ex:s1, owl:sameAs, ex:s2), I would expect the output triple patterns to include the RDF triple (ex:s2, rdf:type, ex:o1) but this triple is not among the ones that you create as `expectedResult`. Why is that?

I will take a look and see! I will put in some comments to try to make it clearer. I understand that it's a little bit hard to understand what is happening and what is being tested. 

s2 is part of the expected results in both cases in TranslateTriplePatternTest but since the predicate is also translated the expected results contain (ex:s2 rdfs:subtypeOf ex:o2). I will double check if I have missed anything, but I think the expected results should be correct.

hartig:
Thanks for fixing my mistake in the Turtle representation!

Now I wonder: Should the mistake that I have made actually result in an error? In other words, should it matter whether the vocab.mapping contains a triple (ex:globalID, owl:sameAs, ex:localID) versus the triple (ex:localID, owl:sameAs, ex:globalID)  for ex:globalID to be rewritten into ex:localID?

Similarly, the example mapping in the unit test contains the triple (rdf:type, owl:inverseOf, ex:notType) which cause a triple pattern (?s, rdf:type, ?o) to be rewritten into (?o, ex:notType, ?s). Shouldn't the mapping triple (ex:notType, owl:inverseOf, rdf:type) have the same effect? I have tried in the unit test and it doesn't have the same effect; instead, the test fails. But, should it fail?

PR #165, opened 2022-03-17
Add an implementation of termination criteria

chengsijin0817:
Add an implementation of termination criteria, which examines the distance of average cost for a number of generations. The number of generations for checking depends on the number of lowest-cost plans in each generation.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
@@ -9,32 +9,37 @@ public class InitializeNrGeneration {
     public static int countNumOfOp( final LogicalPlan plan ) {
         final LogicalOperator lop = plan.getRootOperator();
 
-        int nrGeneration = 1;
         if ( lop instanceof LogicalOpMultiwayJoin) {
-            int count = plan.numberOfSubPlans();
-            nrGeneration = calcFactorial(count);
+            final int count = plan.numberOfSubPlans();
+
+            int nrGeneration = calcFactorial(count);
             for ( int i = 0; i < count; i++ ) {
                 nrGeneration = nrGeneration * countNumOfOp( plan.getSubPlan( i ) );
             }
+            return nrGeneration > 0 ? nrGeneration: Integer.MAX_VALUE;
         }
         else if ( lop instanceof LogicalOpMultiwayUnion) {
-            int count = plan.numberOfSubPlans();
+            final int count = plan.numberOfSubPlans();
+
+            int nrGeneration = 1;
             for ( int i = 0; i < count; i++ ) {
                 nrGeneration = nrGeneration * countNumOfOp( plan.getSubPlan( i ) );
             }
+            return nrGeneration > 0 ? nrGeneration: Integer.MAX_VALUE;
         }
         else if ( lop instanceof LogicalOpRequest) {
-            nrGeneration = 1;
+            return 1;
         }
         else {
             throw new IllegalArgumentException( lop.getClass().getName() );
         }
-        return nrGeneration;
     }
 
     protected static int calcFactorial( final int n ) {
         if ( n <= 1 )
             return 1;
+        else if ( n > 12 )
+            return Integer.MAX_VALUE;
         else
             return n * calcFactorial(n - 1);
     }
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
@@ -26,6 +26,9 @@ public class InitializeNrGeneration {
         else if ( lop instanceof LogicalOpRequest) {
             nrGeneration = 1;
         }
+        else {
+            throw new IllegalArgumentException( lop.getClass().getName() );
+        }
         return nrGeneration;
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
```diff
@@ -0,0 +1,39 @@
...
+import se.liu.ida.hefquin.engine.queryplan.LogicalOperator;
+import se.liu.ida.hefquin.engine.queryplan.LogicalPlan;
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.*;
+
+public class InitializeNrGeneration {
+
+    public static int countNumOfOp( final LogicalPlan plan ) {
+        final LogicalOperator lop = plan.getRootOperator();
+
+        int nrGeneration = 1;
```
hartig, line 12: Do not have this outside of the if-branches. Instead, move it inside each of the following branches and do a return at he end of each branch.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+        else if ( lop instanceof LogicalOpMultiwayUnion) {
+            int count = plan.numberOfSubPlans();
+            for ( int i = 0; i < count; i++ ) {
+                nrGeneration = nrGeneration * countNumOfOp( plan.getSubPlan( i ) );
+            }
+        }
+        else if ( lop instanceof LogicalOpRequest) {
+            nrGeneration = 1;
+        }
```
hartig, line 28: In such cases, always add an else-branch to make sure we get to know when we have some unexpected input.
```suggestion
        }
        else {
            throw new IllegalArgumentException( lop.getClass().getName() );
        }
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
```diff
@@ -0,0 +1,39 @@
...
+        return nrGeneration;
+    }
+
+    protected static int calcFactorial( final int n ) {
+        if ( n <= 1 )
+            return 1;
+        else
+            return n * calcFactorial(n - 1);
+    }
```
hartig, line 37: The maximum value for `int` is 2,147,483,647. The factorial of 12 is 479,001,600, which is smaller than 2,147,483,647, but the factorial of 13 (6,227,020,800) is already greater than that. Therefore, you should include an additional check here: if n>12, return Integer.MAX_VALUE

Similarly, in the calculations above (where you multiply), check if the result of the multiplication has become a negative value and, if so, replace this negative value by Integer.MAX_VALUE

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
```diff
@@ -0,0 +1,47 @@
...
+        final LogicalOperator lop = plan.getRootOperator();
+
+        if ( lop instanceof LogicalOpMultiwayJoin) {
+            final int count = plan.numberOfSubPlans();
+
+            int nrGeneration = calcFactorial(count);
+            for ( int i = 0; i < count; i++ ) {
+                nrGeneration = nrGeneration * countNumOfOp( plan.getSubPlan( i ) );
+            }
+            return nrGeneration > 0 ? nrGeneration: Integer.MAX_VALUE;
```
hartig, line 19: I am afraid that you need to do this check each time you multiply. Otherwise, you cannot catch cases in which the integer became negative and then became positive again.

Alternatively, and probably better, as soon as the value becomes negative **inside** the for-loop, return Integer.MAX_VALUE.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        else if ( lop instanceof LogicalOpMultiwayUnion) {
+            final int count = plan.numberOfSubPlans();
+
+            int nrGeneration = 1;
+            for ( int i = 0; i < count; i++ ) {
+                nrGeneration = nrGeneration * countNumOfOp( plan.getSubPlan( i ) );
+            }
+            return nrGeneration > 0 ? nrGeneration: Integer.MAX_VALUE;
```
hartig, line 28: Same comment here.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class InitializeNrGeneration {
+
+    public static int countNumOfOp( final LogicalPlan plan ) {
+        final LogicalOperator lop = plan.getRootOperator();
+
+        if ( lop instanceof LogicalOpMultiwayJoin) {
+            final int count = plan.numberOfSubPlans();
+
+            int nrGeneration = calcFactorial(count);
```
hartig, line 15: Sorry I didn't say it earlier, but I realize it only now: you can simplify the code by combining the cases of LogicalOpMultiwayJoin and LogicalOpMultiwayUnion into a single if-block. The only thing that you then need to be careful of is how you initialize `nrGeneration` -- it needs to be as follows:
```
int nrGeneration = ( lop instanceof LogicalOpMultiwayJoin ) ? calcFactorial(count) : 1;
```

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/optimizer/evolutionaryAlgorithm/InitializeNrGeneration.java
chengsijin0817: Thanks, fixed now.

hartig (approved):

PR #166, opened 2022-03-21
Filter Operators

SHSenpai:
>Add a corresponding logical operator and a corresponding physical operator. Begin with the logical one (because for implementing the physical one, you will need the logical one).
> 
> The class for the logical operator should be called `LogicalOpFilter`, it should go into the Java package `se.liu.ida.hefquin.engine.queryplan.logical.impl`, and it should implement the `UnaryLogicalOp` interface.
> 
> As an example of another such logical operator, take a look at `LogicalOpTPAdd`. You can essentially copy this class and replace its two member variables by a member variable of type `Expr` (i.e., exactly as you have it in the new executable operator in this PR here).
> 
> When you do this, you will notice that you also have to extend the `LogicalPlanVisitor` interface with an additional function for the new logical operator, and then you will also have to extend all classes that implement this interface (because they need to provide this additional function). The following list lists all these classes and what the added function should do:
> 
> * LogicalPlanCounter: do the same as for LogicalOpTPAdd
> * LogicalPlanPrinterAfterVisitor: do the same as for LogicalOpTPAdd
> * LogicalPlanPrinterBeforeVisitor: do the same as for LogicalOpTPAdd
> * LogicalPlanVisitorBase: do the same as for LogicalOpTPAdd

I've started with LogicalOpFilter, am I doing this correctly? My main pondering here is what to do with the toString() function. I made *a* toString(), but I don't know what utility it's supposed to serve.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
@@ -1,0 +1,51 @@
+package se.liu.ida.hefquin.engine.queryplan.logical.impl;
+
+import org.apache.jena.sparql.expr.Expr;
+
+import se.liu.ida.hefquin.engine.federation.FederationMember;
+import se.liu.ida.hefquin.engine.query.TriplePattern;
+import se.liu.ida.hefquin.engine.queryplan.logical.LogicalPlanVisitor;
+import se.liu.ida.hefquin.engine.queryplan.logical.UnaryLogicalOp;
+
+public class LogicalOpFilter implements UnaryLogicalOp
+{
+	protected final Expr filterExpression;
+
+	public LogicalOpFilter( final Expr filterExpression) {
+		assert filterExpression != null;
+
+		this.filterExpression = filterExpression;
+	}
+
+	@Override
+	public boolean equals( final Object o ) {
+		if ( ! (o instanceof LogicalOpTPAdd) )
+			return false;
+
+		final LogicalOpFilter oo = (LogicalOpFilter) o;
+		if ( oo == this )
+			return true;
+		else
+			return oo.filterExpression.equals(filterExpression); 
+	}
+
+	public Expr getFilterExpression() {
+		return filterExpression;
+	}
+
+	@Override
+	public void visit( final LogicalPlanVisitor visitor ) {
+		visitor.visit(this);
+	}
+
+	@Override
+	public String toString(){ // Unsure about what to do with this one.
+		final int codeOfExpr = filterExpression.toString().hashCode();
+
+		return "> filter" +
+				"[" + codeOfExpr + "]"+
+				" ( "
+				+ filterExpression.getVarName()
+				+ " )";
+	}
+}
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
@@ -2,8 +2,6 @@ package se.liu.ida.hefquin.engine.queryplan.logical.impl;
 
 import org.apache.jena.sparql.expr.Expr;
 
-import se.liu.ida.hefquin.engine.federation.FederationMember;
-import se.liu.ida.hefquin.engine.query.TriplePattern;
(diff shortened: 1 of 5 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
```diff
@@ -0,0 +1,49 @@
...
+
+	public LogicalOpFilter( final Expr filterExpression) {
+		assert filterExpression != null;
+
+		this.filterExpression = filterExpression;
+	}
+
+	@Override
+	public boolean equals( final Object o ) {
+		if ( ! (o instanceof LogicalOpTPAdd) )
```
hartig, line 20:
```suggestion
		if ( ! (o instanceof LogicalOpFilter) )

```

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
```diff
@@ -0,0 +1,49 @@
...
+
+	@Override
+	public String toString(){ // Unsure about what to do with this one.
+		final int codeOfExpr = filterExpression.toString().hashCode();
+
+		return "> filter" +
+				"[" + codeOfExpr + "]"+
+				" ( "
+				+ filterExpression.getVarName()
+				+ " )";
```
hartig, line 47: This function will be used for printing out a string representation of this operator; e.g., for debugging purposes. In particular, it will be used when printing out complete logical plans.

Here is what should do for this type of operator:

```suggestion
	public String toString() {
		return "> filter ( " + filterExpression.toString() + " )";
```

hartig (changes requested):
Looks good. Just a few very minor things.

hartig:
Once you have added my suggested changes, I will merge this PR. After that, you can start the next one with the purpose to also add a corresponding physical operator. More precisely, you need to create a new class called `PhysicalOpFilter`. This class should go into the Java package `se.liu.ida.hefquin.engine.queryplan.physical.impl` and it should implement the `UnaryPhysicalOpForLogicalOp` interface.

It should have a `LogicalOpFilter` as a member variable, assigned through the constructor.

The `createExecOp` function that you need to implement in this class should create and return an `ExecOpFilter` and it can ignore the `inputVars` argument.

The `getExpectedVariables` function that you need to implement in this class should check that the `inputVars` argument is an array of length 1 (otherwise throw an IllegalArgumentException) and, then, it should simply return the one `ExpectedVariables` object contained in that array.

All the other functions that you need to implement should be clear from the context. If not, simply ask.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
```diff
@@ -0,0 +1,44 @@
...
+	public boolean equals( final Object o ) {
+		if ( ! (o instanceof LogicalOpFilter) )
+
+			return false;
+
+		final LogicalOpFilter oo = (LogicalOpFilter) o;
+		if ( oo == this )
+			return true;
+		else
+			return oo.filterExpression.equals(filterExpression);
```
hartig, line 28:
```suggestion
		if ( o == this ) return true;
		if ( ! (o instanceof LogicalOpFilter) ) return false;

		final LogicalOpFilter oo = (LogicalOpFilter) o;
		return oo.filterExpression.equals(filterExpression); 
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpFilter.java
```diff
@@ -0,0 +1,44 @@
...
+	@Override
+	public void visit( final LogicalPlanVisitor visitor ) {
+		visitor.visit(this);
+	}
+
+	@Override
+	public String toString() {
+		return "> filter ( " + filterExpression.toString() + " )";
+	}
+}
```
hartig, line 44:
```suggestion
}

```

hartig (approved):

hartig:
Thanks. I will merge this PR now...

hartig:
Next task: add the corresponding physical operator as described in my comment above: https://github.com/LiUSemWeb/HeFQUIN/pull/166#issuecomment-1074456661

PR #172, opened 2022-03-24
Function for translating solution mappings and simple test (assert no…

JulietteVV:
…t working)

Branched from main, so current translateTriple implementation and tests are not included.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
@@ -1,5 +1,7 @@
 package se.liu.ida.hefquin.engine.data;
 
+import java.util.Set;
+
 import se.liu.ida.hefquin.engine.query.SPARQLGraphPattern;
 import se.liu.ida.hefquin.engine.query.TriplePattern;
 
@@ -16,4 +18,6 @@ public interface VocabularyMapping
 	 * pattern itself.
 	 */
 	SPARQLGraphPattern translateTriplePattern( TriplePattern tp );
+	
+	Set<SolutionMapping> translateSolutionMapping(SolutionMapping sm);
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
@@ -19,5 +19,5 @@ public interface VocabularyMapping
 	 */
 	SPARQLGraphPattern translateTriplePattern( TriplePattern tp );
 	
-	Set<SolutionMapping> translateSolutionMapping(SolutionMapping sm);
+	Set<SolutionMapping> translateSolutionMapping( final SolutionMapping sm );
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
@@ -19,5 +19,15 @@ public interface VocabularyMapping
 	 */
 	SPARQLGraphPattern translateTriplePattern( TriplePattern tp );
 	
+	/**
+	 * Applies this vocabulary mapping to the given Solution Mapping expressed
+	 * in the local vocabulary and returns returns a resulting set of
+	 * Solution Mappings expressed in the global vocabulary. 
+	 * If this mapping is not relevant for the given Solution Mapping (i.e.,
+	 * applying the mapping to the given Solution Mapping does not have any
+	 * effect), then the result of this function is a set containing only the
+	 * given Solution Mapping.
+	 */
 	Set<SolutionMapping> translateSolutionMapping( final SolutionMapping sm );
+
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
@@ -20,14 +20,14 @@ public interface VocabularyMapping
 	SPARQLGraphPattern translateTriplePattern( TriplePattern tp );
 	
 	/**
-	 * Applies this vocabulary mapping to the given Solution Mapping expressed
-	 * in the local vocabulary and returns returns a resulting set of
-	 * Solution Mappings expressed in the global vocabulary. 
-	 * If this mapping is not relevant for the given Solution Mapping (i.e.,
-	 * applying the mapping to the given Solution Mapping does not have any
+	 * Applies this vocabulary mapping to the given solution mapping expressed
+	 * in the local vocabulary and returns a resulting set of solution mappings
+	 * expressed in the global vocabulary. 
+	 * If this mapping is not relevant for the given solution mapping (i.e.,
+	 * applying the mapping to the given solution mapping does not have any
 	 * effect), then the result of this function is a set containing only the
-	 * given Solution Mapping.
+	 * given solution mapping.
 	 */
-	Set<SolutionMapping> translateSolutionMapping( final SolutionMapping sm );
+	Set<SolutionMapping> translateSolutionMapping( SolutionMapping sm );
 
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
@@ -353,8 +353,8 @@ public class VocabularyMappingImpl implements VocabularyMapping
 		Set<BindingBuilder> bbs = new HashSet<>();
 		BindingBuilder translation = BindingBuilder.create();
 		bbs.add(translation);
-		final Iterator<Var> i = sm.asJenaBinding().vars();
 		
+		final Iterator<Var> i = sm.asJenaBinding().vars();
 		while(i.hasNext()) {
 			final Var v = i.next();
 			final Node n = sm.asJenaBinding().get(v);
--- a/src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
@@ -358,7 +358,9 @@ public class VocabularyMappingImpl implements VocabularyMapping
 			final Node n = sm.asJenaBinding().get(v);
 			
 			if(!n.isURI()) {
-				continue;
+				for (final BindingBuilder j : bbs) {
+					j.add(v, n);
+				}
 			}
 			
 			final Set<Node> bindingTranslation = translateBinding(n);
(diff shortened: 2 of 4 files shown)


src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
```diff
@@ -16,4 +18,6 @@
 	 * pattern itself.
 	 */
 	SPARQLGraphPattern translateTriplePattern( TriplePattern tp );
+	
+	Set<SolutionMapping> translateSolutionMapping(SolutionMapping sm);
```
scferrada, line 22:
```suggestion
	Set<SolutionMapping> translateSolutionMapping( final SolutionMapping sm );
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -344,4 +348,69 @@ public Graph getVocabularyMappingAsGraph() {
 		return vocabularyMapping;
 	}
 
+	@Override
+	public Set<SolutionMapping> translateSolutionMapping(SolutionMapping sm) {
```
scferrada, line 352:
```suggestion
	public Set<SolutionMapping> translateSolutionMapping( final SolutionMapping sm ) {		
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -344,4 +348,69 @@ public Graph getVocabularyMappingAsGraph() {
...
+	@Override
+	public Set<SolutionMapping> translateSolutionMapping(SolutionMapping sm) {		
+		Set<BindingBuilder> bbs = new HashSet<>();
+		BindingBuilder translation = BindingBuilder.create();
+		bbs.add(translation);
+		final Iterator<Var> i = sm.asJenaBinding().vars();
+		
+		while(i.hasNext()) {
+			Var v = i.next();
+			Node n = sm.asJenaBinding().get(v);
```
scferrada, line 360:
```suggestion
			final Var v = i.next();
			final Node n = sm.asJenaBinding().get(v);
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -344,4 +348,69 @@ public Graph getVocabularyMappingAsGraph() {
...
+			Set<Node> bindingTranslation = translateBinding(n);
```
scferrada, line 361:
```suggestion
			final Set<Node> bindingTranslation = translateBinding(n);
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -101,5 +103,76 @@ public void TranslateTriplePatternTest() {
...
+		p = OWL.equivalentClass.asNode();
+		o = NodeFactory.createURI("m");
+		testSet.add(new Triple(s, p ,o));
+		
+		final VocabularyMapping vm = new VocabularyMappingImpl(testSet);
+		
+		final BindingBuilder testBuilder = BindingBuilder.create();
+		Var testVar = Var.alloc("v");
+		Node testNode = NodeFactory.createURI("n");
+		testBuilder.add(testVar, testNode);
```
scferrada, line 149: Instead of creating this variables and reassigning them, I would prefer if you create separate vars for each `add` or simply avoid them and just do:
```suggestion
		testBuilder.add( Var.alloc("v"), NodeFactory.createURI("n"));
```

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -101,5 +103,76 @@ public void TranslateTriplePatternTest() {
...
+		n = NodeFactory.createURI("b");
+		second.add(Var.alloc("v"), n);
+		n = NodeFactory.createURI("c");
+		second.add(Var.alloc("w"), n);
+		expectedResults.add(new SolutionMappingImpl(second.build()));
+		
+		System.out.print(expectedResults.toString());
+		System.out.print(translation.toString());
+		
+		//assertEquals(expectedResults, translation);
```
scferrada, line 175: Do you know the cause this doesn't work?

scferrada (changes requested):
Hi Juliette, here's some small comments and a question. I didn't get into the translation details, but will do so shortly.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
JulietteVV: When trying I, had this issue with the SolutionMapping equals function not being called (no prints or changes in i mattered). I tried comparing sizes and doing containsAll but that didn't work either.

scferrada: I will check what is going on there. I will leave a new comment on the meantime.

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
```diff
@@ -101,5 +103,68 @@ public void TranslateTriplePatternTest() {
...
+		testSet.add(new Triple(s, p ,o));
+		
+		final VocabularyMapping vm = new VocabularyMappingImpl(testSet);
+		
+		final BindingBuilder testBuilder = BindingBuilder.create();
+		testBuilder.add(Var.alloc("v"), NodeFactory.createURI("n"));
+		testBuilder.add(Var.alloc("w"), NodeFactory.createURI("m"));
+		
+		final SolutionMapping testSm = new SolutionMappingImpl(testBuilder.build());		
+		Set<SolutionMapping> translation = vm.translateSolutionMapping(testSm);
```
scferrada, line 151: It would be best if the test for translating solution mappings does not rely on the code of vocabulary mapping. They should be tested independently. To do this, please create a set of solution mappings that use a local vocabulary, create a vocabulary mapping, and then translate the set of solution mappings to the global vocabulary.

scferrada (changes requested):

src/test/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingTest.java
JulietteVV: I don't really understand what you mean, should I do the translation, as is done in the function directly in the test rather than using the function in vocabulary mapping? Right now I just create a vocabulary mapping, create a solution mapping and then use the translate solution mapping function to translate it.

scferrada: Oh, I'm so sorry. My IDE was telling me that the `translateSolutionMapping` method didn't exist so I assumed you were calling the triple pattern translation function.

scferrada:
I added the `hashCode` method to `SolutionMappingImpl` and the test passes. It seems it is important to override that method so that Set equality performs correctly.

scferrada (approved):

hartig (changes requested):
There are some things that need to be changed--see the comments of my review.

I also wonder about something else: I don't see how your implementation of `translateSolutionMapping` handles RDF terms (i.e., `Node` objects) that are not affected by the vocabulary mapping. For instance, if there is a solution mapping that maps variable ?x to the IRI http://example.org/x, then I would expect the solution mapping(s) returned by `translateSolutionMapping` to also map ?x in the same way. Your current implementation does not seem to achieve that, does it?

Somehow related to the previous point, you may actual add the following optimization to your implementation: As the first thing within the helper function `translateBinding`, you check whether the given `Node` object represent an IRI. If it does not, then it cannot be affected by the vocab.mapping and, thus, there is no need to do try to apply the mapping in this case.

src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
```diff
@@ -16,4 +18,16 @@
...
+	/**
+	 * Applies this vocabulary mapping to the given Solution Mapping expressed
+	 * in the local vocabulary and returns returns a resulting set of
+	 * Solution Mappings expressed in the global vocabulary. 
+	 * If this mapping is not relevant for the given Solution Mapping (i.e.,
+	 * applying the mapping to the given Solution Mapping does not have any
+	 * effect), then the result of this function is a set containing only the
+	 * given Solution Mapping.
+	 */
+	Set<SolutionMapping> translateSolutionMapping( final SolutionMapping sm );
```
hartig, line 31:
```suggestion
	 * Applies this vocabulary mapping to the given solution mapping expressed
	 * in the local vocabulary and returns a resulting set of solution mappings
	 * expressed in the global vocabulary. 
	 * If this mapping is not relevant for the given solution mapping (i.e.,
	 * applying the mapping to the given solution mapping does not have any
	 * effect), then the result of this function is a set containing only the
	 * given solution mapping.
	 */
	Set<SolutionMapping> translateSolutionMapping( SolutionMapping sm );
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -344,4 +347,76 @@ public Graph getVocabularyMappingAsGraph() {
...
+						}
+						translationCopy.add(v, j);
+						bbsCopy.add(translationCopy);
+					}
+				}
+				
+				bbs = bbsCopy;
+			} else if (bindingTranslation.size() == 0) {
+				continue;
+			} else {
```
hartig, line 382: Same issue here. I think that simply doing `continue` here will resulting in losing the binding for the current variable.

Also here, please cover this case in the additional unit test.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -344,4 +348,70 @@ public Graph getVocabularyMappingAsGraph() {
...
+		final Set<SolutionMapping> results = new HashSet<>();
+		for (final BindingBuilder b : bbs) {
+			results.add(new SolutionMappingImpl(b.build()));
+		}
+		return results;
+	}
+	
+	protected Set<Node> translateBinding(Node n){
+		final Set<Node> results = new HashSet<>();
+		for (final Triple m : getMappings(Node.ANY, Node.ANY, n)){
```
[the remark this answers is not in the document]
hartig: Replying to myself: there cannot be such cases; more precisely, we rule them out per definition (see also: https://github.com/LiUSemWeb/HeFQUIN/pull/164#issuecomment-1083230802)

hartig (changes requested):

hartig (approved):
Looks good now. Thanks!

Issue #168, opened 2022-03-23
An error "thread limit exceeds replacement of blocked workers" may occur, when the number of candidates per generation is large (e.g. 100)


chengsijin0817:
Error message: 

> Failure:Processing the query operator using HeFQUIN failed.
org.apache.jena.query.QueryExecException: Processing the query operator using HeFQUIN failed.
	at se.liu.ida.hefquin.jenaintegration.sparql.engine.main.OpExecutorHeFQUIN$MainQueryIterator.nextStage(OpExecutorHeFQUIN.java:164)
	at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.makeNextStage(QueryIterRepeatApply.java:100)
	at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.hasNextBinding(QueryIterRepeatApply.java:60)
	at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:114)
	at org.apache.jena.sparql.engine.iterator.QueryIterConvert.hasNextBinding(QueryIterConvert.java:58)
	at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:114)
	at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:38)
	at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:114)
	at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:38)
	at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:114)
	at org.apache.jena.sparql.engine.ResultSetStream.hasNext(ResultSetStream.java:64)
	at org.apache.jena.sparql.engine.ResultSetCheckCondition.hasNext(ResultSetCheckCondition.java:55)
	at org.apache.jena.sparql.resultset.ResultSetApply.apply(ResultSetApply.java:38)
	at org.apache.jena.riot.resultset.rw.ResultSetWriterXML.write(ResultSetWriterXML.java:82)
	at org.apache.jena.riot.resultset.rw.ResultsWriter.write(ResultsWriter.java:156)
[...]

PR #174, opened 2022-03-29
Address issue #168

chengsijin0817:
Extend the implementation of CostEstimationUtils.getEstimates and CostEstimationUtils.getEstimates: 
instead of creating CompletableFutures for all given plans, takes plans in batches (temporarily set the batch size as 50).


hartig (approved):
Looks good. Thanks!

PR #177, opened 2022-04-05
ExecOpRequestTPFatTPFServerWithTranslation first push

JulietteVV:


src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,86 @@
...
+
+	@Override
+	protected TPFResponse performPageRequest( final TPFRequest req,
+	                                          final FederationAccessManager fedAccessMgr )
+			throws FederationAccessException
+	{
+		SPARQLGraphPattern translation = fm.getVocabularyMapping().translateTriplePattern(req.getQueryPattern());
+		if(translation instanceof TriplePattern) {
+			TPFResponse res = FederationAccessUtils.performRequest(fedAccessMgr, new TPFRequestImpl((TriplePattern) translation), fm);
+			//How do I get solutionMapping from res?
```
scferrada, line 37: You can use the `getPayload()` method that returns a `Iterable<Triple>` object that allows you the triples that match the triple pattern, which is the kind of response of the TPF Server. You may need to manually create the Solution Mappings.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,86 @@
...
+			return res;
+		}
+		
+		//How do I create a union of all results?
+		else if(translation instanceof SPARQLUnionPattern) {
+			for (final SPARQLGraphPattern i : ((SPARQLUnionPattern) translation).getSubPatterns()) {
+				if (i instanceof TriplePattern) {
+					FederationAccessUtils.performRequest(fedAccessMgr, new TPFRequestImpl((TriplePattern) i), fm);
+				}
+				//TODO: Add all other option of instance
```
scferrada, line 48: Beware that the TPF Server can only handle TriplePattern Requests, so you need to "manually" perform any joins or unions required by the translation.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,86 @@
...
+			//Return intersection
+		}
+		
+		else {
+			throw new FederationAccessException(translation.toString(), req, fm);
+		}
+		
+		return null;
+	}
+}
```
scferrada, line 86: add empty newline at the end of the file

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	@Override
+	protected TPFResponse performPageRequest( final TPFRequest req,
+	                                          final FederationAccessManager fedAccessMgr )
+			throws FederationAccessException
+	{
+		SPARQLGraphPattern translation = fm.getVocabularyMapping().translateTriplePattern(req.getQueryPattern());
+		if(translation instanceof TriplePattern) {
+			TPFResponse res = FederationAccessUtils.performRequest(fedAccessMgr, new TPFRequestImpl((TriplePattern) translation), fm);
```
scferrada, line 36:
```suggestion
			final TPFResponse res = FederationAccessUtils.performRequest(fedAccessMgr, new TPFRequestImpl((TriplePattern) translation), fm);
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,86 @@
...
+			//How do I get solutionMapping from res?
+			//Return translation of res
+			return res;
+		}
+		
+		//How do I create a union of all results?
```
scferrada, line 42: You can create a `Set` and store the results there. It will take care of the duplicates as well.

scferrada (commented):
I added some responses to your queries and some observations.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,152 @@
...
+			final List<Triple> resTranslation = new ArrayList<>();
+			for(final Triple i : res.getPayload()) {
+				BindingBuilder bb = BindingBuilder.create();
+				final Var s = Var.alloc("s");
+				final Var p = Var.alloc("p");
+				final Var o = Var.alloc("o");
+				bb.add(s, i.asJenaTriple().getSubject());
+				bb.add(p, i.asJenaTriple().getPredicate());
+				bb.add(o, i.asJenaTriple().getObject());
+				SolutionMapping sm = new SolutionMappingImpl(bb.build());
```
scferrada, line 61: In this part, I think you need to use the variables that are present in the triple pattern from `req.getQueryPattern()`, instead of introducing your own variables. For instance, if the triple pattern is `(?m, ex:name, "Louise")` you should bind  `?m` to the subjects of the triples that you receive.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,152 @@
...
+				for (final SolutionMapping j : fm.getVocabularyMapping().translateSolutionMapping(sm)) {
+					Binding b = j.asJenaBinding();
+					resTranslation.add(new TripleImpl(b.get(s), b.get(p), b.get(o)));
+				}
+			}
+			final List<Triple> allTriples = new ArrayList<>();
+			res.getMetadata().forEach(allTriples::add);
```
scferrada, line 68: Why are you accessing the metadata here?

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,152 @@
...
+				}
+				//TODO: Add all other option of instance
+				else {
+					throw new FederationAccessException(i.toString(), req, fm);
+				}
+			}
+			//Return intersection of all results
+		}
+		
+		else if(reqTranslation instanceof BGP) {
```
scferrada, line 117: Here, I'm not sure if we defined how the mapping-based translation works over a BGP. Anyhow, the result of translating a BGP should be a join between all the translation results of each triple pattern in the BGP.

scferrada (changes requested):
Here are further comments.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,236 @@
...
+						partialRes.add(j);
+					}
+				}
+				bgpTranslation = partialRes;
+			}
+		}
+		return bgpTranslation;
+	}
+	
+	protected Set<Triple> translateResultTriple(final Triple t, final TPFRequest q){
```
scferrada, line 177: Are you using the `TPFRequest` for something else than getting the triple pattern? If not, please consider passing the triple pattern as parameter instead.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,236 @@
...
+			} else {
+				throw new FederationAccessException(i.toString(), req, fm);
+			}
+		}
+		return groupTranslation;
+	}
+	
+	protected Set<Triple> handleBGP(final BGP bgp, final FederationAccessManager fedAccessMgr) throws FederationAccessException{
+		Set<Triple> bgpTranslation = null;
+		for(final TriplePattern i : bgp.getTriplePatterns()) {
```
scferrada, line 161: I don't think that this is correct, because you are not performing a join between the results of evaluating each triple pattern. Please use/create a Join executable operator, and feed it with the results of evaluating the translation of each triple pattern in the BGP.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,236 @@
...
+				unionTranslation.addAll(handleBGP(((BGP) i), fedAccessMgr));
+			} else {
+				throw new FederationAccessException(i.toString(), req, fm);
+			}
+		}
+		return unionTranslation;
+	}
+	
+	protected Set<Triple> handleGroupPattern(final SPARQLGroupPattern gp, final FederationAccessManager fedAccessMgr) throws FederationAccessException{
+		Set<Triple> groupTranslation = null;
```
scferrada, line 101: Same here, I think that we need to perform a join between the results of evaluating each part of the Group Pattern, instead of performing a union.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,236 @@
...
+		final List<Triple> resList = new ArrayList<>();
+		if(reqTranslation instanceof TriplePattern) {
+			resList.addAll(handleTriplePattern((TriplePattern) reqTranslation, fedAccessMgr));
+		}
+		
+		else if(reqTranslation instanceof SPARQLUnionPattern) {
+			resList.addAll(handleUnionPattern(((SPARQLUnionPattern) reqTranslation), fedAccessMgr));
+		}
+		
+		else if(reqTranslation instanceof SPARQLGroupPattern) {
```
scferrada, line 57: I don't think that your translation method works for arbitrary SPARQL queries, right? So, I'm not sure that this case should be supported here. Notice that if the translation of a triple pattern produces a SPARQLGroupPattern, we have to support that, but we do not know how to translate a SPARQLGroupPattern (right?)

scferrada (changes requested):
Here are some more comments, mostly about performing joins (intersections) instead of unions

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
JulietteVV: SPARQLGroupPatterns are used to create intersections when all memebers of the intersection aren't triplepatterns (so BGP can't be used). So there could be a SPARQLGroupPattern in the translation

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
JulietteVV: The triplepatterns which are part of the result are only added if they where already in the solution. Only the first time a request is performed all the solutions are added to the results. Wouldn't this become a join?

scferrada: I don't think so, because we need that the resulting Triples are solutions for ALL the triple patterns in the BGP, so by adding all the triples matching the first triple pattern, you may be adding extra solutions (which would change, if the order of the triple patterns of the BGP changes).

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,215 @@
...
+		return groupTranslation;
+	}
+	
+	protected Set<Triple> handleBGP(final BGP bgp, final FederationAccessManager fedAccessMgr) throws FederationAccessException{
+		List<Set<Triple>> partialTranslations = new ArrayList<>();
+		for(final TriplePattern i : bgp.getTriplePatterns()) {
+			partialTranslations.add(handleTriplePattern(i, fedAccessMgr));
+		}
+		Set<Triple> bgpTranslation = new HashSet<>();
+		for(final Triple j : partialTranslations.get(0)) {
```
scferrada, line 139: I have been thinking a bit about this and I have come up with the following idea, that uses Set functionality and can make this code a bit easier to follow:
First, initialize a set with the resulting triples of evaluating the first triple pattern (e.g. `final Set<Triple> res`).
For each other triple pattern in the bgp, use the function `res.retainAll(otherSet)` that only keeps elements in the intersection.
At the end you'll have the intersection of all the sets of triples.

scferrada (changes requested):
I left a comment for improving the code readability, and probably its performance.

We also need some testing. You can draw inspiration from class `ExecOpRequestTPFatTPFServerTest `

JulietteVV:
Hi! I made a first verison of the test as I thought it would work. But it seams like no matter how I change the query or the vocabulary mapping the test still passes. So I think I have missunderstood how it works. Only changing the data in the execution context seams to change the outcome of the test. I am not sure what I have missed?

scferrada (changes requested):
Thanks Juliette, the joins look much better now! As for the tests, I may have found something, so please check.

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslationTest.java
```diff
@@ -37,50 +41,51 @@ public class ExecOpRequestTPFatTPFServerWithTranslationTest extends ExecOpTestBa
...
-		final Node s = NodeFactory.createURI("http://example.org/a");
+		final Node a = NodeFactory.createURI("http://example.org/a");
 		final Node p = NodeFactory.createURI("http://example.org/p");
 		final Var v = Var.alloc("v");
-		final TriplePattern tp = new TriplePatternImpl(s,p,v);
+		final TriplePattern tp = new TriplePatternImpl(a,p,v);
+		
+		final Graph g = GraphFactory.createDefaultGraph();
+
+		//TODO: add our data
```
scferrada, line 51: Is this TODO still relevant?

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslationTest.java
```diff
@@ -37,50 +41,51 @@ public class ExecOpRequestTPFatTPFServerWithTranslationTest extends ExecOpTestBa
...
-		assertTrue( b2.contains(v) );
-		assertEquals( "http://example.org/o2", b2.get(v).getURI() );
-
-		assertFalse( it.hasNext() );
+		final Set<SolutionMapping> results = new HashSet<>();
+		while (it.hasNext()) {
+			results.add(it.next());
+		}
+		
+		assertTrue(expectedResults.equals(results));
```
scferrada, line 82: This can be simplified:
```suggestion
		assertEquals(expectedResults, results);
```

scferrada (changes requested):
Thanks Juliette, there's just two small changes.

scferrada:
Thanks Juliette, we can wait now for @hartig 's final comments. On the meantime, you may proceed in PR #180 by creating the abstract superclass from which `ExecOpRequestTPFatTPFServerWithTranslation` and `ExecOpRequestTPFatBRTPFServerWithTranslation` will extend.

hartig (changes requested):
Unfortunately, there are several smaller but also two bigger issues with this PR.

One of the bigger issues is that the handling of BGPs and of group graph patterns is incorrect. See my comment within the code.

The other bigger issue is that I am not entirely convinced of the way the result is translated back from the local to the global vocab (i.e., this idea of translating the retrieved triples, as implemented by the helper function `translateResultTriple`). First of all, it is inefficient to produce `SolutionMapping` objects for all of these triples and then, after rewriting these solution mappings, creating new triples again. More importantly, however, I am not convinced that this is guaranteed to be correct for all cases (and there is no test coverage to check that it does). What I was thinking how this part will be implemented is that the solution mappings produced by this operator will be intercepted and rewritten before they are sent to the output sink. This can be done either by overriding the method `consumeMatchingTriples` (see the base class `ExecOpGenericRequestWithTPFPaging`) or--and even better--by overriding the method `convert` (see the base class `ExecOpGenericTriplePatternRequestWithTPF`).

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpRequest.java
```diff
@@ -46,6 +47,9 @@ public NullaryExecutableOp createExecOp( final ExpectedVariables ... inputVars )
 			return new ExecOpRequestSPARQL( (SPARQLRequest) req, (SPARQLEndpoint) fm );
 		}
 		else if ( fm instanceof TPFServer && req instanceof TriplePatternRequest ) {
+			if(fm.getVocabularyMapping() != null) {
+				return new ExecOpRequestTPFatTPFServerWithTranslation( (TriplePatternRequest) req, (TPFServer) fm );
+			}
```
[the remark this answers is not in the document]
hartig: I have taken care of this one: Commit https://github.com/LiUSemWeb/HeFQUIN/pull/177/commits/19ea3074bf9454e18b397e116b809b54678a90a1 adds a `PhysicalOpRequestWithTranslation`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,202 @@
...
+	}
+	
+	protected Set<Triple> handleBGP(final BGP bgp, final FederationAccessManager fedAccessMgr) throws FederationAccessException{
+		Set<Triple> bgpTranslation = null;
+		for(final TriplePattern i : bgp.getTriplePatterns()) {
+			if (bgpTranslation == null) {
+				bgpTranslation = new HashSet<>();
+				bgpTranslation.addAll(handleTriplePattern(i, fedAccessMgr));
+			} else {
+				bgpTranslation.retainAll(handleTriplePattern(i, fedAccessMgr));
```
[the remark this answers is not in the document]
JulietteVV: It was my understanding that this implementation did join the individual results with help of the retainAll method. I will take a look at it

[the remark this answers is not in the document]
hartig: Perhaps it is a good time to have a meeting again with the three of us and talk about all of this in detail?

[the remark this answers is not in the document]
scferrada: Good idea, I'll coordinate by mail.

PR #180, opened 2022-04-12
ExecOpRequestTPFatBRTPFServerWithTranslation.java

JulietteVV:
I did not see any difference for the performePageRequest function between ExecOpRequestTPFatTPFServer and ExecOpRequestTPFatBRTPFServer. Therefore I simply copied the code from ExecOpRequestTPFatTPFServerWithTranslation. I don't know if this is correct?


scferrada:
Yes, the thing is that since the request is a TPF request, we do not use the BR capabilities of the server. Given this, we need then to create an abstract superclass that contains all the repeated code, and make this class, and the one of the first PR extend from said abstract class. We need to merge the first PR before continuing with this.

JulietteVV:
I created an abstract superclass, I am not 100% sure this is the right way to do so, but I also created a test (copied from the previous pr) and it worked with this implementation.

JulietteVV:
I created an abstract superclass, I am not 100% sure this is the right way to do so, but I also created a test (copied from the previous pr) and it worked with this implementation.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFWithTranslation.java
```diff
@@ -0,0 +1,202 @@
...
+import se.liu.ida.hefquin.engine.federation.access.utils.FederationAccessUtils;
+import se.liu.ida.hefquin.engine.query.BGP;
+import se.liu.ida.hefquin.engine.query.SPARQLGraphPattern;
+import se.liu.ida.hefquin.engine.query.SPARQLGroupPattern;
+import se.liu.ida.hefquin.engine.query.SPARQLUnionPattern;
+import se.liu.ida.hefquin.engine.query.TriplePattern;
+
+public abstract class ExecOpRequestTPFWithTranslation<MemberType extends FederationMember> extends ExecOpGenericTriplePatternRequestWithTPF<MemberType>{
+
+	public ExecOpRequestTPFWithTranslation(TriplePatternRequest req, MemberType fm) {
```
scferrada, line 36:
```suggestion
	public ExecOpRequestTPFWithTranslation( final TriplePatternRequest req, final MemberType fm) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFWithTranslation.java
```diff
@@ -0,0 +1,202 @@
...
+				partialTranslation.addAll(handleGroupPattern(((SPARQLGroupPattern) i), fedAccessMgr));
+			} else if (i instanceof BGP) {
+				partialTranslation.addAll(handleBGP(((BGP) i), fedAccessMgr));
+			} else {
+				throw new FederationAccessException(i.toString(), req, fm);
+			}
+			
+			if (groupTranslation == null) {
+				groupTranslation = new HashSet<>();
+				groupTranslation.addAll(partialTranslation);
```
scferrada, line 120: Why not initialize the HashSet outside the for loop (this way, groupTranslation can be made final), and here check if isEmpty()

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFWithTranslation.java
```diff
@@ -0,0 +1,202 @@
...
+		}
+		return groupTranslation;
+	}
+	
+	protected Set<Triple> handleBGP(final BGP bgp, final FederationAccessManager fedAccessMgr) throws FederationAccessException{
+		Set<Triple> bgpTranslation = null;
+		for(final TriplePattern i : bgp.getTriplePatterns()) {
+			if (bgpTranslation == null) {
+				bgpTranslation = new HashSet<>();
+				bgpTranslation.addAll(handleTriplePattern(i, fedAccessMgr));
```
scferrada, line 133: Same comment here

src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFWithTranslationTest.java
```diff
@@ -0,0 +1,108 @@
...
+			@Override public boolean isExperimentRun() { return false; }
+		};
+	}
+	
+	public static VocabularyMapping createVocabularyMappingForTests() {
+		final Set<org.apache.jena.graph.Triple> mappingTriples = new HashSet<>();
+		//Equality
+		Node s = NodeFactory.createURI("http://example.org/a");
+		Node p = OWL.sameAs.asNode();
+		Node o  = NodeFactory.createURI("http://example.org/s");
```
scferrada, line 102:
```suggestion
		final Node s = NodeFactory.createURI("http://example.org/a");
		final Node p = OWL.sameAs.asNode();
		final Node o  = NodeFactory.createURI("http://example.org/s");
```

scferrada (changes requested):
Hi Juliette, here are some small comments. After this, it should be ready.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFWithTranslation.java
JulietteVV: I think I did it this way, in case the retain all at some point leads to the results being empty. Is my thought process correct or should I change it?

scferrada (changes requested):
Hi Juliette, the PR looks good, thanks! There's just one minor comment. Ask for Olaf's review after you address it.

JulietteVV:
@hartig This PR is ready to be reviewed once you have time!

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpRequestWithTranslation.java
```diff
@@ -55,10 +56,9 @@ public NullaryExecutableOp createExecOp( final ExpectedVariables ... inputVars )
 		else if ( fm instanceof TPFServer && req instanceof TriplePatternRequest ) {
 			return new ExecOpRequestTPFatTPFServerWithTranslation( (TriplePatternRequest) req, (TPFServer) fm );
 		}
-		// TODO: we need executable operators for the following cases
-//		else if ( fm instanceof BRTPFServer && req instanceof TriplePatternRequest ) {
-//			return new ExecOpRequestTPFatBRTPFServer( (TriplePatternRequest) req, (BRTPFServer) fm );
-//		}
+		else if ( fm instanceof BRTPFServer && req instanceof TriplePatternRequest ) {
+			return new ExecOpRequestTPFatBRTPFServerWithTranslation( (TriplePatternRequest) req, (BRTPFServer) fm );
+		}
```
hartig, line 61: I think that the line below is incorrect. It is the same condition as above but, instead, it should cover the case in which req is a BindingsRestrictedTriplePatternRequest.

hartig (changes requested):
Just one small thing.

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpRequestWithTranslation.java
JulietteVV: This should be fixed now!

hartig (approved):
Thanks for the final fix. I will merge this PR now.

PR #181, opened 2022-04-12
ExecOpRequestBRTPFWithTranslation

JulietteVV:
Again mostly the same as ExecOpRequestTPFatTPFServer but using BRTPFRequest

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestBRTPFWithTranslation.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestBRTPFWithTranslation.java
@@ -1,0 +1,208 @@
+package se.liu.ida.hefquin.engine.queryplan.executable.impl.ops;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.Binding;
+import org.apache.jena.sparql.engine.binding.BindingBuilder;
+
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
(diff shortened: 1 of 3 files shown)


JulietteVV:
I changed this operator to work like the RequestTPFatTPFServerWithTranslation, so that it calls the original operator within the translator one. I hope this was the right way to go about it!

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestBRTPFWithTranslation.java
```diff
@@ -0,0 +1,103 @@
...
+			throw new ExecOpExecutionException(p.toString(), this);
+		}
+	}
+	
+	protected Iterable<SolutionMapping> handleTriplePattern(final TriplePattern tp, final ExecutionContext execCxt) throws ExecOpExecutionException{
+		
+		final Set<SolutionMapping> translatedReqSM = new HashSet<>();
+		for (final SolutionMapping sm : req.getSolutionMappings()) {
+			translatedReqSM.addAll(fm.getVocabularyMapping().translateSolutionMappingFromGlobal(sm));
+		}
```
hartig, line 62: Can you please move this part to the `_execute` function (to avoid repeating this translation over and over again for every triple pattern), and then you pass `translatedReqSM` into the `handle...` functions.

hartig (changes requested):
Looks good! Just one thing that I would like you to change; see my in-code comment.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestBRTPFWithTranslation.java
JulietteVV: I've commited this change now!

hartig (approved):
Perfect. Thanks!

PR #182, opened 2022-04-29
ExecOpBindJoinSPARQLwithFILTERandTranslation

JulietteVV:


src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
```diff
@@ -22,12 +22,14 @@
...
+	 * expressed in the global vocabulary, if boolean fromLocal is true. Else the
+	 * vocabulary mapping is used to from the global vocabulary to the local
+	 * vocabulary.
 	 * If this mapping is not relevant for the given solution mapping (i.e.,
 	 * applying the mapping to the given solution mapping does not have any
 	 * effect), then the result of this function is a set containing only the
 	 * given solution mapping.
 	 */
-	Set<SolutionMapping> translateSolutionMapping( SolutionMapping sm );
+	Set<SolutionMapping> translateSolutionMapping( SolutionMapping sm , boolean fromLocal);
```
scferrada, line 33: I would prefer to have two methods here instead of choosing based on a flag, specially since in the implementation you have a call to two different methods based on this boolean value.

scferrada (changes requested):
The implementation looks quite good! There's just one comment, and we need some testing, both for the new global-to-local solution mapping translation, and for the executable operator.

hartig:
@JulietteVV @scferrada  I assume that this PR here is ready to be reviewed by me, right?
(Sorry, I have lost track a bit with all these PRs floating around ;-)

scferrada:
> @JulietteVV @scferrada I assume that this PR here is ready to be reviewed by me, right? (Sorry, I have lost track a bit with all these PRs floating around ;-)

Hi, it seems that I have to r

scferrada:
Sorry, my mouse went a bit crazy and I accidentally closed the PR. It seems that I had to review Juliette's latest changes, but it seems that is no longer needed, right?

PR #183, opened 2022-05-09
Adds nested-loop join between Solution Mapping Iterables

scferrada:


hartig (approved):
Good! I will merge it.

Thinking about it, instead of materializing the complete join result, it may actually be better to have an iterator that can enumerate the joined solution mappings. That iterator may then be wrapped in an iterable. Although I know better and should actually start answering emails from today now, I want to have a bit of fun and write such an iterator ;-)

PR #184, opened 2022-05-09
Adds JoiningIteratorForSolMaps and JoiningIterableForSolMaps

hartig:
This PR adds `JoiningIteratorForSolMaps` and `JoiningIterableForSolMaps` where the latter uses the former. These helper classes can be used to enumerate the result of a join of two collections of solution mappings. This is a streaming alternative to [SolutionMappingUtils.nestedLoopJoin](https://github.com/LiUSemWeb/HeFQUIN/blob/f21134e657c875705ee14d8269f0a38ff25dc656/src/main/java/se/liu/ida/hefquin/engine/data/utils/SolutionMappingUtils.java#L134) (see PR #183).

/cc @scferrada @JulietteVV

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/data/utils/JoiningIterableForSolMaps.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/utils/JoiningIterableForSolMaps.java
@@ -8,6 +8,9 @@ import se.liu.ida.hefquin.engine.data.SolutionMapping;
  * This is an iterable of solution mappings that enumerates the result of
  * joining two collections of solution mappings. The implementation uses
  * a {@link JoiningIteratorForSolMaps}.
+ *
+ * If you need the join result materialized, use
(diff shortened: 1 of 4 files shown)


hartig:
@JulietteVV you can use these helper classes as a cheap way of joining two sets/lists of solution mappings. Two this end, assuming these two sets/lists are given by the program variables `solmaps1` and `solmaps2`, you can simply do:
```java
final Iterable<SolutionMapping> joinResult = new JoiningIterableForSolMaps(solmaps1, solmaps2);
for ( final SolutionMapping sm : joinResult ) {
        // ... do whatever you want with each of the solution mappings in the join result
}
```

If you want to use an iterator instead, do:
```java
final Iterator<SolutionMapping> it = new JoiningIteratorForSolMaps(solmaps1, solmaps2);
while ( it.hasNext() ) {
        final SolutionMapping sm = it.next();
        // ... do whatever you want with each of the solution mappings in the join result
}
```

PR #185, opened 2022-05-10
SPARQL2GraphQLTranslator

WiktorKarlsson:
SPARQL2GraphQLTranslator with relating files


src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/RdfViewConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Verify that @param uri that corresponds to a property exist for @param className
+     */
+    public boolean containsPropertyURI(String className, String uri);
+
+    /**
+     * Verify that @param property of @param className links to a nested object
+     */
+    public boolean isObjectProperty(String className, String propertyName);
```
scferrada, line 36: does `containsProperty` refer to datatype properties? why do we need a different method for both types of properties?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLEntrypoint.java
```diff
@@ -0,0 +1,21 @@
...
+    public String getFieldName();
+
+    /**
+     * @return the parameter definitions for the entrypoint as a map
+     * consisting of parameternames mapped to parameter value types.
+     */
+    public TreeMap<String,String> getParameterDefinitions();
+
+    /**
+     * @return the sgp type of the entrypoint
```
scferrada, line 18: What is sgp?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLEntrypoint.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * @return the GraphQL query fieldname
+     */
+    public String getFieldName();
+
+    /**
+     * @return the parameter definitions for the entrypoint as a map
+     * consisting of parameternames mapped to parameter value types.
+     */
+    public TreeMap<String,String> getParameterDefinitions();
```
scferrada, line 15: Why do you need the concrete type `TreeMap`? Please consider using the interface `Map` as the return type.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/RdfViewConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public boolean containsClass(String className);
+
+    /**
+     * Verify that @param uri that correponds to a class exists
+     */
+    public boolean containsClassURI(String uri);
+
+    /**
+     * Verify that @param propertyName of @param className exist
```
scferrada, line 24:
```suggestion
     * Verifies that @param propertyName of @param className exists
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLEntrypoint.java
```diff
@@ -0,0 +1,21 @@
+package se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data;
+
+import java.util.TreeMap;
+
+public interface GraphQLEntrypoint {
```
scferrada, line 5: Add some javadoc that explains the purpose of this class as a whole

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/impl/GraphQLEntrypointType.java
```diff
@@ -0,0 +1,7 @@
+package se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data.impl;
+
+public enum GraphQLEntrypointType {
```
scferrada, line 3: Add javadoc that explains the purpose of the enum and its different constant values

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/impl/GraphQLFieldType.java
```diff
@@ -0,0 +1,6 @@
+package se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data.impl;
+
+public enum GraphQLFieldType {
```
scferrada, line 3: same here.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/RdfViewConfigurationImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+    // Maps classnames to a map which in turn maps property names to a GraphQLViewProperty object
+    protected final Map<String, Map<String,GraphQLViewProperty>> classToProperty;
+
+    // Maps classnames to a map which in turn maps a GraphQLEntrypointType to a GraphQLEntrypoint
+    protected final Map<String, Map<GraphQLEntrypointType,GraphQLEntrypoint>> classToEntrypoint;
+
+    // The URI prefixes used by the virtual RDF view
+    protected final String classPrefix = "c:";
+    protected final String propertyPrefix = "p:";
```
scferrada, line 23: Maybe, for the sake of extensibility, these prefixes should be customizable, and have `c:` and `p:` as the default values. Please add constructor parameters for these.
Also, consider using full URI notation. This means something like `http://example.org/c/` instead of just `c:`.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/RdfViewConfigurationImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        return scalarURIs;
+    }
+
+    @Override
+    public final String getPropertyValueType(String className, String propertyName) {
+        if(containsClass(className) && containsProperty(className, propertyName)){
+            return classToProperty.get(className).get(propertyName).getValueType();
+        }
+        return null;
```
scferrada, line 111: please state in the javadoc of the interface that this method returns `null` if the property is not found.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/RdfViewConfigurationImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    public String getClassPrefix() {
+        return classPrefix;
+    }
+
+    @Override
+    public final GraphQLEntrypoint getEntrypoint(String className, GraphQLEntrypointType type) {
+        if(containsClass(className) && classToEntrypoint.get(className).containsKey(type)){
+            return classToEntrypoint.get(className).get(type);
+        }
+        return null;
```
scferrada, line 136: same here

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,505 @@
...
+            else{
+                Set<TP> sgp = new HashSet<>();
+                sgp.add(wrappedTriplePattern);
+                subgraphPatterns.put(subject, sgp);
+            }
+            ++idCounter;
+        }
+
+        // Check for dependencies between the SGPs and create connectors
+        Map<Node,SgpNode> sgpNodes = new HashMap<>();
```
scferrada, line 56:
```suggestion
        final Map<Node,SgpNode> sgpNodes = new HashMap<>();
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,505 @@
...
+        public TP(final int id, final TriplePattern tp) {
+            this.id = id;
+            this.tp = tp;
+        }
+
+        public final int getId() {
+            return id;
+        }
+
+        public final TriplePattern getTp(){
```
scferrada, line 492: You are returning a `TriplePattern` object, not a `TP` object, so name the method accordingly.
```suggestion
        public final TriplePattern getTriplePattern(){
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,505 @@
...
+        }
+    }
+
+    /**
+     * Helper class representing the nodes used to check for cyclic bindings between SGPs
+     */
+    private class SgpNode {
+        public Map<Integer,SgpNode> adjacentNodes = new HashMap<>();
+        public boolean visited = false;
+        public boolean finished = false;
```
scferrada, line 503: add getters and setters, these attributes should not be directly modifiable.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLEntrypoint.java
WiktorKarlsson: The reason for why I changed to TreeMap is because without the guaranteed ordering of the keys in the keyset of the TreeMap, the GraphQLQueryImpl .toString() method might return different answers each time. The toString method in the GraphQLQueryImpl adds the parameters of the query through iterating over the keyset. By having a normal Map or HashMap the ordering of how these parameters are added can be different. This makes it difficult to create a testcase since when comparing the translated query to the expected "hard coded" query, the translated query might be different from time to time.

Perhaps I should change to the `SortedMap` interface?

scferrada: I have two follow-up comments/questions:
1. Is it completely necessary to have the `toString` method to be deterministic? I'm worried that we are adding computing time to keep sorted something that isn't 100% required
2. If it is necessary to keep the map sorted, yes, please use the interface `SortedMap` as a return type. Otherwise, just use `Map`.

PR #186, opened 2022-05-11
Fixed approach of VocabularyMappingImpl

JulietteVV:
Removed union and intersection as possibilities for Predicate translation and fixed the test accordingly


scferrada (commented):
Thanks, looks good to me. Let's wait for @hartig 's comments.

hartig (approved):

PR #187, opened 2022-05-13
GraphQL2RDFConfiguration and extra

WiktorKarlsson:
I removed the SPARQL2GraphQLTranslatorImpl and its tests to make this PR smaller. Every comment from the previous PR should be fixed in this latest commit. Since the translator implementation and its tests have dependencies to files in this PR, I'll create a new PR for those once this is merged.


src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/GraphQL2RDFConfigurationImpl.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        return containsProperty(className, propertyName);
+    }
+
+    @Override
+    public final GraphQLFieldType getPropertyFieldType(final String className, final String propertyName) {
+        if(containsClass(className) && containsProperty(className, propertyName)){
+            final GraphQLFieldType type = classToProperty.get(className).get(propertyName).getFieldType();
+            return type;
```
scferrada, line 90: Why do you need a variable here? Simply return the expression.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/CyclicFinder.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Set;
+
+import org.apache.jena.graph.Node;
+
+public class CyclicFinder {
+
+    /**
+     * Helper function used to call the DFS function for each SgpNode
+     */
+    public static final Set<Integer> determineCyclicConnectors(final Map<Node,SGPNode> sgpNodes){
```
scferrada, line 14: Why is this method `final`?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SGPNode.java
```diff
@@ -0,0 +1,37 @@
+package se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.utils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Helper class representing the nodes used to check for cyclic bindings between SGPs
+ */
+public class SGPNode {
+    private Map<Integer,SGPNode> adjacentNodes = new HashMap<>();
```
scferrada, line 10:
```suggestion
    private final Map<Integer,SGPNode> adjacentNodes = new HashMap<>();
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/query/impl/GraphQLQueryImpl.java
```diff
@@ -33,7 +33,8 @@ public String toString() {
         return query;
```
scferrada, line 33: I cannot leave the comment in the respective line, but please, use a `StringBuilder` object instead of concatenating Strings. Using `+=` for strings is very inefficient.

WiktorKarlsson:
> Hi Wiktor, thanks a lot, here are some more comments.
> 
> There's a more general matter that I want to discuss too. I think that there is no distinction between the Configuration and the Endpoint. In my opinion, an Endpoint should have a list of Entrypoints for the different classes, and thus be responsible to respond to the calls of `containsClass` and such. Similarly, the Configuration should simply be concerned with IRI matters, so for instance, the method `containsClassURI` should receive both the URI and the Endpoint: the configuration is in charge to check that the URI is valid for a class, extract the class name, and ask the Endpoint if it contains the class. Furthermore, instead of simply call the implementation of the `GraphQ2RDFConfiguration` as `GraphQ2RDFConfigurationImpl`, I suggest you call it `DefaultGraphQ2RDFConfiguration`, in case we want to add different configurations with different prefixes.
> 
> @hartig what do you think of this?

I have made fixes to the code parts that were commented on in this latest commit. I haven't moved parts from the GraphQL2RDFConfiguration to the GraphQLEndpoint yet though since you asked for @hartig opinion on this. My original idea for the configuration was that it would be created by another object sending an introspection query to the endpoint, parsing the result and creating a configuration based on that retrieved information.

scferrada:
> I have made fixes to the code parts that were commented on in this latest commit. I haven't moved parts from the GraphQL2RDFConfiguration to the GraphQLEndpoint yet though since you asked for @hartig opinion on this. My original idea for the configuration was that it would be created by another object sending an introspection query to the endpoint, parsing the result and creating a configuration based on that retrieved information.

I'll respond tho this here, before I go through the new commits. By your answer I'm more convinced now that GraphQLEndpoint should be the class that manages the methods I mentioned. I agree that there should be a utils method that can perform an introspection query to the endpoint and return an object that models such an endpoint (which is a GraphQLEndpoint and not a configuration), with the classes and other info obtained through the query. The configuration should simply provide functionalities to map class/property names to IRIs and vice versa, and also to check that a given URI is a valid class/property, and nothing more.

In light of this discussion, I'm confident that @hartig will agree with this. So, please, reorganize the code to better separate the endpoint from the configuration.

scferrada (commented):
Thanks for addressing my comments!.

hartig:
I agree with @scferrada

WiktorKarlsson:
@scferrada I have moved the necessary methods from the config to `GraphQLEndpointImpl`, which is an implementation of the `GraphQLEndpoint`. I was a bit unsure where the implementation of the endpoint was supposed to go so I added it to the graphqlwrapper/impl folder for now. I was also slightly unsure whether you meant if I should actually add the methods to the `GraphQLEndpoint` or the `GraphQLInterface`? If the latter is the case then I'll fix it quickly in a new commit.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/GraphQL2RDFConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @return the prefix used for classes
+     */
+    public String getClassPrefix();
+
+    /**
+     * @return the predefined namespace prefix for rdf
+     * https://www.w3.org/1999/02/22-rdf-syntax-ns#
+     */
+    public String getRDFPrefix();
```
scferrada, line 44: I'm still not sold on this method. Jena has a constant value with this URI, you can simply use `org.apache.jena.vocabulary.RDF.uri`, wherever you need this value. However, if you need to be able to customize the URI to indicate that an object belongs to a class (which in the default configuration is `rdf:type`) you can rename this method to `getClassMembershipURI` for instance, and make the default configuration return Jena's `RDF.type.getURI()`.

scferrada (changes requested):
Thanks Wiktor, this last commit addressed my comment as I had requested. There is just one comment here.

I assume that the utility method to set up a GraphQL endpoint through introspection will come in a different PR.

@hartig, I think this is ready for you to go through.

WiktorKarlsson:
> Thanks Wiktor, this last commit addressed my comment as I had requested. There is just one comment here.
> 
> I assume that the utility method to set up a GraphQL endpoint through introspection will come in a different PR.

Yes, I plan to add it in a different PR. I don't want to risk making this PR to large again. Hope that's okay.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/GraphQL2RDFConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+import se.liu.ida.hefquin.engine.federation.GraphQLEndpoint;
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data.impl.GraphQLFieldType;
+
+/**
+ * Used to represent and retrieve information about a virtual RDF view for a GraphQL endpoint
+ */
+public interface GraphQL2RDFConfiguration {
+
+    /**
+     * Verifies that @param uri that correponds to a class exists for @param endpoint
```
hartig, line 14: My confusion continues. Now there seem to be URIs that may exist for GraphQL endpoints. How? / "exists" in what sense? 

...and how is "verifies" meant? In the sense that objects of classes that implement this interface can perform such a check? If so, then it sounds more like a helper function to me and I wonder why such a helper function is part of something that is called a "configuration"?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/GraphQL2RDFConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @return a set of URI of all properties for @param className that links to a scalar value
+     * for @param endpoint
+     */
+    public Set<String> getPropertyURIs(final String className, final GraphQLFieldType fieldType, 
+        final GraphQLEndpoint endpoint);
+
+    /**
+     * @return the prefix used for properties
```
hartig, line 31: What are "properties" and what is a "prefix used for [such] properties"?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/GraphQL2RDFConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public String getPropertyPrefix();
+
+    /**
+     * @return the prefix used for classes
```
hartig, line 36: Similar question here.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLEntrypoint.java
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * @return the GraphQL query fieldname
+     */
+    public String getFieldName();
+
+    /**
+     * @return the parameter definitions for the entrypoint as a map
+     * consisting of parameter names mapped to parameter value types.
+     */
+    public Map<String,String> getParameterDefinitions();
```
hartig, line 19: Is this meant to be about the arguments defined for a corresponding field? If so, please use the term "argument" instead of "parameter" to be consistent with the terminology usually used in the context of GraphQL.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLEntrypoint.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * @return the GraphQL type the entrypoint represents
+     */
+    public String getType();
```
hartig, line 24: Is the type a string? Shouldn't there be a Java class (or interface) whose instances represents such types? (I mean, for other kinds of elements of GraphQL schemas you also have Java classes/interface; for instance, this very interface here.)

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLProperty.java
```diff
@@ -0,0 +1,24 @@
+package se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data;
+
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data.impl.GraphQLFieldType;
+
+/**
+ * Used to represent information about a field in a GraphQL object
+ */
+public interface GraphQLProperty {
```
hartig, line 8: If this is about fields, then it shouldn't be named "property"! That's highly confusing.

hartig (changes requested):
I'm sorry but I'm totally confused by all of this. See my detailed comments. Maybe it is just a matter of terminology but I really don't understand the purpose of many of the things here.

src/main/java/se/liu/ida/hefquin/engine/federation/GraphQLEndpoint.java
```diff
@@ -1,10 +1,56 @@
...
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data.impl.GraphQLEntrypointType;
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data.impl.GraphQLFieldType;
 
 public interface GraphQLEndpoint extends FederationMember
 {
 	@Override
 	GraphQLInterface getInterface();
 
+	/**
+     * Verifies that @param className exists
```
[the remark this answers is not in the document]
hartig: > What I actually mean with a "class" is [...] 

Thanks for the clarification!

> Do you think clarity of the files in the PR would improve if I changed methods like containsClass to containsGraphQLObjectType and reworked the JavaDoc comments to be more informative? 

Yes, I totally think so. Please do it.

hartig (approved):
Looks good now. Thanks for changing the names to make things more understandable.

PR #189, opened 2022-05-16
New PR ExecOpRequestTPFatTPFServerWithTranslation

JulietteVV:


src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,223 @@
...
+ * Implementation of an operator to request a (complete) TPF from a TPF server.
+ * This implementation handles pagination of the TPF; that is, it requests all
+ * the pages, one after another.
+ */
+public class ExecOpRequestTPFatTPFServerWithTranslation extends ExecOpGenericRequest<TriplePatternRequest, TPFServer>
+{
+	public ExecOpRequestTPFatTPFServerWithTranslation( final TriplePatternRequest req, final TPFServer fm ) {
+		super( req, fm );
+
+		assert fm.getVocabularyMapping() != null;
```
scferrada, line 38: maybe this assertion must be the first instruction of the constructor?

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	
+	@Override
+	protected void _execute(IntermediateResultElementSink sink, ExecutionContext execCxt)
```
scferrada, line 43:
```suggestion
	protected void _execute( final IntermediateResultElementSink sink, final ExecutionContext execCxt)
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,223 @@
...
+			throws ExecOpExecutionException {
+		
+		FederationAccessManager fedAccessMgr = execCxt.getFederationAccessMgr();
```
scferrada, line 46:
```suggestion
		final FederationAccessManager fedAccessMgr = execCxt.getFederationAccessMgr();
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,223 @@
...
+			catch ( final FederationAccessException e ) {
+				throw new ExecOpExecutionException("Issuing a page request caused an exception: " + e.toString(), this);
+			}
+			
+			// consume the matching triples retrieved via the page request
+			op.consumeMatchingTriples( currentPage.getPayload(), sink );
+
+		}	
+	
+		return sink.getMaterializedIntermediateResult();
```
scferrada, line 95: On the interest of having homogeneous return types for the handle methods, please directly return the iterator here, instead of calling the iterator method in _execute and in every other place you use handleTriplePattern.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,223 @@
...
+				bgpTranslation = handleTriplePattern((TriplePattern) i, fedAccessMgr).iterator();
+			} else {
+				bgpTranslation = new JoiningIteratorForSolMaps(getIterableFromIterator(bgpTranslation), handleTriplePattern((TriplePattern) i, fedAccessMgr));
+			}
+		}
+		return bgpTranslation;
+	}
+	
+	//Helper class
+	protected static class MaterializingIntermediateResultElementSinkWithTranslation implements ClosableIntermediateResultElementSink {
```
scferrada, line 180: Is this class copied from somewhere else? If so, would it be possible to make a sub-class that handles vocabulary mappings and overrides just the send method?

scferrada (changes requested):
Thanks Juliette, the code looks quite good. There are however some comments and questions.

A more general comment is that I'm not a fan of all the comparisons to `null` that happen in `handleUnionPattern` and `handleGroupPattern`. I will give it a think to see if I can come up with a way to get rid of those.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,223 @@
...
+
+	}
+	
+	protected Iterator<SolutionMapping> handleUnionPattern(final SPARQLUnionPattern up, final FederationAccessManager fedAccessMgr) throws ExecOpExecutionException {
+		Iterator<SolutionMapping> unionTranslation = null;
+		for(final SPARQLGraphPattern i : up.getSubPatterns()) {
+			if (i instanceof TriplePattern) {
+				if (unionTranslation == null) {
+					unionTranslation = handleTriplePattern(((TriplePattern) i), fedAccessMgr).iterator();
+				} else {
```
scferrada, line 105: I think that it would be best to iterate through the iterator of `up.getSubPatterns()`. That way, you can consume the first element first, and store it in `unionTranslation` and then iterate through the rest to create the union tree.

Furthermore, you can add a generic `handlePattern` method that evaluates all the `instanceof` conditions and calls the `handle...` method appropriate for the pattern type, which will simplify all the code of this class.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
scferrada: Ah, right, let's leave it like that then. Thanks!

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/MaterializingIntermediateResultElementSinkWithTranslation.java
```diff
@@ -0,0 +1,29 @@
...
+	protected final VocabularyMapping vocabularyMapping;
+	protected final List<SolutionMapping> l = new ArrayList<>();
+	private boolean closed = false;
+	
+	public MaterializingIntermediateResultElementSinkWithTranslation(final VocabularyMapping vm) {
+		this.vocabularyMapping = vm;
+	}
+	
+	@Override
+	public void send(SolutionMapping element) {
```
scferrada, line 20:
```suggestion
	public void send(final SolutionMapping element) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/MaterializingIntermediateResultElementSinkWithTranslation.java
```diff
@@ -0,0 +1,29 @@
...
+		if ( ! closed ) {
+			for (SolutionMapping sm : vocabularyMapping.translateSolutionMapping(element)) {
```
scferrada, line 22:
```suggestion
			for (final SolutionMapping sm : vocabularyMapping.translateSolutionMapping(element)) {
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,149 @@
...
+			op.consumeMatchingTriples( currentPage.getPayload(), sink );
+
+		}	
+	
+		return sink.getMaterializedIntermediateResult().iterator();
+
+	}
+	
+	protected Iterator<SolutionMapping> handleUnionPattern(final SPARQLUnionPattern up, final FederationAccessManager fedAccessMgr) throws ExecOpExecutionException {
+		Iterator<SPARQLGraphPattern> i = up.getSubPatterns().iterator();
```
scferrada, line 111:
```suggestion
		final Iterator<SPARQLGraphPattern> i = up.getSubPatterns().iterator();
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,149 @@
...
+
+	
+	@Override
+	protected void _execute( final IntermediateResultElementSink sink, final ExecutionContext execCxt)
+			throws ExecOpExecutionException {
+		
+		final FederationAccessManager fedAccessMgr = execCxt.getFederationAccessMgr();
+		final SPARQLGraphPattern reqTranslation = fm.getVocabularyMapping().translateTriplePattern(req.getQueryPattern());
+		Iterator<SolutionMapping> res;
+		if(reqTranslation instanceof TriplePattern) {
```
scferrada, line 46: Here you can also call `handlePattern` to simplify.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,149 @@
...
+		Iterator<SPARQLGraphPattern> i = up.getSubPatterns().iterator();
+		Iterator<SolutionMapping> unionTranslation = handlePattern(i.next(), fedAccessMgr);
+		while(i.hasNext()){
+			unionTranslation = new UnionIteratorForSolMaps(unionTranslation, handlePattern(i.next(), fedAccessMgr));
+		}
+		return unionTranslation;
+	}
+	
+	protected Iterator<SolutionMapping> handleGroupPattern(final SPARQLGroupPattern gp, final FederationAccessManager fedAccessMgr) throws ExecOpExecutionException {
+		Iterator<SPARQLGraphPattern> i = gp.getSubPatterns().iterator();
```
scferrada, line 120:
```suggestion
		final Iterator<SPARQLGraphPattern> i = gp.getSubPatterns().iterator();
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,149 @@
...
+			}
+		}
+		return bgpTranslation;
+	}
+
+	
+	//Source: https://www.geeksforgeeks.org/convert-iterator-to-iterable-in-java/
+	// Function to get the Spliterator
+    public static <T> Iterable<T>
+    getIterableFromIterator(Iterator<T> iterator)
```
scferrada, line 144: I'm not sure why we need this. I'm sure there is a version of the join and union iterators that Olaf made that receive iterators directly.

scferrada (changes requested):
Hi Juliette,

Just a couple of small things. After you address them, request a review from Olaf.

Thanks!

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
JulietteVV: the union iterator does handle either iterator or iterable inputs, but the join version requires iterables. I could make a constructor for JoiningIteratorForSolMaps which takes iterators? Or is it better to keep it this way?

scferrada: Hi, I just saw that problem and added a new constructor. We may need to change this later, but on the meantime, please use the new code.

JulietteVV:
@hartig I think this PR is ready to be reviewed, once you have time.

hartig (changes requested):
In addition to the comment in my review, @scferrada's recent change with  the constructor that takes iterators instead of iterables needs to be reverted. This constructor does not work because, as soon as the second input iterator has been consumed once, it cannot be consumed again.

My proposal to make possible that the `JoiningIteratorForSolMaps`/`JoiningIterableForSolMaps` can be constructed using iterables instead of iterators is to simply change the return type of all these `handle...` functions; make it `Iterable<SolutionMapping`. The base step of all this recursion is the function `handleTriplePattern` and that one can simply return `sink.getMaterializedIntermediateResult()` instead of `sink.getMaterializedIntermediateResult().iterator()`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,124 @@
...
+				currentPage = op.performPageRequest( pageRequest, fedAccessMgr );
+			}
+			catch ( final FederationAccessException e ) {
+				throw new ExecOpExecutionException("Issuing a page request caused an exception: " + e.toString(), this);
+			}
+			
+			// consume the matching triples retrieved via the page request
+			op.consumeMatchingTriples( currentPage.getPayload(), sink );
+
+		}
```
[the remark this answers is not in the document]
JulietteVV: I think this latest commit includes all the changes you mentioned, I did add the getIterableFromIterator function back in order to make the JoiningIteratorForSolMaps  and UnionIteratorForSolMaps iterables.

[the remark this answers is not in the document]
hartig: You don't need to make them iterables. Instead, there already exist corresponding iterables. They are in the same Java package as the iterators. Simply use these.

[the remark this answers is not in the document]
JulietteVV: Okay, thank you!

src/main/java/se/liu/ida/hefquin/engine/data/utils/JoiningIteratorForSolMaps.java
```diff
@@ -23,6 +23,11 @@ public class JoiningIteratorForSolMaps implements Iterator<SolutionMapping>
 	protected SolutionMapping currentInputElement;
 	protected SolutionMapping nextOutputElement;
 
+	public JoiningIteratorForSolMaps( final Iterator<SolutionMapping> input1,
+									  final Iterator<SolutionMapping> input2) {
+		this(()->input1, ()->input2);
+	}
+
```
hartig, line 30:
```suggestion
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/MaterializingIntermediateResultElementSinkWithTranslation.java
```diff
@@ -0,0 +1,29 @@
...
+		this.vocabularyMapping = vm;
+	}
+	
+	@Override
+	public void send(final SolutionMapping element) {
+		if ( ! closed ) {
+			for (final SolutionMapping sm : vocabularyMapping.translateSolutionMapping(element)) {
+				l.add(sm);
+			}
+		}
```
hartig, line 25:
```suggestion
public class MaterializingIntermediateResultElementSinkWithTranslation extends MaterializingIntermediateResultElementSink
{
	protected final VocabularyMapping vm;
	protected final List<SolutionMapping> l = new ArrayList<>();

	public MaterializingIntermediateResultElementSinkWithTranslation( final VocabularyMapping vm ) {
		assert vm != null;
		this.vm = vm;
	}
	
	@Override
	public void send( final SolutionMapping element ) {
		for ( final SolutionMapping sm : vm.translateSolutionMapping(element) ) {
			super.send(sm);
		}
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,101 @@
...
+			if (bgpTranslation == null) {
+				bgpTranslation = handleTriplePattern(i, execCxt);
+			} else {
+				bgpTranslation = new JoiningIterableForSolMaps(bgpTranslation, handleTriplePattern(i, execCxt));
+			}
+		}
+		return bgpTranslation;
+	}
+
+}
```
hartig, line 101:
```suggestion
}

```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpRequestTPFatTPFServerWithTranslation.java
```diff
@@ -0,0 +1,101 @@
...
+	
+	@Override
+	protected void _execute( final IntermediateResultElementSink sink, final ExecutionContext execCxt)
+			throws ExecOpExecutionException {
+		
+		final SPARQLGraphPattern reqTranslation = fm.getVocabularyMapping().translateTriplePattern(req.getQueryPattern());
+		final Iterator<SolutionMapping> res = handlePattern(reqTranslation, execCxt).iterator();	
+		while(res.hasNext()) {
+			sink.send(res.next());
+		}
```
hartig, line 43:
```suggestion
		for ( final SolutionMapping sm : handlePattern(reqTranslation, execCxt) ) {
			sink.send(sm);
		}	
```

hartig:
Looks good now. Thanks @JulietteVV !

PR #193, opened 2022-05-24
SPARQL2GraphQLTranslatorImpl and tests

WiktorKarlsson:
Implementation of the SPARQL2GraphQLTranslator.

The translated queries from the tests have also been verified to work against my local GraphQL endpoint.


src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+                    subjectSgpNode.addAdjacentNode(tp.getId(),objectSgpNode);
+                    connectors.put(tp.getId(), object);
+                }
+            }
+        }
+        // Remove all potential cyclic connector bindings
+        final Set<Integer> connectorsToBeRemoved = GraphCycleDetector.determineCyclicConnectors(sgpNodes);
+        for(final int i : connectorsToBeRemoved){
+            connectors.remove(i);
+        }
```
scferrada, line 84: Here I think you can use `removeAll` to avoid this iteration.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+
+        // Creating necessary variables for the GraphQL query
+        final TreeSet<String> fieldPaths = new TreeSet<>();
```
scferrada, line 87: Please use `Set` as the type. Also, the usual question if it is really necessary to keep a sorted map instead of a `HashSet`

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+        final JsonObject argumentValues = new JsonObject();
+        final Map<String, String> argumentDefinitions = new HashMap<>();
+
+        // Get all SGP without a connector, if they have undeterminable GraphQL type then materializeAll
+        final Map<Node,String> withoutConnector = new HashMap<>();
+        boolean isDeterminable = true;
+        for(final Node n : subgraphPatterns.keySet()){
+            final String sgpType = determineSgpType(subgraphPatterns.get(n),config);
+            if(!connectors.values().contains(n) && sgpType == null){
```
scferrada, line 96: Use `containsValue(n)` instead of `values().contains(n)`

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+                materializeAll(fieldPaths,subgraphPatterns,connectors,config,endpoint);
+                isDeterminable = false;
+                break;
+            }
+            if(!connectors.values().contains(n)){
```
scferrada, line 101: Same here

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+
+        // If the entire view doesn't need to be materialized
+        if(isDeterminable){
+            int entrypointCounter = 0;
+            int variableCounter = 0;
+            for(final Node n : withoutConnector.keySet()){
+                final Map<String,LiteralLabel> sgpArguments = getSgpArguments(subgraphPatterns.get(n),config);
+                final String sgpType = withoutConnector.get(n);
+                final StringBuilder currentPath = new StringBuilder();
+                GraphQLEntrypoint e;
```
scferrada, line 114:
```suggestion
                final GraphQLEntrypoint e;
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+
+                if(hasNecessaryArguments(sgpArguments.keySet(), 
+                        endpoint.getEntrypoint(sgpType,GraphQLEntrypointType.SINGLE).getArgumentDefinitions().keySet(), 
+                        true)){
+                    // Get single object entrypoint
+                    e = endpoint.getEntrypoint(sgpType,GraphQLEntrypointType.SINGLE);
+                    final StringBuilder entrypointAlias = new StringBuilder("ep_single")
+                    .append(entrypointCounter).append(":").append(e.getFieldName());
+                    currentPath.append(entrypointAlias.toString());
```
scferrada, line 123: No need to call the `toString` method here

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+                final StringBuilder currentPath = new StringBuilder();
+                GraphQLEntrypoint e;
+
+                if(hasNecessaryArguments(sgpArguments.keySet(), 
+                        endpoint.getEntrypoint(sgpType,GraphQLEntrypointType.SINGLE).getArgumentDefinitions().keySet(), 
+                        true)){
+                    // Get single object entrypoint
+                    e = endpoint.getEntrypoint(sgpType,GraphQLEntrypointType.SINGLE);
+                    final StringBuilder entrypointAlias = new StringBuilder("ep_single")
+                    .append(entrypointCounter).append(":").append(e.getFieldName());
```
scferrada, line 122: maybe consider adding a method to the Entrypoint class that, given a fieldname and counter returns this alias

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+
+    /**
+     * Recursive function used to add fields from the TPs in a given SGP
+     */
+    protected void addSgp(final TreeSet<String> fieldPaths, final Map<Node,Set<TP>> subgraphPatterns, 
+            final Map<Integer,Node> connectors, final GraphQL2RDFConfiguration config, final GraphQLEndpoint endpoint, 
+            final Node subgraphNode, final String currentPath, final String sgpType){
+
+        // Necessary id field present in all objects used to indentify the GraphQL object
+        fieldPaths.add(new StringBuilder(currentPath).append("id_").append(sgpType).append(":id").toString());
```
scferrada, line 183: There's no need to use `StringBuilder`s everywhere, they are mainly useful when you're making a string on a loop, not necessarily when creating just one string in one line.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+        final String nestedType = endpoint.getGraphQLFieldValueType(sgpType, fieldName);
+
+        addSgp(fieldPaths,subgraphPatterns,connectors,config,endpoint,
+            nestedSubgraphNode,newPath.toString(),nestedType);
+    }
+
+    /**
+     * Helper function to add a field to the query that represents a nested object that only needs an id field (no more recursive calls)
+     */
+    protected void addEmptyObjectField(final TreeSet<String> fieldPaths, final Map<Node,Set<TP>> subgraphPatterns,
```
scferrada, line 273: You don't use `subgraphPatterns` nor `connectors`, please remove them from the argument list

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+            final Node object = t.getTriplePattern().asJenaTriple().getObject();
+
+            if(predicate.isURI() && predicate.getURI().startsWith(config.getPropertyPrefix())){
+                return config.getClassFromPropertyURI(predicate.getURI());
+            }
+            else if(predicate.isURI() && predicate.getURI().equals(config.getClassMembershipURI())){
+                if(object.isURI() && object.getURI().startsWith(config.getClassPrefix())){
+                    final int splitIndex = config.getClassPrefix().length();
+                    final String sgpType = object.getURI().substring(splitIndex);
+                    return sgpType;
```
scferrada, line 357: directly return the computed value, no need to store it in a variable before returning

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+
+        if(type.equals(int.class)){
+            return JsonNumber.value((int) literal.getValue());
+        }
+        else if(type.equals(boolean.class)){
+            return new JsonBoolean((boolean) literal.getValue());
+        }
+        else if(type.equals(double.class) || type.equals(float.class)){
+            return JsonNumber.value((double) literal.getValue());
+        }
```
scferrada, line 428: My IDE tells me that these conditions are always `false`. Probably because `int.class` and such return something with a different type than `Class<?>`. Please reconsider the utility of this method.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,454 @@
...
+
+        public final int getId() {
+            return id;
+        }
+
+        public final TriplePattern getTriplePattern(){
+            return triplePattern;
+        }
+    }
+}
```
scferrada, line 454: add a newline at the end of the file

scferrada (changes requested):
Hi Wiktor. Here goes my initial feedback. The class is still a bit hard to digest, so I think we will need to break it down a bit more after you address my comments.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
WiktorKarlsson: The reason `fieldPaths` is a TreeSet is because of how the toString method in the `GraphQLQueryImpl` works. Having it as a TreeSet ensures all fields of the same object are together in successive order which really facilitates an easy printing of the query. I could change it to a Set in all the files but then I would have to sort the Set into a TreeSet every time the toString method is used. What do you think?

scferrada: Maybe there is something I'm not seeing, but I don't think that it is worth it to keep a sorted set just to print the GraphQL queries nicely, unless it is a requirement (which I don't think it is).

WiktorKarlsson:
> Hi Wiktor. Here goes my initial feedback. The class is still a bit hard to digest, so I think we will need to break it down a bit more after you address my comments.

@scferrada Do you want me to move methods from the implementation class like `hasVariablePredicate`, `determineSgpType`, `getSgpArguments`, `hasNecessaryArguments`, `literalToJsonValue` and the `TP` class to their own PR and in the utils folder? Or did you mean that you want the larger methods like `translateBGP` and `addSgp` to be divided into smaller functions perhaps?

scferrada:
> > Hi Wiktor. Here goes my initial feedback. The class is still a bit hard to digest, so I think we will need to break it down a bit more after you address my comments.
> 
> @scferrada Do you want me to move methods from the implementation class like `hasVariablePredicate`, `determineSgpType`, `getSgpArguments`, `hasNecessaryArguments`, `literalToJsonValue` and the `TP` class to their own PR and in the utils folder? Or did you mean that you want the larger methods like `translateBGP` and `addSgp` to be divided into smaller functions perhaps?

No need to create a different PR, you can do some of this refactoring here. I think it would make sense to move most of the helper methods to a Utils class, as well as to break down the `translateBGP` method into smaller components, where appropriate.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,247 @@
...
+        if(checkAll){
+            return sgpArguments.containsAll(entrypointArguments);
+        }
+        else{
+            for(final String argName : sgpArguments){
+                if(entrypointArguments.contains(argName)){
+                    return true;
+                }
+            }
+        }
```
scferrada, line 223: these are essentially two different methods, please just split this instead of providing a boolean flag.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,221 @@
...
+            final GraphQLEndpoint endpoint) {
+
+        // Initialize necessary data structures
+        final Map<Node, Set<TP>> subgraphPatterns = createSubGraphPatterns(bgp);
+        final Map<Integer, Node> connectors = createConnectors(subgraphPatterns);
+
+        // Creating necessary variables for the GraphQL query
+        final Set<String> fieldPaths = new HashSet<>();
+        final JsonObject argumentValues = new JsonObject();
+        final Map<String, String> argumentDefinitions = new HashMap<>();
```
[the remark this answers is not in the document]
WiktorKarlsson: Can I get some clarification on if you mean having a **mutable** `GraphQLFieldPath` class that has a Stringbuilder member variable that is added to through internal methods? Or just `GraphQLFieldPath` as a static helper class?

scferrada:
No, it shouldn't be mutable. It can have methods that combine paths, but always return a new, fresh object. Alternatively, you can have a PathBuilder object where you can append things and then build and get a Path object

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+
+import se.liu.ida.hefquin.engine.federation.GraphQLEndpoint;
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.GraphQL2RDFConfiguration;
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.data.impl.GraphQLFieldType;
+
+public class SPARQL2GraphQLHelper {
+    /**
+     * Recursive function used to add fields from the triple patterns in a given SGP
+     */
+    public static void addSgp(final Set<String> fieldPaths, final Map<Node,Set<TP>> subgraphPatterns,
```
scferrada, line 24: instead of adding stuff to a given Set, please return the set of created Paths.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+    /**
+     * Returns the GraphQL object type @param sgp correponds to. 
+     * If the type is undeterminable @return null
+     */
+    public static String determineSgpType(final Set<TP> sgp, final GraphQL2RDFConfiguration config){
+        for(final TP t : sgp){
+            final Node predicate = t.getTriplePattern().asJenaTriple().getPredicate();
+            final Node object = t.getTriplePattern().asJenaTriple().getObject();
+
+            if(predicate.isURI() && predicate.getURI().startsWith(config.getPropertyPrefix())){
```
scferrada, line 161: the second condition should be a method of the Configuration interface, that receives a string and says if it is a possible property (is that the correct GraphQL nomenclature?) or not.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+                return config.getClassFromPropertyURI(predicate.getURI());
+            }
+            else if(predicate.isURI() && predicate.getURI().equals(config.getClassMembershipURI())){
+                if(object.isURI() && object.getURI().startsWith(config.getClassPrefix())){
```
scferrada, line 165: same here, the configuration should have methods to check if a uri is the class membership uri, and if an uri is a valid class/type. Having this String equality and startsWith calls is forcing a given type of implementation that might change in the future.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+                    final int splitIndex = config.getClassPrefix().length();
+                    return object.getURI().substring(splitIndex);
```
scferrada, line 167: the configuration should be able to map to/from RDF View URIs to GraphQL type and field names. See the `LPG2RDFConfiguration` class.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+        }
+
+        return null;
+    }
+
+    /**
+     * @return a map consisting of what can be used as arguments from the given @param sgp
+     * The predicate from a TP needs to be a property URI and the object needs to be a literal
+     */
+    public static final Map<String,LiteralLabel> getSgpArguments(final Set<TP> sgp, final GraphQL2RDFConfiguration config){
```
scferrada, line 179: why is the method final?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+        final Map<String,LiteralLabel> args = new HashMap<>();
+        for(final TP t : sgp){
+            final Node predicate = t.getTriplePattern().asJenaTriple().getPredicate();
+            final Node object = t.getTriplePattern().asJenaTriple().getObject();
+
+            if(predicate.isURI() && predicate.getURI().startsWith(config.getPropertyPrefix())){
+                if(object.isLiteral() && object.getLiteral().isWellFormed()){
```
scferrada, line 186: is it usually a concern that a literal is not well formed?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+                    // Remove prefix and suffix before adding to map
+                    final String s = config.removePropertyPrefix(predicate.toString());
+                    final String argName = config.removePropertySuffix(s);
```
scferrada, line 189: this mapping should be handled by the configuration (same with the predicate condition in line 185)

scferrada (changes requested):
Hi Wiktor, here are some more comments. I think we're getting close :).

The main changes I'd like to have are the inclusion of the FieldPath classes, and the direct map/unmap and check of URIs in the configuration.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/impl/GraphQLEntrypointImpl.java
```diff
@@ -9,12 +9,14 @@ public class GraphQLEntrypointImpl implements GraphQLEntrypoint {
     protected final String fieldName;
     protected final Map<String,String> argumentDefinitions;
     protected final String type;
+    protected final GraphQLEntrypointType epType;
 
     public GraphQLEntrypointImpl(final String fieldName, final Map<String,String> argumentDefinitions,
-            final String type){
+            final String type, final GraphQLEntrypointType epType){
```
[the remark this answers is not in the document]
WiktorKarlsson: the "type" string refers to the name of what type of objects the entrypoint fetches (i.e. the graphql types) so the names it can have isn't pre-determined. The epType is simply an enum that describes whether the entrypoint fetches a single object, a filtered list of objects or a full list of objects.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SPARQL2GraphQLTranslatorImpl.java
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Generates query data for @param fieldPaths, @param queryArgs
+     */
+    protected void generateQueryData(final GraphQL2RDFConfiguration config, final GraphQLEndpoint endpoint,
+            final Map<Node,Set<TP>> subgraphPatterns, final Map<Integer,Node> connectors, final Set<String> fieldPaths,
+            final Set<GraphQLArgument> queryArgs, final Map<Node,String> withoutConnector){
+
+        // Counters for creating unique variable and entrypoint names
+        int entrypointCounter = 0;
+        final MutableInt variableCounter = new MutableInt(0);
```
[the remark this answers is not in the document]
WiktorKarlsson: The reason this is not a regular integer is because of the method `addEntrypoint` in the `GraphQLFieldPathBuilder` where the entrypoint and arguments are added to fieldpaths. Its primary purpose is to make sure that the variables used in the arguments doesn't have the same name which means consecutive calls to `addEntrypoint` should still keep creating unique variable names. If it were a normal integer, the incrementing that is done in the loop of `addEntrypoint` after a new unique variable name is created wouldn't change the value of the integer outside of the method. Hope that explains it.

WiktorKarlsson:
> Hi Wiktor, here are some more comments. I think we're getting close :).
> 
> The main changes I'd like to have are the inclusion of the FieldPath classes, and the direct map/unmap and check of URIs in the configuration.

I'll hold of on pushing the PR GraphQLEndpointInitializer (the introspection part) until this is done since the changes in this PR would affect it.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/utils/SPARQL2GraphQLHelper.java
```diff
@@ -0,0 +1,249 @@
...
+     */
+    public static void addSgp(final Set<String> fieldPaths, final Map<Node,Set<TP>> subgraphPatterns, 
+            final Map<Integer,Node> connectors, final GraphQL2RDFConfiguration config, final GraphQLEndpoint endpoint, 
+            final Node subgraphNode, final String currentPath, final String sgpType){
+
+        // Necessary id field present in all objects used to indentify the GraphQL object
+        fieldPaths.add(GraphQLFieldPathBuilder.addID(currentPath, sgpType));
+
+        // Retrieve necessary information about current sgp
+        final boolean addAllFields = hasVariablePredicate(subgraphPatterns.get(subgraphNode));
```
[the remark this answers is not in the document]
WiktorKarlsson: Not sure about this. The boolean you refer to is only ever reached if there is a connector triple pattern in the SGP. There can be variable predicates for the SGP without having any connectors.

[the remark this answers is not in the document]
scferrada: Thanks for the clarification!

scferrada (commented):
Thanks Wiktor, it looks good to me now. Good job. 

We can wait for @hartig's comments now.

hartig (commented):
I have a hard time understanding this, mainly because some of the terminology that you are using here (e.g., SGP, connectors) is unknown to be because I have not read the thesis draft. Do you have these things described in the current draft?

In any case, one thing that I wonder: Do you really need the IDs for the triple patterns? You seem to use these IDs as keys in a Map. Can't you simply use the TriplePattern instead?

WiktorKarlsson:
> I have a hard time understanding this, mainly because some of the terminology that you are using here (e.g., SGP, connectors) is unknown to be because I have not read the thesis draft. Do you have these things described in the current draft?
> 
> In any case, one thing that I wonder: Do you really need the IDs for the triple patterns? You seem to use these IDs as keys in a Map. Can't you simply use the TriplePattern instead?

@hartig I do explain in more detail what I mean with SGP and connectors in the draft. What I mean with SGP (subgraph pattern), is a subset of triple patterns created from the original BGP. The different SGPs that are created are disjunctive when it comes to what triple patterns they contain. The triple patterns are sorted into these SGPs in regard to what their triple pattern subject node is. 

As for the connectors, they are basically pointers from certain triple patterns to subject nodes (which represents entire SGPs) and are used to determine how the different SGPs are connected to each other (which is needed information when constructing the query later). A connector is created if a triple pattern object node equals a subject node of a different SGP.

Finally, I think you're right about triple patterns not needing an integer id so I have refactored out the wrapper class for the triple patterns in my latest commit (and verified that the code still works).

WiktorKarlsson:
> @WiktorKarlsson thanks for the clarification. What you call "SGP" or "subgraph pattern" is usually referred to as 'star pattern' or, even more precisely, 'subject-based star pattern'. The conversion of a BGP into its subject-based star patterns is called 'subject-based star decomposition'. Are you okay if I update the code in this PR to reflect this terminology? (... to help me understand this code a bit better.)

Thanks for your clarification on the terminology, I'll be sure to update my thesis draft accordingly. You are free to update the PR if you want, I am not able to to do changes to the PR this weekend as I'm quite busy but I will be ready to fix things on monday again if necessary.

hartig:
@WiktorKarlsson I have reorganized the code quite a bit, mainly to separate functionality into more relevant places, but also to avoid passing a lot of objects from function to function and to avoid repeated+unnecessary look ups into HashMaps. Overall, this simplifies the code and makes it more readable (I think). While more such changes are possible, we can leave it at that for the moment. Please take a look.

WiktorKarlsson:
@hartig I have looked over the changes and I do agree that it does make the code more easier to understand. I added a small commit to fix something I forgot to cover in my previous commits. Other than that I think it looks okay so unless you or sebastian have anything else do you think its ready to merge?

hartig:
Wiktor, regarding your additional new commit, what exactly is the purpose of creating such a TreeSet from the HashSet? If the aim is to achieve a deterministic order of the elements in the set, I am not sure it will be guaranteed this way. Additionally, it is an unnecessary effort to first create and populate the set as a HashSet (in my new `determineRootStarPatterns` helper function) and then recreate it as a TreeSet. How's about switching to a List for this set and create the list already in `determineRootStarPatterns`?

WiktorKarlsson:
@hartig I have followed your advice and removed the TreeSet completely in the latest commit. The test class is reworked and should be able to handle the tests correctly even without the TreeSet.

hartig (approved):
Looks okay now.

PR #195, opened 2022-05-27
Implemented the logical and physical Local2Global operators.

SHSenpai:
Both the logical and physical LocalToGlobal operators are now implemented, the visitors and printers have been updated accordingly. I'm assuming that the expected number of input variables to the operator, given that it is a unary operator, is 1.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpLocalToGlobal.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpLocalToGlobal.java
@@ -1,0 +1,24 @@
+package se.liu.ida.hefquin.engine.queryplan.logical.impl;
+
+import se.liu.ida.hefquin.engine.data.VocabularyMapping;
+import se.liu.ida.hefquin.engine.queryplan.logical.LogicalPlanVisitor;
+import se.liu.ida.hefquin.engine.queryplan.logical.UnaryLogicalOp;
+
+public class LogicalOpLocalToGlobal implements UnaryLogicalOp {
+
+	protected final VocabularyMapping vocabularyMapping;
+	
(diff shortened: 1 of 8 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpLocalToGlobal.java
```diff
@@ -0,0 +1,24 @@
...
+	
+	public LogicalOpLocalToGlobal(final VocabularyMapping mapping){
+		this.vocabularyMapping = mapping;
+	}
+	
+	@Override
+	public void visit(final LogicalPlanVisitor visitor) {
+		visitor.visit(this);
+	}
+
```
hartig, line 19:
```suggestion

	@Override
	public String toString() {
		return "> l2g " + "(vocab.mapping: " + vocabularyMapping.hashCode() + ")";
	}

```

src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpLocalToGlobal.java
```diff
@@ -0,0 +1,44 @@
...
+	@Override
+	public UnaryExecutableOp createExecOp(final ExpectedVariables... inputVars) {
+		return new ExecOpLocalToGlobal( logicall2g.getVocabularyMapping() );
+	}
+
+	@Override
+	public UnaryLogicalOp getLogicalOperator() {
+		return this.logicall2g;
+	}
+
```
hartig, line 43:
```suggestion

	@Override
	public String toString() {
		return "> l2g " + "(vocab.mapping: " + logicall2g.getVocabularyMapping().hashCode() + ")";
	}

```

hartig (changes requested):
Looks good. Just two small additions.

PR #200, opened 2022-06-16
GlobalToLocal operators, with printer and visitor methods.

SHSenpai:
The same thing that was done for LocalToGlobal, but for GlobalToLocal.


src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanVisitorBase.java
```diff
@@ -29,4 +30,6 @@ public void visit( final LogicalOpMultiwayUnion op )     {}
 	public void visit( final LogicalOpFilter op )     {}
 
 	public void visit(LogicalOpLocalToGlobal op)	{}
+
+	public void visit(LogicalOpGlobalToLocal op)	{}
```
hartig, line 34:
```suggestion
	public void visit( final LogicalOpLocalToGlobal op )	{}

	public void visit( final LogicalOpGlobalToLocal op )	{}
```

hartig (approved):
Great. Thanks!

PR #202, opened 2022-06-22
Update parameters of terminate criteria

chengsijin0817:


hartig:
The classes that implemented `TerminationCriterion` were meant to be stateless (i.e., an instance of such a class can be used in parallel by multiple query optimizations). Based on what you did in the PR, they are not stateless anymore. As a consequence, if a second query optimization calls your `initialize` function, then that changes the behavior of a first query optimization that uses the same instance of the `TerminationCriterion`. I have fixed this issue by introducing a factory patterns.

Additionally, I have introduced `TerminationCriterionBase` as an abstract base class for all implementations of `TerminationCriterion`. This way, the code that deals with the given logical plan to initialize `nrGenerations` does not need to be repeated in all the actual classes.

Take a look at the files in this PR.

I will merge the PR now.

PR #204, opened 2022-06-30
JSON2SolutionMappings created with tests

WiktorKarlsson:
JSON2SolutionMappings module with some unit tests.


scferrada (changes requested):
Hi Wiktor,

Here are some comments. They are mainly regarding the responsibilities of the different objects. In this sense, I think that this JSON2SolutionMappings shouldn't be the one responsible to query the produced model. So, I'd recommend that this object is renamed and that the translateJSON method returns an RDF graph with the triples generated from the JSON.

Then, a different object can query this RDF graph and produce the corresponding Solution Mappings.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/JSON2SolutionGraphImpl.java
```diff
@@ -0,0 +1,233 @@
...
+     */
+    protected Resource parseJSON(final JsonObject root, 
+                                 final Map<BlankNodeID,Resource> blankNodes, 
+                                 final Model solutionGraph) throws ParseException {
+
+        // Get the key that represent the current json objects id
+        final String idKey = root.keySet()
+            .stream()
+            .filter(s -> s.startsWith(config.getJsonIDKeyPrefix()))
+            .findAny()
```
scferrada, line 85: why does finding any work here, is the intuition that there will only be one?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/JSON2SolutionGraphImpl.java
```diff
@@ -0,0 +1,233 @@
...
+            .orElse(null);
+
+        if(idKey == null){
+            throw new ParseException("JsonObject root did not provide a valid id key.");
+        }
+
+        // Initialize the necessary data needed for parsing the current json object
+        final String graphqlID = root.getString(idKey);
+        root.keySet().remove(idKey);
```
scferrada, line 94: why is it necessary to modify the incoming object like this? If it is needed so that the next time this method is called, we don't consume the same object again, then this logic should be managed by the method that makes the call and not this method.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/JSON2SolutionGraphImpl.java
```diff
@@ -0,0 +1,233 @@
...
+        final String graphqlType = removeJsonKeyPrefix(idKey);
```
scferrada, line 95: Maybe this should be managed by the configuration? Alternatively, this can be a method on a utility class.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/JSON2SolutionGraphImpl.java
```diff
@@ -0,0 +1,233 @@
...
+                return model.createTypedLiteral(primitive.getAsNumber().value(), XSDPlainType.XSDdouble);
+            default:
+                return model.createTypedLiteral(primitive.getAsString().value(), XSDPlainType.XSDstring);
+        }
+    }
+
+    /**
+     * Utility class used to represent a double valued key for Maps etc.
+     */
+    protected class BlankNodeID {
```
scferrada, line 204: This class can be replaced for a HeFQUIN's `Pair<String, String>` or a Jena's `PairOfSameType<String>`, since it doesn't add any special functionality.

scferrada (changes requested):
Thanks for addressing my previous comments. There's still a few things I would like to change.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/JSON2SolutionGraphImpl.java
WiktorKarlsson: yes, each json object must have **exactly one** key with the "ID" prefix  for the approach to work.
The amount of scalar or object keys in the json can vary though.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/JSON2SolutionGraphImpl.java
WiktorKarlsson: The original reason for why I removed the id key from the keyset was so I wouldn't need to check key prefixes on all keys later on while iterating. If the id key was left in the keyset while iterating over the keys then it would have been identified as a "scalar" which would be wrong. In the latest commit ([4784f02](https://github.com/LiUSemWeb/HeFQUIN/pull/204/commits/4784f02cab6ccce64dc491a9e213161a9f511226)), all of the if-cases now uses key prefix checks and I have removed the `root.keySet().remove(idKey);`.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/JSON2SolutionGraph.java
```diff
@@ -0,0 +1,18 @@
+package se.liu.ida.hefquin.engine.wrappers.graphqlwrapper;
+
+import org.apache.jena.atlas.json.io.parserjavacc.javacc.ParseException;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.atlas.json.JsonException;
+
+import se.liu.ida.hefquin.engine.federation.access.JSONResponse;
+
+public interface JSON2SolutionGraph {
```
scferrada, line 9: Add an action suffix to this name, so that someone that doesn't know the approach can know what this class does (e.g., `JSON2SolutionGraphConverter`

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/SolutionGraph2SolutionMappings.java
```diff
@@ -0,0 +1,20 @@
+package se.liu.ida.hefquin.engine.wrappers.graphqlwrapper;
+
+import java.util.List;
+
+import org.apache.jena.rdf.model.Model;
+
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.query.SPARQLQuery;
+
+public interface SolutionGraph2SolutionMappings {
```
scferrada, line 10: Also find a more suitable name here. See that the only method executes a SPARQL Query over a Model.

I'm also now curious if there is simply a Jena funcionality/utility that can be used instead of implementing this method.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SolutionGraph2SolutionMappingImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.data.impl.SolutionMappingImpl;
+import se.liu.ida.hefquin.engine.query.SPARQLQuery;
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.SolutionGraph2SolutionMappings;
+
+public class SolutionGraph2SolutionMappingImpl implements SolutionGraph2SolutionMappings {
+
+    @Override
+    public List<SolutionMapping> querySolutionGraph(final Model solutionGraph, 
+                                                    final SPARQLQuery query) throws RuntimeException {
```
scferrada, line 20: I don't think that Runtime exceptions need to be declared in the method signature.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/SolutionGraph2SolutionMappingImpl.java
```diff
@@ -0,0 +1,35 @@
...
+        final List<SolutionMapping> solutionMappings = new ArrayList<>();
+
+        try (final QueryExecution qexec = QueryExecutionFactory.create(query.asJenaQuery(), solutionGraph)) {
+            final ResultSet results = qexec.execSelect();
+            for (; results.hasNext();) {
+                solutionMappings.add(new SolutionMappingImpl(results.nextBinding()));
+            }
+        }
+        catch(final RuntimeException e){
+            throw new RuntimeException("Something went wrong when executing the SPARQL query over the provided solutionGraph!",e);
```
scferrada, line 30: Add something to this message so that we know it is about querying the RDF-view of a GraphQL endpoint. Querying a solution graph is kind of a common thing that happens with this engine.

scferrada (changes requested):
Hi Wiktor, it looks much better now, thanks. There's some small things left, and one question.

scferrada (approved):
Thanks Wiktor for your replies. I wonder now if this Jena Model querying utility should be on a different package, since it is a common interest utility. We can leave it like this for the time being and revisit this idea later.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/GraphQLSolutionGraphSolver.java
```diff
@@ -0,0 +1,20 @@
...
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.query.SPARQLQuery;
+
+public interface GraphQLSolutionGraphSolver {
+    
+    /**
+     * Queries @param solutionGraph using the SELECT query @param query and
+     * @return a list of corresponding solution mappings.
+     * @throws RunTimeException if for some reason the execution of the query over the
+     * given solutionGraph did not work correctly.
```
hartig, line 16: Please don't throw runtime exceptions. Instead, throw ordinary exceptions (create special classes for them); that way, these exceptions need to be explicitly declared and the compiler complains if they are not caught. Hence, when using this code somewhere else, one is immediately aware of the exceptions and has to take them into account when coding. In contrast, runtime exceptions have the tendency to be forgotten to be handled.

hartig (changes requested):
Just one thing.

hartig (approved):
Thanks!

PR #206, opened 2022-06-30
GraphQL endpoint initializer

WiktorKarlsson:
GraphQL endpoint initializer with unit tests.


src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLField.java
```diff
@@ -14,6 +14,7 @@ public interface GraphQLField {
 
     /**
      * @return the GraphQL value type of the field (String,Int,... etc.)
+     * Note: does not include List or Non-nullable identifiers
```
scferrada, line 17: this means that the method returns null in such cases? or the empty string? or an error?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/GraphQLEndpointInitializerImpl.java
```diff
@@ -0,0 +1,232 @@
...
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.query.GraphQLQuery;
+import se.liu.ida.hefquin.engine.wrappers.graphqlwrapper.query.impl.GraphQLQueryImpl;
+
+public class GraphQLEndpointInitializerImpl implements GraphQLEndpointInitializer {
+
+    @Override
+    public GraphQLEndpoint initializeEndpoint(final String url) throws FederationAccessException {
+        final GraphQLInterface iface = new GraphQLInterfaceImpl(url);
+        final GraphQLEndpoint tmpEndpoint = new GraphQLEndpointImpl(null, null, iface);
+        final GraphQLRequestProcessor requestProcessor = new GraphQLRequestProcessorImpl(5000,5000);
```
scferrada, line 38: Maybe there should be a way to customize these parameters, instead of hardcoding 5000s

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/GraphQLEndpointInitializerImpl.java
```diff
@@ -0,0 +1,232 @@
...
+        final TreeSet<String> fieldPaths = new TreeSet<>();
+
+        // Name and kind of type (OBJECT, SCALAR etc.)
+        fieldPaths.add("__schema/types/name");
+        fieldPaths.add("__schema/types/kind");
+
+        // Fields information
+        fieldPaths.add("__schema/types/fields/name");
+        fieldPaths.add("__schema/types/fields/type/name");
+        fieldPaths.add("__schema/types/fields/type/kind");
```
scferrada, line 69: aren't this paths available as objects now?

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/GraphQLEndpointInitializerImpl.java
```diff
@@ -0,0 +1,232 @@
...
+        catch (final FederationAccessException e) {
+            throw e;
+        }
+    }
+
+    /**
+     * Initializes the GraphQL introspection query to be used.
+     */
+    protected GraphQLQuery getQuery(){
+        final TreeSet<String> fieldPaths = new TreeSet<>();
```
scferrada, line 60: Is the TreeSet necessary? or can this be done with a regular Set?

scferrada (changes requested):
Hi Wiktor,

My main concern with this code is all the hardcode that is going on, specially with all of the strings. I would assume that these introspection queries will not change in the future, but anyhow, I would prefer to have constants with the required paths, so that we can change those in only one place of the code, should we need to change them in the future.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/data/GraphQLField.java
WiktorKarlsson: What I intended for the comment to mean was if a GraphQL type has a field `books: [Book!]!` for example, then the function `getValueType()` would return `Book` and not `[Book!]!`.

scferrada (commented):
Thanks Wiktor, it looks good now. We can wait for @hartig's opinion on this.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/GraphQLEndpointInitializerImpl.java
```diff
@@ -0,0 +1,281 @@
...
+            return determineTypeInformation(field.getObj(ofTypeKey));
+        }
+
+        final String name = field.getString(iName);
+        final String kind = field.getString(iKind);
+        final GraphQLFieldType fieldType;
+        if(kind.equals("OBJECT")){
+            fieldType = GraphQLFieldType.OBJECT;
+        }
+        else{
```
scferrada, line 176: Oh, one small thing. Here, compare directly with "SCALAR" on an else if branch, and on the else branch throw an error.

PR #207, opened 2022-07-01
EntityMappingImpl

SHSenpai:
Checkpointing to take a break.

For applying to TriplePattern, I've assumed that any of (subject, predicate, object) can be a URI.

I'm not quite sure exactly how I'm going to tackle the SolutionMapping methods.
As far as I understand it by looking at code, SolutionMapping contains one binding.
The binding can contain any number of nodes, bound to different variables.
Each node, if it is a URI, can have any number of nodes mapped to them in the g2lMap and l2gMap respectively.

Is this correct? @hartig @chengsijin0817

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
@@ -64,26 +64,36 @@ public class EntityMappingImpl implements EntityMapping
 	@Override
 	public Set<TriplePattern> applyToTriplePattern( final TriplePattern tp ) {
 		final Set<Triple> workingSet = new HashSet<Triple>();
-		final Set<Node> localSubjects = g2lMap.get( tp.asJenaTriple().getSubject() );
-		if( localSubjects != null ) {
-			for (final Node localSubject : localSubjects) {
-				final Triple workingTriple = new Triple(localSubject,tp.asJenaTriple().getPredicate(), tp.asJenaTriple().getObject());
-				workingSet.add(workingTriple);
+		if(tp.asJenaTriple().getSubject().isURI()) {
+			final Set<Node> localSubjects = g2lMap.get( tp.asJenaTriple().getSubject() );
+			if( localSubjects != null ) {
+				for (final Node localSubject : localSubjects) {
+					final Triple workingTriple = new Triple(localSubject,tp.asJenaTriple().getPredicate(), tp.asJenaTriple().getObject());
+					workingSet.add(workingTriple);
+				}
+			} else {
+				workingSet.add(tp.asJenaTriple());
 			}
 		} else {
 			workingSet.add(tp.asJenaTriple());
 		}
-		
 		final Set<TriplePattern> resultSet = new HashSet<TriplePattern>();
-		final Set<Node> localObjects = g2lMap.get(tp.asJenaTriple().getObject());
-		if(g2lMap.containsKey(tp.asJenaTriple().getObject())) {
-			for (final Triple workingTriple : workingSet) {
-				for (final Node localObject : localObjects) {
-					final Triple resultTriple = new Triple(workingTriple.getSubject(), workingTriple.getPredicate(), localObject);
-					resultSet.add(new TriplePatternImpl(resultTriple));
+			if(tp.asJenaTriple().getObject().isURI()) {
+			final Set<Node> localObjects = g2lMap.get(tp.asJenaTriple().getObject());
+			if(g2lMap.containsKey(tp.asJenaTriple().getObject())) {
+				for (final Triple workingTriple : workingSet) {
+					for (final Node localObject : localObjects) {
+						final Triple resultTriple = new Triple(workingTriple.getSubject(), workingTriple.getPredicate(), localObject);
+						resultSet.add(new TriplePatternImpl(resultTriple));
+					}
 				}
+				return resultSet;
+			} else {
+				for (final Triple workingTriple : workingSet ) {
+					resultSet.add(new TriplePatternImpl(workingTriple));
+				}
+				return resultSet;
 			}
-			return resultSet;
 		} else {
 			for (final Triple workingTriple : workingSet ) {
 				resultSet.add(new TriplePatternImpl(workingTriple));
--- a/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
@@ -10,7 +10,7 @@ import org.apache.jena.graph.Graph;
 import org.apache.jena.graph.Node;
 import org.apache.jena.sparql.core.Var;
 import org.apache.jena.sparql.engine.binding.Binding;
-import org.apache.jena.sparql.engine.binding.BindingFactory;
+import org.apache.jena.sparql.engine.binding.BindingBuilder;
 import org.apache.jena.graph.Triple;
 import org.apache.jena.vocabulary.OWL2;
 
@@ -107,25 +107,74 @@ public class EntityMappingImpl implements EntityMapping
 
 	@Override
 	public Set<SolutionMapping> applyToSolutionMapping( final SolutionMapping sm ) {
-		final Set<SolutionMapping> newMappings = new HashSet<SolutionMapping>();
+		final Set<Map<Var,Node>> cartesianProduct = new HashSet<>();
 		final Binding binding = sm.asJenaBinding();
 		final Iterator<Var> it = binding.vars();
+		
+		final Var firstVar = it.next(); // Create a number of single var-node combinations to populate the set with.
+		final Node firstNode = binding.get(firstVar);
+		if (firstNode.isURI()) {
+			final Set<Node> mappedNodes = g2lMap.get(firstNode);
+			if (mappedNodes == null) { // Local Node not different from global. 
+				final Map<Var,Node> newMap = new HashMap<>();
+				newMap.put(firstVar,firstNode);
+				cartesianProduct.add(newMap);
+			} else { // Local different from global exists, use that.
+				for (Node node : mappedNodes) {
+					final Map<Var,Node> newMap = new HashMap<>();
+					newMap.put(firstVar,node);
+					cartesianProduct.add(newMap);
+				}
+			}
+		} else {
+			final Map<Var,Node> newMap = new HashMap<>();
+			newMap.put(firstVar,firstNode);
+			cartesianProduct.add(newMap);
+		}
+		
 		while (it.hasNext()) {
 			final Var var = it.next();
 			final Node node = binding.get(var);
-			final Set<Node> mappedNodes = g2lMap.get(node);
-			if (mappedNodes == null) { // Local Node not different from global. 
-				final Binding newBinding = BindingFactory.binding(var,node);
-				final SolutionMapping newMapping = new SolutionMappingImpl(newBinding);
-				newMappings.add(newMapping);
-			} else { // Local different from global exists, use that.
-				for (final Node localNode : mappedNodes) {
-					final Binding newBinding = BindingFactory.binding(var,localNode);
-					final SolutionMapping newMapping = new SolutionMappingImpl(newBinding);
-					newMappings.add(newMapping);
+			if (node.isURI()) {
+				final Set<Node> mappedNodes = g2lMap.get(node);
+				if (mappedNodes == null) { // Local Node not different from global. 
+					for(Map<Var,Node> map : cartesianProduct) { // Multiply by one by adding the pair to all existing combinations.
+						map.put(var, node);
+					}
+				} else { // Local different from global exists, use that.
+					final Set<Map<Var,Node>> newCartesianProducts = new HashSet<>();
+					for(Map<Var,Node> map : cartesianProduct) {
+						final Iterator<Node> nodeIt = mappedNodes.iterator();
+						final Node newNode = nodeIt.next(); // Add the first of the combinations to each map.
+						map.put(var, newNode);
+						
+						while (nodeIt.hasNext()) { // Add the rest of the multiplications from 2 thru n, if such exists.
+							final Node anotherNewNode = it.next();
+							final Map<Var,Node> newMap = new HashMap<>();
+							newMap.putAll(map);
+							newMap.replace(var, anotherNewNode);
+							newCartesianProducts.add(newMap);
+						}
+					}
+					cartesianProduct.addAll(newCartesianProducts); // Merge the two to create the next step cartesian product.
+				}
+			} else {
+				for(Map<Var,Node> map : cartesianProduct) { // Multiply by one by adding the pair to all existing combinations.
+					map.put(var, node);
 				}
 			}
 		}
+
+		final Set<SolutionMapping> newMappings = new HashSet<SolutionMapping>();
+		
+		for (Map<Var,Node> map : cartesianProduct) { // For each entry in the product, create a builder, build the binding, and make it a solution mapping.
+			final BindingBuilder builder = BindingBuilder.create();
+			map.forEach((var,node) -> builder.add(var, node));
+			final Binding newBinding = builder.build();
+			final SolutionMapping newMapping = new SolutionMappingImpl(newBinding);
+			newMappings.add(newMapping);
+		}
+		
 		return newMappings;
 	}
 
--- a/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
@@ -6,9 +6,12 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiConsumer;
 
 import org.apache.jena.graph.Graph;
 import org.apache.jena.graph.Node;
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.Binding;
 import org.apache.jena.graph.Triple;
 import org.apache.jena.vocabulary.OWL2;
 
@@ -104,6 +107,22 @@ public class EntityMappingImpl implements EntityMapping
 
 	@Override
 	public Set<SolutionMapping> applyToSolutionMapping( final SolutionMapping sm ) {
+		/*final BiConsumer<Var,Node> apply = (var,node) -> {
+			
+		};*/
+		final Binding binding = sm.asJenaBinding();
+		final Iterator<Var> it = binding.vars();
+		while(it.hasNext()) {
+			final Var var = it.next();
+			final Node node = binding.get(var);
+			final Set<Node> mappedNodes = g2lMap.get(node);
+			if(mappedNodes == null) {
+				// No translation exists, use original node.
+			} else {
+				// Local different from global exists, use that.
+			}
+		}
+		
 		// TODO: implement this function
 		return Collections.singleton(sm);
 	}
--- a/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
@@ -115,28 +115,8 @@ public class EntityMappingImpl implements EntityMapping
 		}
 
 		final Set<Map<Var,Node>> cartesianProduct = new HashSet<>();
+		cartesianProduct.add(new HashMap<Var,Node>());
 		final Iterator<Var> it = binding.vars();
-		
-		final Var firstVar = it.next(); // Create a number of single var-node combinations to populate the set with.
-		final Node firstNode = binding.get(firstVar);
-		if (firstNode.isURI()) {
-			final Set<Node> mappedNodes = g2lMap.get(firstNode);
-			if (mappedNodes == null) { // Local Node not different from global. 
-				final Map<Var,Node> newMap = new HashMap<>();
-				newMap.put(firstVar,firstNode);
-				cartesianProduct.add(newMap);
-			} else { // Local different from global exists, use that.
-				for (Node node : mappedNodes) {
-					final Map<Var,Node> newMap = new HashMap<>();
-					newMap.put(firstVar,node);
-					cartesianProduct.add(newMap);
-				}
-			}
-		} else {
-			final Map<Var,Node> newMap = new HashMap<>();
-			newMap.put(firstVar,firstNode);
-			cartesianProduct.add(newMap);
-		}
 		
 		while (it.hasNext()) {
 			final Var var = it.next();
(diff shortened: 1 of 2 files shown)


src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,86 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
 		// corresponding local URI
 		final Iterator<Triple> it = mappingDescription.find(Node.ANY, OWL2.sameAs.asNode(), Node.ANY);
 		while ( it.hasNext() ) {
-			// TODO ...
+			if(g2lMap.containsKey(it.next().getSubject())) {
+				// There is already at least one local URI for this global URI,
+				// We don't want to override it, so we'll add the new one to the set.
+				final Set<Node> locals = g2lMap.get(it.next().getSubject());
+				locals.add(it.next().getObject());
+				g2lMap.replace(it.next().getSubject(),locals);
```
hartig, line 41: This should not be necessary. The program variable `locals` is a direct reference to the Java object that is also referenced from within the map. Manipulating that Java object via `locals.add` is enough. You don't need to put the reference to that Java object again into the map.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,86 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
 		}
 	}
 
 	@Override
 	public Set<TriplePattern> applyToTriplePattern( final TriplePattern tp ) {
-		// TODO: implement this function
-		return Collections.singleton(tp);
+		final Set<Triple> workingSet = new HashSet<Triple>();
+		if(g2lMap.containsKey(tp.asJenaTriple().getSubject())) {
+			final Iterator<Node> it = g2lMap.get(tp.asJenaTriple().getSubject()).iterator();
```
hartig, line 66: Same comment as above. Avoid doing twice the same type of lookup into `g2lMap`.
```suggestion
		final Set<Node> localSubjects = g2lMap.get( tp.asJenaTriple().getSubject() );
		if( localSubjects != null ) {
			final Iterator<Node> it = localSubjects.iterator();
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,86 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+			while (it.hasNext()) {
```
hartig, line 67: Once you have the `localSubjects` set as suggested above, you can simplify this code by avoiding the use of an iterator and, instead, do:
```java
		for ( final Node localSubject : localSubjects ) {
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,86 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+				final Triple workingTriple = new Triple(it.next(),tp.asJenaTriple().getPredicate(),tp.asJenaTriple().getObject());
```
hartig, line 68: Please add a space after each comma. That makes the code a bit more readable because the three arguments are easier to spot.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,86 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+				final Triple workingTriple0 = it0.next();
+				final Iterator<Node> it = g2lMap.get(tp.asJenaTriple().getPredicate()).iterator();
+				while (it.hasNext()) {
+					final Triple workingTriple = new Triple(workingTriple0.getSubject(),it.next(),workingTriple0.getObject());
+					workingSet2.add(workingTriple);
+				}
+			}
+		} else {
+			workingSet2.addAll(workingSet);
+		}
```
hartig, line 88: This whole part can be removed. URIs for which semantic equivalence can be expressed by means of owl:sameAs statements can be used only in the subject or object position of triples, but not in the predicate position. Therefore, triple patterns that have such a URI in the predicate position can, in principle, have no matching triples anyways.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,86 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
 		// statements in the given RDF graph; assume the the subject of
 		// any owl:sameAs statement is a global URI and the object is the
 		// corresponding local URI
 		final Iterator<Triple> it = mappingDescription.find(Node.ANY, OWL2.sameAs.asNode(), Node.ANY);
 		while ( it.hasNext() ) {
-			// TODO ...
+			if(g2lMap.containsKey(it.next().getSubject())) {
+				// There is already at least one local URI for this global URI,
+				// We don't want to override it, so we'll add the new one to the set.
+				final Set<Node> locals = g2lMap.get(it.next().getSubject());
```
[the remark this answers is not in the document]
hartig: And one more thing: before trying to change `g2lMap` and `l2gMap`, check that both the subject and the object of `t` is a URI and, if it is not, then ignore the current `t` and move to the next one.
```java
if ( ! t.getSubject().isURI() || ! t.getObject().isURI() ) {
	continue; // owl:sameAs statements that are not about URIs are not relevant
}
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
hartig: Also, as another small optimization, before even attempting to replace the subject of the given triple pattern, check that the subject is a URI. If it is not, put `tp` into `workingSet`.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,86 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+				}
+			}
+			return resultSet;
+		} else {
+			final Iterator<Triple> it = workingSet2.iterator();
+			while (it.hasNext()) {
+				resultSet.add(new TriplePatternImpl(it.next()));
+			}
+			return resultSet;
+		}
```
hartig, line 109: All the comments from above apply here too (check first that the object of `tp` is a URI, avoid unnecessary lookups into `g2lMap`, instead of iterators better use for-loops).

hartig (changes requested):
Here are my comments on the current code. There are some issues to be fixed and some things to be improved.

Regarding your question related to the SolutionMapping-related methods, your understanding is correct.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,67 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+				continue;
+			}
+			final Set<Node> locals = g2lMap.get(workingTriple.getSubject());
+			if(locals != null) {
+				// There is already at least one local URI for this global URI,
+				// We don't want to override it, so we'll add the new one to the set.
+				locals.add(workingTriple.getObject());
+			} else {
+				// g2lmap doesn't contain any local URI for this global yet
+				final Set<Node> localSet = new HashSet<Node>();
```
hartig, line 47: You need to put this new set into the `g2lMap`.
```suggestion
				final Set<Node> localSet = new HashSet<>();
				g2lMap.put( workingTriple.getSubject(), localSet );
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,67 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+				localSet.add(workingTriple.getObject());
+			}
+			
+			// The same for l2g.
+			final Set<Node> globals = l2gMap.get(workingTriple.getObject());
+			if(globals != null) {
+				globals.add(workingTriple.getSubject());
+			} else {
+				final Set<Node> newGlobalSet = new HashSet<Node>();
```
hartig, line 56: Same issue here.
```suggestion
				final Set<Node> newGlobalSet = new HashSet<>();
				l2gMap.put( workingTriple.getObject(), newGlobalSet );
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,67 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+			}
 		}
 	}
 
 	@Override
 	public Set<TriplePattern> applyToTriplePattern( final TriplePattern tp ) {
-		// TODO: implement this function
-		return Collections.singleton(tp);
+		final Set<Triple> workingSet = new HashSet<Triple>();
+		final Set<Node> localSubjects = g2lMap.get( tp.asJenaTriple().getSubject() );
```
hartig, line 65: Since the mapping is only about URIs, doing `g2lMap.get( tp.asJenaTriple().getSubject() )` can be successful only if the subject of the triple pattern is a URI. Consequently, if the subject is not a URI, you can simply put the triple pattern as is into `workingSet` and move on to the part about the object of the triple pattern.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,20 +27,67 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+			for (final Node localSubject : localSubjects) {
+				final Triple workingTriple = new Triple(localSubject,tp.asJenaTriple().getPredicate(), tp.asJenaTriple().getObject());
+				workingSet.add(workingTriple);
+			}
+		} else {
+			workingSet.add(tp.asJenaTriple());
+		}
+		
+		final Set<TriplePattern> resultSet = new HashSet<TriplePattern>();
+		final Set<Node> localObjects = g2lMap.get(tp.asJenaTriple().getObject());
```
hartig, line 76: Same issue here. If the object of the triple pattern is not a URI, you can simply return `workingSet` as the result of this function.

hartig (changes requested):

hartig (commented):
Looks good for the given part now.

SHSenpai:
@hartig I've started with the applyToSolutionMapping-method and I'm unsure how to best create the new bindings to use in the new solution mappings. I'm currently iterating through the Var-Node pairs of the binding in the SolutionMapping that is passed to the method as a parameter, and looking into g2lMap to get the local-vocabulary nodes.

I'm currently thinking that I will use BindingFactory to create a one-pair binding for each Var-Node pair, create a new solution mapping using each binding, and returning a set of those, but I'm not sure how it's going to work when it comes to the same Var being bound to multiple Nodes, as may happen when g2lMap returns a set of several Nodes.

scferrada:
Hi Simon,

I believe that `BindingFactory` is deprecated, you may use `BindingBuilder` instead to construct variable bindings. Notice that the builder does not allow you to bind the same variable to two different `Node` objects in the same binding. This means, that you need to create a builder for each one of the solution mappings in the set.

There is code where I use the binding builder here: https://github.com/LiUSemWeb/HeFQUIN/blob/updateConvertFunction/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/Record2SolutionMappingTranslatorImpl.java

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,26 +30,103 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+		final Binding binding = sm.asJenaBinding();
+		final Iterator<Var> it = binding.vars();
+		while (it.hasNext()) {
+			final Var var = it.next();
+			final Node node = binding.get(var);
+			final Set<Node> mappedNodes = g2lMap.get(node);
+			if (mappedNodes == null) { // Local Node not different from global. 
+				final Binding newBinding = BindingFactory.binding(var,node);
+				final SolutionMapping newMapping = new SolutionMappingImpl(newBinding);
+				newMappings.add(newMapping);
```
hartig, line 120: That's incorrect because it considers only one of the variables in the given solution mapping. Even worse, it considers each variable in the given solution mapping separately.

Consider the following example: Assume the given solution mapping `sm` is: sm = { ?x -> ex:Bob, ?y -> ex:Ali }.

If neither ex:Bob nor ex:Ali are mentioned in the entity mapping, then the `applyToSolutionMapping` function should return `sm` unchanged. Instead, your current implementation returns two new solution mappings, one of them only with ?x -> ex:Bob and the other one only with ?y -> ex:Ali.

If, in contrast, the entity mapping maps ex:Bob to ex:Robert and ex:Ali to ex:Alibaba, then the `applyToSolutionMapping` function should return a single new solution mapping with ?x -> ex:Robert and ?y -> ex:Alibaba. Instead, your current implementation (in the else-branch below) creates two separate solution mappings, one of them only with ?x -> ex:Robert and the other one only with ?y -> ex:Alibaba.

If the entity mapping does not map ex:Ali, but it maps ex:Bob to both ex:Robert and ex:Bobby, then the `applyToSolutionMapping` function should return two new solution mappings, one with ?x -> ex:Robert and ?y -> ex:Ali, and the other one with ?x -> ex:Bobby and ?y -> ex:Ali. Instead, your current implementation would return three solution mappings.

hartig (changes requested):
Your current implementation of `applyToSolutionMapping` is incorrect (see my comment in the code). In addition to fixing this issue, use BindingBuilder instead of BindingFactory, as suggested by @scferrada

SHSenpai:
Thanks to both of you for your replies!
I've overhauled the applyToSolutionMapping more or less from scratch to use BindingBuilder as well as make the bindings themselves be correct.

For a given solution mapping { ?x -> ex:Bob, ?y -> ex:Ali, ?z -> ex:Charles }, it will first create the combinations
{ ?x -> ex:Robert }
{ ?x -> ex:Bobby}
when the entity mapping maps ex:Bob to ex:Robert and ex:Bobby.

It will then iterate over them, adding Ali to each when the entity mapping only maps ex:Ali to ex:Alibaba.
{ ?x -> ex:Robert, ?y -> ex:Alibaba }
{ ?x -> ex:Bobby, ?y -> ex:Alibaba}

It will then iterate over them, multiplying each for each local node that ex:Charles is mapped to.
{ ?x -> ex:Robert, ?y -> ex:Alibaba, ?z -> ex:Charlie }
{ ?x -> ex:Robert, ?y -> ex:Alibaba, ?z -> ex:Karl }
{ ?x -> ex:Robert, ?y -> ex:Alibaba, ?z -> ex:Carolus}
{ ?x -> ex:Bobby, ?y -> ex:Alibaba, ?z -> ex:Charlie }
{ ?x -> ex:Bobby, ?y -> ex:Alibaba, ?z -> ex:Karl }
{ ?x -> ex:Bobby, ?y -> ex:Alibaba, ?z -> ex:Carolus }
when the entity maps ex:Charles to ex:Charlie, ex:Karl, and ex:Carolus.

Once the cartesian product has been created, the number of possibilities are iterated over and a BindingBuilder creates a binding for each of them. Each binding is then made into a solution mapping and these are then returned as a set together.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,26 +30,152 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
 
 	@Override
 	public Set<SolutionMapping> applyToSolutionMapping( final SolutionMapping sm ) {
-		// TODO: implement this function
-		return Collections.singleton(sm);
+		final Set<Map<Var,Node>> cartesianProduct = new HashSet<>();
+		final Binding binding = sm.asJenaBinding();
+		final Iterator<Var> it = binding.vars();
+		
+		final Var firstVar = it.next(); // Create a number of single var-node combinations to populate the set with.
```
hartig, line 114: There is potential for a bug here. This code assumes that the given solution mapping `sm` is defined for at least one variable. However, `sm` may also be the empty solution mapping. This case needs to be handled:
```suggestion
		final Binding binding = sm.asJenaBinding();

		// If the given solution mapping is the empty mapping, simply return it unchanged.
		if ( binding.isEmpty() ) {
			return Collections.singleton(sm);
		}

		final Set<Map<Var,Node>> cartesianProduct = new HashSet<>();
		final Iterator<Var> it = binding.vars();
		
		final Var firstVar = it.next(); // Create a number of single var-node combinations to populate the set with.
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,26 +30,152 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+					final Map<Var,Node> newMap = new HashMap<>();
+					newMap.put(firstVar,node);
+					cartesianProduct.add(newMap);
+				}
+			}
+		} else {
+			final Map<Var,Node> newMap = new HashMap<>();
+			newMap.put(firstVar,firstNode);
+			cartesianProduct.add(newMap);
+		}
```
hartig, line 133: This part is almost the same as the code inside the following while loop. An easy way to avoid such code duplication is to simply initialize `cartesianProduct` to contain a single **empty** `Map`. Then, you can go directly to the while loop without any need for treating the first variable-node pair differently.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -25,26 +30,152 @@ public EntityMappingImpl( final Graph mappingDescription ) {
...
+					final Set<Map<Var,Node>> newCartesianProducts = new HashSet<>();
+					for(Map<Var,Node> map : cartesianProduct) {
+						final Iterator<Node> nodeIt = mappedNodes.iterator();
+						final Node newNode = nodeIt.next(); // Add the first of the combinations to each map.
+						map.put(var, newNode);
+						
+						while (nodeIt.hasNext()) { // Add the rest of the multiplications from 2 thru n, if such exists.
+							final Node anotherNewNode = it.next();
+							final Map<Var,Node> newMap = new HashMap<>();
+							newMap.putAll(map);
```
hartig, line 154: Better do the following, which achieves.
```suggestion
							final Map<Var,Node> newMap = new HashMap<>(map);
```

hartig (changes requested):
Now it is generally correct. However, there are a few things to simplify this code.

SHSenpai:
Thanks for the pointers. I've replaced the first-var special case in the method, as well as copied it over to the inverse method - with global to local switched around to local to global.

I'm going to be creating some tests, and I think this can be done in parallel with starting on SchemaMappingImpl. I'm a little bit confused as to what exactly the SchemaMapping entails. SchemaMapping's "applyToSolutionMapping" mentions "this entity mapping". Is SchemaMapping dependent on EntityMapping?

hartig (changes requested):
Generally, `applyToSolutionMapping` and `applyInverseToSolutionMapping` look good now. However, the code of these two functions is essentially identical except for the one line in which one of them uses `g2lMap` and the other one uses `l2gMap`, right? In order to make this aspect more explicit and, also, to avoid code duplication, I suggest you move the code into a helper function that both of these functions call by passing their respective Map to it.

Having a few simple unit tests for these functions (at least for one of the two) would indeed be helpful.

hartig:
Another suggestions regarding the unit tests: To make it easier for you to test exactly these functions I suggest you add another constructor via which you can directly pass `g2lMap` and `l2gMap`. That way, you are not overloading the tests with also testing `parseMappingDescription` within the same test.

SHSenpai:
@hartig 
All fixed regarding the helper function and constructor. I've also created a test class which tests the implementation using the cartesian product example I gave in yesterday's comment.

As a proposal for an eventual test for parseMappingDescription, I would create a getter-method to get the g2lMap and l2gMap from the EntityMappingImpl, and then test that parseMappingDescription produces g2lMap and l2gMap correctly, as the test I made tonight already tests that applyToSolutionMapping works given a correct g2lMap.
I don't think that this would take long, but I also don't want to do it tonight as it is 2:11 AM as of me writing this. With your permission, I would like to receive instructions for SchemaMappingImpl so that I can begin work on it in the same 'session' as finishing the tests for EntityMappingImpl.

src/test/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+		// Create cartesian product
+		final SolutionMapping robertCharlie = SolutionMappingUtils.createSolutionMapping(x, robert, y, alibaba, z, charlie);
+		final SolutionMapping robertCarolus = SolutionMappingUtils.createSolutionMapping(x, robert, y, alibaba, z, carolus);
+		final SolutionMapping robertKarl = SolutionMappingUtils.createSolutionMapping(x, robert, y, alibaba, z, karl);
+		final SolutionMapping bobbyCharlie = SolutionMappingUtils.createSolutionMapping(x, bobby, y, alibaba, z, charlie);
+		final SolutionMapping bobbyCarolus = SolutionMappingUtils.createSolutionMapping(x, bobby, y, alibaba, z, carolus);
+		final SolutionMapping bobbyKarl = SolutionMappingUtils.createSolutionMapping(x, bobby, y, alibaba, z, karl);
+	
+		// See to it that the result set contains the cartesian product and only the cartesian product.
+		assertEquals(resultSet.size(),6);
```
hartig, line 71: The expected value comes first in `assertEquals`. With the order switched, the error message that this function prints in case the test fails would be confusing.
```suggestion
		assertEquals( 6, resultSet.size() );
```

src/test/java/se/liu/ida/hefquin/engine/data/mappings/impl/EntityMappingImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.jena.sparql.core.Var;
+import org.junit.Test;
+
+import se.liu.ida.hefquin.engine.data.SolutionMapping;
+import se.liu.ida.hefquin.engine.data.mappings.EntityMapping;
+import se.liu.ida.hefquin.engine.data.utils.SolutionMappingUtils;
+
+public class EntityMappingImplTest {
+	@Test
+	public void TranslateCartesianProduct() {
```
hartig, line 22: Function names should always start with a lower case letter (i.e., `translateCartesianProduct()` instead of `TranslateCartesianProduct()`). Additionally, for functions that represent unit tests that test a specific function of the corresponding class (such as the `applyToSolutionMapping` function in this case), the name of the unit test function should reflect what function is tested.
```suggestion
	public void applyToSolutionMappingTest() {
```

hartig (changes requested):
Now it looks good. Just two very minor things.

Regarding the test for `parseMappingDescription`, please don't add the getter-methods directly into `EntityMappingImpl`. Instead, create a subclass class of `EntityMappingImpl` within the class that contains the unit tests (i.e., the class `EntityMappingImplTest`) and, then, add the getter-methods into that subclass.

SHSenpai:
I've created the class extension, but regarding the test I'm unsure what implementation of Graph to use as mappingDescription for it. There are a lot of them and none of them seem like a clear choice.

hartig:
> I've created the class extension, but regarding the test I'm unsure what implementation of Graph to use as mappingDescription for it.

Take a look at lines 335-346 in `VocabularyMappingTest`.

SHSenpai:
ParseMappingDescriptionTest has now been implemented, and it holds. Interesting to see how, at least for TURTLE, it 'unfolds' the ex:-terms into its full URI. Because I believe EntityMappingImpl and its tests have now been successfully implemented, I will merge the pull request so that I can do SchemaMappingImpl as a branch from main with EntityMappingImpl available as a reference.

hartig:
@SHSenpai in the future, please let someone else (e.g., me) have the final say whether a PR is ready to be merged.

SHSenpai:
@hartig Sorry, I thought it was okay since I recalled you telling me during that it was fine for me to merge after fixing minor issues but forgot that it had to be stated explicitly.

PR #221, opened 2022-08-05
Left Join Tests

SHSenpai:
twoOptsWithNoJoinPartners(): No matching join-values in either.
twoOptsWithOneJoinPartnerInBoth(): Base test created by Olaf, renamed to be consistent with the other names.
twoOptsWithOneJoinPartnerInOne(): Only one of the two OPTIONAL has a matching join-value.
twoOptsWithOneJoinPartnerInOneMirrored(): Same as the above, but it is the other one.
twoOptsWithIdenticalNonOptionals(): Contains multiple solution mappings with the same join-value as an input, as well as one mapping with a join-value which finds no matches in the OPTIONALs.
twoOptsWithMultiplePartnersInBoth(): Multiple triples in both OPTIONAL clauses have matching join-values.
twoOptsWithMultiplePartnersInBothAndMultipleNonOptionals(): The most complicated and thorough test. Multiple triples in both OPTIONAL clauses have matching join-values. There are several solution mappings as input, all with different join-values, all of which match different triples from the OPTIONAL clauses but not all of whom has any matches.
twoOptsWithMultiplePartnersInOne(): Multiple triples in one of the OPTIONAL clauses match.
twoOptsWithMultiplePartnersInOneMirrored(): Same as above, but it is the other OPTIONAL clause.

### DIFF
--- a/src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
+++ b/src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
@@ -60,6 +60,20 @@ public abstract class EngineTestBase
 	public static boolean skipLocalGraphQLTests = true;
 
 
+	/**
+	 * Asserts that the solution mappings being iterated over are exactly the same, in any order, as the ones contained within the list.
+	 * All solution mappings in the list must be found in the iterator, to the same amount.
+	 */
+	protected void assertFullComposition( final Iterator<SolutionMapping> it,
+										  final List<SolutionMapping> inputList)
+	{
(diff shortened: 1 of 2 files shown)


hartig:
Simon, just for your information, I looked at your `assertFullComposition` function to make sure it is okay before you start using it in your recent unit test, and it is okay. Thanks! However, I improved it a bit to provide more detailed assertion error messages. See commit https://github.com/LiUSemWeb/HeFQUIN/pull/221/commits/9971e7cc42d93545fe7e05b39bac4025982ecf65 (and since this is a new commit in your branch, you need to do `git pull` before you continue working in this branch)

PR #234, opened 2022-09-01
CLI arguments for declaring TPF and brTPF servers

hartig:
@chengsijin0817 here it is, both for TPF and brTPF ;-)


chengsijin0817:
Thanks :-)

PR #235, opened 2022-09-02
RW and rewriteReqOf

SHSenpai:
Since these are algorithms that work on operators and not operators in themselves, I found it fitting to add them to LogicalOpUtils as both their inputs and outputs are logical ops.

In HeFQUIN, the class 'LogicalOpRequest' which I assume is the request operator has in addition to a federation member also a 'reqtype'. Is the ExpectedVariables in ReqType what I should be looking at when seeing what sort of pattern the Request is requesting?

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
@@ -97,12 +97,13 @@ public class LogicalOpUtils {
 		}
 		
 		if(P instanceof SPARQLUnionPattern) {
-			final LogicalOpMultiwayUnion newRoot = LogicalOpMultiwayUnion.getInstance();
-			final List<LogicalPlan> subPlans = new ArrayList<LogicalPlan>();
+			final List<LogicalPlan> subPlans = new ArrayList<>();
 			for ( final SPARQLGraphPattern subP : ((SPARQLUnionPattern) P).getSubPatterns() ) {
 				final LogicalPlan subPlan = rewriteReqOf(subP,fm);
 				subPlans.add(subPlan);
 			}
+
+			final LogicalOpMultiwayUnion newRoot = LogicalOpMultiwayUnion.getInstance();
 			final LogicalPlan newPlan = new LogicalPlanWithNaryRootImpl(newRoot, subPlans);
 			return newPlan;
 		}
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
@@ -110,12 +110,13 @@ public class LogicalOpUtils {
 		}
 		
 		if(P instanceof SPARQLGroupPattern) {
-			final LogicalOpMultiwayJoin newRoot = LogicalOpMultiwayJoin.getInstance();
-			final List<LogicalPlan> subPlans = new ArrayList<LogicalPlan>();
+			final List<LogicalPlan> subPlans = new ArrayList<>();
 			for ( final SPARQLGraphPattern subP : ((SPARQLGroupPattern) P).getSubPatterns() ) {
 				final LogicalPlan subPlan = rewriteReqOf(subP,fm);
 				subPlans.add(subPlan);
 			}
+
+			final LogicalOpMultiwayJoin newRoot = LogicalOpMultiwayJoin.getInstance();
 			final LogicalPlan newPlan = new LogicalPlanWithNaryRootImpl(newRoot, subPlans);
 			return newPlan;
 		}
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
@@ -2,6 +2,8 @@ package se.liu.ida.hefquin.engine.queryplan.utils;
 
 import org.apache.jena.sparql.core.BasicPattern;
 import org.apache.jena.sparql.syntax.*;
+
+import se.liu.ida.hefquin.engine.data.VocabularyMapping;
 import se.liu.ida.hefquin.engine.federation.FederationMember;
 import se.liu.ida.hefquin.engine.federation.access.BGPRequest;
 import se.liu.ida.hefquin.engine.federation.access.DataRetrievalRequest;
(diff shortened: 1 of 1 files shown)


hartig:
In your thesis draft, both of these functions that you want to implement now are returning a FedQPL expression and not just a single operator. In the same sense, both of these functions that you are adding here should return a whole `LogicalPlan` not just a `LogicalOperator`.

> In HeFQUIN, the class 'LogicalOpRequest' which I assume is the request operator [...]

Correct.

> [...] has in addition to a federation member also a 'reqtype'. Is the ExpectedVariables in ReqType what I should be looking at when seeing what sort of pattern the Request is requesting?

No. `ReqType` can be any form of `DataRetrievalRequest`, including any of its sub-interfaces (e.g., `SPARQLRequest` or `TriplePatternRequest`). You can use `instanceof` to check which one it is, then cast to this sub-interface and access the actual pattern from the resulting object.  (Let me know if you want me to be more specific about these things mentioned in the last sentence here.)

SHSenpai:
@hartig 
I've pushed up changes to the code, and I have a few questions:

I currently have RW take as its argument a LogicalOpRequest. Should I instead have it take a LogicalPlan and make sure what comes out of the getRootOperator()-method is a LogicalOpRequest?

Is the way I cast to a TriplePatternRequest correct?

Are all interfaces which support BGP requests SPARQL endpoints? The thesis is written with the assumption that there are TPF-servers and SPARQL endpoints, where TPF-servers do not support BGP requests, so I am currently using the supportsBGPRequests() method as a test for whether fm is a TPF-server or a SPARQL endpoint.

I assume LogicalOPMultiwayJoin and LogicalOPMultiwayUnion is what I should use for mj and mu. Given that I have phis 1 through n, how do I create a multiway join with them? I assume the return of the method should be a logical plan with N-ary root.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
```diff
@@ -2,31 +2,89 @@
...
+		final SPARQLGraphPattern newP = vm.translateTriplePattern(tp);
+		
+		return rewriteReqOf(newP, fm);
+		
+	}
+	
+	public static LogicalPlan rewriteReqOf(final SPARQLGraphPattern P, final FederationMember fm) {
+		if (fm.getInterface().supportsBGPRequests()) { // Are all interfaces which support BGP requests SPARQL endpoints? Based on the assumption that there are two types of interfaces: TPF-server and SPARQL-endpoint, and TPF-servers do not.
+			final SPARQLRequest reqP = new SPARQLRequestImpl(P);
+			final LogicalOpRequest<SPARQLRequest, FederationMember> req = new LogicalOpRequest<SPARQLRequest, FederationMember>(fm,reqP);
```
hartig, line 58: If you are checking for support of BGP requests, you then should not create a SPARQL request but, instead, a BGP request (see the `BGPRequest` interface and the implementing class, `BGPRequestImpl`).

However, thinking about this again now, as you know, the result of the `translateTriplePattern` function may not be only a triple pattern or a BGP, but also some other type of graph pattern. Consequently, what you are doing here inside the if-block is correct, but the if-condition is wrong. The condition should actually _**not**_ be whether the interface supports BGP requests but whether the interface is a SPARQL endpoint (because only this type of interface can handle arbitrary graph patterns).

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
```diff
@@ -2,31 +2,89 @@
...
+		
+		if(P instanceof SPARQLUnionPattern) {
+			// Multiunion
+		}
+		
+		if(P instanceof SPARQLGroupPattern) {
+			// Multijoin
+		}
+		
+		return null; // Return statement. When this algorithm is finished, an error should be thrown if it for some reason gets here.
```
hartig, line 86:
```suggestion
		throw new IllegalArgumentException( P.getClass().getName() );
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
```diff
@@ -2,31 +2,89 @@
...
+		if (fm.getInterface().supportsBGPRequests()) { // Are all interfaces which support BGP requests SPARQL endpoints? Based on the assumption that there are two types of interfaces: TPF-server and SPARQL-endpoint, and TPF-servers do not.
+			final SPARQLRequest reqP = new SPARQLRequestImpl(P);
+			final LogicalOpRequest<SPARQLRequest, FederationMember> req = new LogicalOpRequest<SPARQLRequest, FederationMember>(fm,reqP);
+			final LogicalPlan newPlan = new LogicalPlanWithNullaryRootImpl(req);
+			return newPlan;
+		} // If not, continue.
+		
+		if(P instanceof TriplePattern){
+			final SPARQLRequest reqP = new SPARQLRequestImpl(P);
+			final LogicalOpRequest<SPARQLRequest, FederationMember> req = new LogicalOpRequest<SPARQLRequest, FederationMember>(fm,reqP);
```
hartig, line 65: No. Instead of a `SPARQLRequest`, create a `TriplePatternRequest` here.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
```diff
@@ -2,31 +2,89 @@
...
+			final LogicalPlan newPlan = new LogicalPlanWithNullaryRootImpl(req);
+			return newPlan;
+		}
+		
+		if(P instanceof BGP) {
+			// Create something regarding multijoin
+			for (TriplePattern tp : ((BGP) P).getTriplePatterns() ) {
```
hartig, line 72:
```suggestion
			for ( final TriplePattern tp : ((BGP) P).getTriplePatterns() ) {
```

hartig (commented):
Here are some comments about the current version of the code.

SHSenpai:
> Yes. The class that you need that represents such plans is called `LogicalPlanWithNaryRootImpl`. If you look at the constructor of that class, you need to pass a `NaryLogicalOp` (such a `LogicalOPMultiwayJoin` or a `LogicalOPMultiwayUnion`) and a list of `LogicalPlan` objects that represent your "phis 1 through n."

How do I construct a LogicalOpMultiwayJoin of requests or a LogicalOpMultiwayJoin of phis?

hartig:
> How do I construct a LogicalOpMultiwayJoin of requests or a LogicalOpMultiwayJoin of phis?

`LogicalOpMultiwayJoin.getInstance()` and `LogicalOpMultiwayUnion.getInstance()`

SHSenpai:
@hartig 
Thanks for the pointer, it took me a little bit to realize that it wasn't the operator that should be constructed with the subplans but rather the plan object itself.
I believe that this portion of the implementation is - barring possible minor adjustments needed - complete. The UP and GP rewriting have been implemented to the specifics of how it was defined in the paper.

hartig (changes requested):
Just a few minor things. See my comments.

SHSenpai:
@hartig 
It now checks to see if the fm's interface supports BGP requests for the BGP case, and I've made it so that it checks if it supports tp-requests in the other cases. This should be a little bit of future-insurance, in case we have a future interface which doesn't support tp-requests, as is an implied possibility given your comments.

hartig (approved):

PR #236, opened 2022-09-06
Unwind support

scferrada:
This PR adds the `UnwindIterator` interface and its implementation. The interface `CypherMatchQuery` now provides a method to obtain its iterators.


hartig (commented):
Looks all good. Only one point:
Shouldn't the combine function also combine the UNWIND clauses of the two match queries? Or do you consider this as a TODO for another PR? Similarly, the `CypherQueryBuilder` is also incomplete now. Supporting the UNWIND clauses in that one shouldn't be too difficult or controversial. So, you may still do it in this PR if you want.

scferrada:
I added support for UNWIND to the `CypherQueryBuilder`. The combine function I want to leave for a future PR. After this PR here, I will make a new one regarding translation of triple patterns according to the latest rules, and then a third PR regarding BGP translation and thus, the latest combine function.

hartig:
Okay. Sounds good.

PR #237, opened 2022-09-07
GraphQLEndpoint initializer now checks that types have id

WiktorKarlsson:
Necessary fix that makes sure no types are initialized in GraphQLEndpoint initializer unless they have the ID field. Removes initialized fields if they would lead to such a removed type.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/federation/GraphQLEndpoint.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/federation/GraphQLEndpoint.java
@@ -59,8 +59,4 @@ public interface GraphQLEndpoint extends FederationMember {
      */
     public GraphQLEntrypoint getEntrypoint(final String objectTypeName, final GraphQLEntrypointType fieldType);
 
-    /*
-     * Uses System.out to print information about the endpoint such as GraphQL object types and their fields.
-     */
-    public void printInformation();
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/federation/GraphQLEndpoint.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/federation/GraphQLEndpoint.java
@@ -58,4 +58,9 @@ public interface GraphQLEndpoint extends FederationMember {
      * If the parameter names provided doesn't correspond to an object type or a GraphQLEntrypointType for the endpoint return null.
      */
     public GraphQLEntrypoint getEntrypoint(final String objectTypeName, final GraphQLEntrypointType fieldType);
+
+    /*
+     * Uses System.out to print information about the endpoint such as GraphQL object types and their fields.
+     */
+    public void printInformation();
 }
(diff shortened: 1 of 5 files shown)


src/main/java/se/liu/ida/hefquin/engine/federation/GraphQLEndpoint.java
```diff
@@ -58,4 +58,9 @@ public interface GraphQLEndpoint extends FederationMember {
      * If the parameter names provided doesn't correspond to an object type or a GraphQLEntrypointType for the endpoint return null.
      */
     public GraphQLEntrypoint getEntrypoint(final String objectTypeName, final GraphQLEntrypointType fieldType);
+
+    /*
+     * Uses System.out to print information about the endpoint such as GraphQL object types and their fields.
+     */
+    public void printInformation();
```
scferrada, line 65: Is this method really necessary? Usually this kinds of systems do not use the `System.out.print` methods to log information. If you want this info for debugging or to generate files with, I would recommend to implement the `toString()` method instead.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/GraphQLEndpointInitializerImpl.java
```diff
@@ -151,6 +152,67 @@ protected static GraphQLQuery getIntrospectionQuery(){
...
+
+        final Set<String> typesToRemove = new HashSet<>();
+
+        // Check if any types needs to be removed
+        for(final String type : typesAndFields.keySet()){
+            final Map<String,GraphQLField> fields = typesAndFields.get(type);
+
+            // Check if type has an id field
+            if(!fields.containsKey("id")){
+                System.out.println("Type: " + type + " is removed due to not having an id field!");
```
scferrada, line 171: Also here, are these prints really necessary or were useful for you while developing?

scferrada (changes requested):
Hi Wiktor, here's my review. I don't think any of these print utilities are necessary, so please check that.

src/main/java/se/liu/ida/hefquin/engine/wrappers/graphqlwrapper/impl/GraphQLEndpointInitializerImpl.java
WiktorKarlsson: They are not necessary. They are now removed in latest commit: [94c8ce5](https://github.com/LiUSemWeb/HeFQUIN/pull/237/commits/94c8ce59988204af9c00434a7f9acc14d5c8e053)

scferrada:
Thanks, it looks good now.

hartig:
@scferrada can I merge this PR now?

scferrada:
Yes, please.

scferrada (approved):

PR #240, opened 2022-09-12
Translations for nested triple patterns

scferrada:


hartig (changes requested):
Looks good!
Yet, can you please do the following. I see that you added a `handleTriplePattern` function to `SPARQLStar2CypherTranslatorImpl`, and there still is also the protected `translateTriplePattern` function. Can you please add a small JavaDoc comment to both of them in which you specify briefly what each of them is about.

scferrada:
Done!

hartig (approved):
Perfect. These comments are really helpful.

PR #242, opened 2022-09-14
Bgp translation

scferrada:
Adds initial support to BGP translations to Cypher. It does not support joins on the predicate position, since I need to do some refactor to be able to reproduce the function from the paper. This refactor will contain a new interface: `CypherExpression` that will replace some of the strings that are being used, specially in `UnwindIteratorImpl`.

There's initial test on leveraging boundedness properties, as well as on capturing joins on literals.


hartig (approved):
Looks good.

PR #244, opened 2022-09-23
Complete combine function

scferrada:
This PR contains the version of the combination function that allows any kinds of joins. There is just one test so far, I can add some more.

I make extensive use of the `stream` framework for map/filter/match efficiently over different sets, I think I've read somewhere here that we shouldn't really use it, so I can change this if necessary, although the code is complicated enough as it is.


hartig (approved):
Looks good. (Apologies for the delay!)

scferrada:
Please merge this one, and then I'll push one more commit on #245 afterwards. Thanks.

PR #245, opened 2022-09-27
Update convert function

scferrada:
This PR updates the convert function, which translates Cypher Records into Solution Mappings, considering the new Expression interface and the Unwind Iterators.

Furthermore, this PR includes the addition of markers, which help us to know from which subquery any given record comes from. This includes adding new markers to each sub-query during combination of Union Queries.

Probably this PR will conflict with PR #244 since it also covers the combine function, so #244 needs to be merged first.


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/impl/expression/MarkerExpression.java
```diff
@@ -0,0 +1,18 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.expression;
+
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherExpression;
+
+public class MarkerExpression extends AliasedExpression {
```
hartig, line 5: Can you please add a JavaDoc comment that explains the purpose of this type of expression.

hartig (changes requested):
Just a minor thing.

scferrada:
Hi @hartig, this PR is complete now and your comment is addressed.

hartig (approved):

PR #246, opened 2022-09-28
Removes Gamma functions

scferrada:
Removes the getVarFor and getEdgeVarsFor methods from the Variable Generator, thus changing the combination function and the translation rules.


hartig (approved):
Looks good!

PR #247, opened 2022-09-30
rewriteLogPlan

SHSenpai:
Initial commits, checkpointing for the night.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
@@ -53,9 +53,19 @@ public class LogicalOpUtils
 	public static LogicalPlan rewriteLogPlan( final LogicalPlan input) {
 		if (input.getRootOperator() instanceof LogicalOpRequest) {
 			final LogicalOpRequest request = (LogicalOpRequest) input.getRootOperator();
+			
 			if(request.getFederationMember().getVocabularyMapping() != null) { // If fm has a vocabulary mapping vm
 				final LogicalOpLocalToGlobal l2g = new LogicalOpLocalToGlobal(request.getFederationMember().getVocabularyMapping());
-				final LogicalPlan rw = rewriteToUseLocalVocabulary(request);
+				/*
+				if (request.getRequest() instanceof SPARQLRequest) {
+					final SPARQLRequest requestRequest = (SPARQLRequest) request.getRequest();
+					final LogicalPlan rw = rewriteReqOf(requestRequest.getQueryPattern(), request.getFederationMember());
+					return new LogicalPlanWithUnaryRootImpl(l2g,rw);
+				} else {
+					throw new IllegalArgumentException( "The given plan is a non-SPARQL request." );
+				}
+				*/
+				final LogicalPlan rw = rewriteReqOf(request);
 				return new LogicalPlanWithUnaryRootImpl(l2g,rw);
 			} else {
 				return new LogicalPlanWithNullaryRootImpl(request);
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
@@ -23,11 +23,9 @@ import se.liu.ida.hefquin.engine.query.impl.BGPImpl;
 import se.liu.ida.hefquin.engine.query.impl.GenericSPARQLGraphPatternImpl1;
 import se.liu.ida.hefquin.engine.queryplan.logical.LogicalOperator;
 import se.liu.ida.hefquin.engine.queryplan.logical.LogicalPlan;
-import se.liu.ida.hefquin.engine.queryplan.logical.LogicalPlanWithNaryRoot;
 import se.liu.ida.hefquin.engine.queryplan.logical.UnaryLogicalOp;
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpBGPAdd;
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpBGPOptAdd;
-import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpLocalToGlobal;
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpMultiwayJoin;
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpMultiwayUnion;
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpRequest;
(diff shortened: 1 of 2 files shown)


hartig:
@SHSenpai I have not looked into the details of your code (as I understand that you still wanted to continue working on it). However, here is one comment already: Whatever you do here will still need to be integrated into the query processing steps of HeFQUIN. Since this is about rewriting logical plans, the easiest option is to add this as one of the rewritings that the logical optimizer does. To this end, you need to create a class that implements the `HeuristicForLogicalOptimization` interface. Call this class `ApplyVocabularyMappings` and place it in the Java package `se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics` (where you already find another class that implements the `HeuristicForLogicalOptimization` interface). Once you have created this new class, you can register it to be used by the logical optimizer by adding it in the constructor of `LogicalOptimizerImpl`

I think your current code can be moved into the `apply` function that you have to implement in this new class.

SHSenpai:
I've moved the code into the new ApplyVocabularyMappings-class.

As related to the email exchange: I've checkpointed my current work for tonight to see on the moving of code from Juliette's class into the Utils-class. It was good that you brought my attention to the existence of ExecOpRequestSPARQLWithTranslation, because it has some code which does the same thing that parts of the code I was about to write does, and is in line with the assumptions made for the thesis.

For ApplyVocabularyMapping itself, all the code of rewriteLogPlan (the 'apply' function) is ready for your review other than the case of the expression being a request operator with a vocabulary mapping.
There is a part of the code, namely line 27, where I am not sufficiently familiar with the intricacies of Java to see how to best handle this. I'm quite insecure on polymorphism in general for Java, but I believe that the casts on lines 48 and 58 are 'safe', although there might be better ways to get this done.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,70 @@
...
+					return new LogicalPlanWithUnaryRootImpl(l2g,rw);
+				} else {
+					throw new IllegalArgumentException( "The given plan is a non-SPARQL request." );
+				}
+				*/
+				// final LogicalPlan rw = LogicalOpUtils.rewriteReqOf(request) // , to call after the relevant function has been fixed
+				//return new LogicalPlanWithUnaryRootImpl(l2g,rw);
+				return inputPlan; // Temporary return for error suppression.s
+			} else {
+				return new LogicalPlanWithNullaryRootImpl(request);
```
hartig, line 44: No need to create a new object here. Just return the given input plan.
```suggestion
				return inputPlan;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,70 @@
...
+import se.liu.ida.hefquin.engine.queryplan.utils.LogicalOpUtils;
+
+public class ApplyVocabularyMappings implements HeuristicForLogicalOptimization {
+
+	/**
+	 * Rewrites an initial logical plan into a second plan which incorporates translations of local to global vocabulary and request-operator rewriting.
+	 * This method implements the rewriteLogPlan pseudocode of Helgesson's B.Sc thesis.
+	 */
+	@Override
+	public LogicalPlan apply(LogicalPlan inputPlan) {
```
hartig, line 25:
```suggestion
	public LogicalPlan apply( final LogicalPlan inputPlan ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,70 @@
...
+			} else {
+				return new LogicalPlanWithNullaryRootImpl(request);
+			}
+		} else if (inputPlan.getRootOperator() instanceof LogicalOpMultiwayJoin) {
+			final List<LogicalPlan> rewrittenSubplans = new ArrayList<>();
+			final Iterator<LogicalPlan> it = ((LogicalPlanWithNaryRoot) inputPlan).getSubPlans();
+			while(it.hasNext()) {
+				final LogicalPlan rewrittenSubplan = apply(it.next());
+				rewrittenSubplans.add(rewrittenSubplan);
+			}
```
hartig, line 52: Please maintain a boolean flag to record whether at least one of the subplans has actually been rewritten. **Only if** that is the case, create a new `LogicalPlanWithNaryRootImpl` object after this loop as you currently do. If none of the subplans has been rewritten, return the given input plan instead of creating a new object.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,70 @@
...
+			final LogicalOpMultiwayJoin newRoot = LogicalOpMultiwayJoin.getInstance();
+			final LogicalPlanWithNaryRootImpl newPlan = new LogicalPlanWithNaryRootImpl(newRoot,rewrittenSubplans);
```
hartig, line 54: Better do:
```suggestion
			final LogicalPlanWithNaryRootImpl newPlan = new LogicalPlanWithNaryRootImpl( (NaryLogicalOp) inputPlan.getRootOperator(), rewrittenSubplans);
```
This way, you can cover both the case of `LogicalOpMultiwayJoin` and the case of `LogicalOpMultiwayUnion` within a single else-if branch, which avoids code duplication and, thus, makes the code easier to understand.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,70 @@
...
+			final Iterator<LogicalPlan> it = ((LogicalPlanWithNaryRoot) inputPlan).getSubPlans();
+			while(it.hasNext()) {
+				final LogicalPlan rewrittenSubplan = apply(it.next());
+				rewrittenSubplans.add(rewrittenSubplan);
+			}
+			final LogicalOpMultiwayUnion newRoot = LogicalOpMultiwayUnion.getInstance();
+			final LogicalPlanWithNaryRootImpl newPlan = new LogicalPlanWithNaryRootImpl(newRoot,rewrittenSubplans);
+			return newPlan;
+		} else {
+			return inputPlan;
```
hartig, line 67: No!! Throw an exception to make sure that we will get to see an error if this function happens to be called with an unsupported type of logical plan.
```suggestion
			throw new IllegalArgumentException("The given logical plan is not supported by this function because it has a root operator of type: " + inputPlan.getRootOperator().getClass().getName() );
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,70 @@
...
+public class ApplyVocabularyMappings implements HeuristicForLogicalOptimization {
+
+	/**
+	 * Rewrites an initial logical plan into a second plan which incorporates translations of local to global vocabulary and request-operator rewriting.
+	 * This method implements the rewriteLogPlan pseudocode of Helgesson's B.Sc thesis.
+	 */
+	@Override
+	public LogicalPlan apply(LogicalPlan inputPlan) {
+		if (inputPlan.getRootOperator() instanceof LogicalOpRequest) {
+			final LogicalOpRequest request = (LogicalOpRequest) inputPlan.getRootOperator();
```
hartig, line 27: That's the line that you say you are not entirely sure about. Generally, it is fine. However, I assume that this line gives you a compiler warning, which you can avoid by better doing it in the following way.
```suggestion
			final LogicalOpRequest<?,?> request = (LogicalOpRequest<?,?>) inputPlan.getRootOperator();
```

hartig: Additionally, better name this program variable `reqOp` to make clear that this is actually a request operator rather than the request itself. The request you can then get by doing `reqOp.getRequest()`.

hartig (changes requested):
Here is my feedback on the current snapshot.

SHSenpai:
I've copied over and slightly modified the code from ExecOpRequestSPARQLWithTranslation, notably changing the type of exception from a FederationAccessException to a IllegalArgumentException because the latter doesn't need to pass a request operator to it and having an additional parameter just to be able to throw a certain exception seemed needless.

I've also made Union and Group operators into the same case and used a boolean to check if any of the subplans have been rewritten, as per your requested change.

hartig:
The idea was not that you copy over the code. Instead, as mentioned in my other email, I had already extracted the relevant parts of Juliette's ExecOpRequestSPARQLWithTranslation class and put them into a separate helper class from where you can simply call them. Again, here is the corresponding commit with which I created that helper class for you:
https://github.com/LiUSemWeb/HeFQUIN/commit/4f552299d12979c9d3fb974ab81f338e83c77dc9

SHSenpai:
Ah, I see. I thought you meant that you were thinking about doing it in the future, not that you had already done it. I didn't notice the commit because I didn't have alerts set up for the main branch.

Is there a way where I can mirror/emulate a commit made in another branch for this branch, or should we merge this branch and then create a new pull request with a fresh branch?

hartig:
I sent you an email on Sunday at 22:21 that explicitly points you to the commit.

I now have merged the main branch into the branch of this PR, which should give you access to the commit. In other words, you should have the new `VocabularyMappingUtils` class in your branch now. Please remove the copied code and use the `translateGraphPattern` function from class instead.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
```diff
@@ -36,34 +44,31 @@
...
 	 */
-	public static LogicalPlan rewriteToUseLocalVocabulary( final LogicalOpRequest<TriplePatternRequest, ?> reqOp ) {
-		final TriplePatternRequest tpReq = reqOp.getRequest();
-		final TriplePattern tp = tpReq.getQueryPattern();
-
+	public static LogicalPlan rewriteToUseLocalVocabulary( final LogicalOpRequest<?, ?> reqOp ) {
 		final FederationMember fm = reqOp.getFederationMember();
-		final VocabularyMapping vm = fm.getVocabularyMapping();
+		if (fm.getVocabularyMapping() == null) { // If no vocabulary mapping, nothing to translate.
+			return new LogicalPlanWithNullaryRootImpl(reqOp);
```
hartig, line 57: I would prefer that we are not creating new `LogicalPlan` objects in such cases. In order to be able to avoid doing so, the argument to this function here cannot be a `LogicalOpRequest` but must be a `LogicalPlan` (or a `LogicalPlanWithNullaryRoot` if you want) that is assumed to have a `LogicalOpRequest` as its root operator.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
```diff
@@ -36,34 +44,31 @@
...
+		}
+		final SPARQLRequest req = (SPARQLRequest) reqOp.getRequest();
```
hartig, line 59: There should be another if block first to check that the request is indeed a `SPARQLRequest` (with an `IllegalArgumentException` if it is not).

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
```diff
@@ -36,34 +44,31 @@
...
+		final SPARQLGraphPattern p = req.getQueryPattern();
 
-		final SPARQLGraphPattern newP = vm.translateTriplePattern(tp);
 
-		if ( newP.equals(tp) ) {
+		final SPARQLGraphPattern newP = VocabularyMappingUtils.translateGraphPattern(p, fm.getVocabularyMapping());
+		if ( newP.equals(p) ) {
 			return new LogicalPlanWithNullaryRootImpl(reqOp);
```
hartig, line 65: Also here I would prefer that we are not creating a new `LogicalPlan` object but instead return the one that would be given to this function.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalOpUtils.java
hartig: After changing the signature of this function in this way, it would perhaps be more logical to move the function directly into `ApplyVocabularyMappings`.

hartig (changes requested):
Generally okay now, but there is still something that I realize now that I would like to be changed. See my comments in the code.

SHSenpai:
I agree that it would make sense to move the method into ApplyVocabularyMapping since it now has both an input and an output that are LogicalPlans, and thus less appropriate for a utility class that works on request operators.

I decided to use LogicalPlan as the input over LogicalPlanWithNullaryRoot, because it removes the need for a cast in apply. I added appropriate input checking to make sure that the root operator is a request, which would've been necessary with or without LogicalPlanWithNullaryRoot as its input (as future nullary operators aren't guaranteed to be requests).

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,97 @@
...
+		if (inputPlan.getRootOperator() instanceof LogicalOpRequest) {
+			final LogicalOpRequest<?,?> requestOp = (LogicalOpRequest<?,?>) inputPlan.getRootOperator();
+			if(requestOp.getFederationMember().getVocabularyMapping() != null) { // If fm has a vocabulary mapping vm
+				final LogicalOpLocalToGlobal l2g = new LogicalOpLocalToGlobal(requestOp.getFederationMember().getVocabularyMapping());
+				if (requestOp.getRequest() instanceof SPARQLRequest) {
+					final LogicalPlan rw = rewriteToUseLocalVocabulary(inputPlan);
+					return new LogicalPlanWithUnaryRootImpl(l2g,rw);
+				} else {
+					throw new IllegalArgumentException( "The given plan is a non-SPARQL request." );
+				}
```
hartig, line 71:
```suggestion
				final LogicalPlan rw = rewriteToUseLocalVocabulary(inputPlan);
				return new LogicalPlanWithUnaryRootImpl(l2g,rw);
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,97 @@
...
+		final SPARQLRequest req = (SPARQLRequest) reqOp.getRequest();
+		final SPARQLGraphPattern p = req.getQueryPattern();
+
+		final SPARQLGraphPattern newP = VocabularyMappingUtils.translateGraphPattern(p, fm.getVocabularyMapping());
+		if ( newP.equals(p) ) {
+			return inputPlan;
+		}
+		else {
+			return LogicalOpUtils.rewriteReqOf(newP, fm);
+		}
```
hartig, line 52:
```suggestion
		return ( newP.equals(p) ) ? inputPlan : LogicalOpUtils.rewriteReqOf(newP, fm);
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -0,0 +1,97 @@
...
+				final LogicalPlan subPlan = it.next();
+				final LogicalPlan rewrittenSubplan = apply(subPlan);
+				rewrittenSubplans.add(rewrittenSubplan);
+				if(!subPlan.equals(rewrittenSubplan)) {
+					rewritten = true;
+				}
+			}
+			if (rewritten) {
+				final LogicalPlanWithNaryRootImpl newPlan = new LogicalPlanWithNaryRootImpl( (NaryLogicalOp) inputPlan.getRootOperator(), rewrittenSubplans);
+				return newPlan;
```
hartig, line 89:
```suggestion
				return new LogicalPlanWithNaryRootImpl( (NaryLogicalOp) inputPlan.getRootOperator(), rewrittenSubplans);
```

hartig (approved):

PR #251, opened 2022-10-26
Extend federation description

chengsijin0817:
- Extend the ExampleFederation with a property for specifying the path to vocabulary mappings
- Update the method parseFedDescr
- Add an example 'vocabularyMappings' file for testing

In the method 'parseFedDescr', only the federation members with a SPARQL endpoint are extended for considering vocabulary mappings when parsing federation descriptions, at the moment.
If it looks okay, I can use the same implementation for parsing all different types of federation members.

### DIFF
--- a/ExampleFederation.ttl
+++ b/ExampleFederation.ttl
@@ -6,9 +6,8 @@ PREFIX ex:     <http://example.org/>
 ex:dbpediaSPARQL
       a            fd:FederationMember ;
       fd:interface [ a                  fd:SPARQLEndpointInterface ;
-                     fd:endpointAddress <http://dbpedia.org/sparql> ;
-                     fd:vocabularyMappings "dbpedia/vocabularyMappings.nt"
-                     ] .
+                     fd:endpointAddress <http://dbpedia.org/sparql> ] ;
+      fd:vocabularyMappingsFile "dbpedia/vocabularyMappings.nt" .
 
 ex:dbpediaTPF
       a            fd:FederationMember ;
@@ -19,6 +18,11 @@ ex:wikidataSPARQL
       a            fd:FederationMember ;
       fd:interface [ a                  fd:SPARQLEndpointInterface ;
                      fd:endpointAddress <https://query.wikidata.org/sparql> ] .
+
+ex:europaSPARQL
+      a            fd:FederationMember ;
+      fd:interface [ a                  fd:SPARQLEndpointInterface ;
+                     fd:endpointAddress <https://data.europa.eu/sparql> ] .
 
 #ex:neo4jTest
 #      a            fd:FederationMember ;
--- a/ExampleFederation.ttl
+++ b/ExampleFederation.ttl
@@ -6,7 +6,9 @@ PREFIX ex:     <http://example.org/>
 ex:dbpediaSPARQL
       a            fd:FederationMember ;
       fd:interface [ a                  fd:SPARQLEndpointInterface ;
-                     fd:endpointAddress <http://dbpedia.org/sparql> ] .
+                     fd:endpointAddress <http://dbpedia.org/sparql> ;
+                     fd:vocabularyMappings "dbpedia/vocabularyMappings.nt"
+                     ] .
 
 ex:dbpediaTPF
       a            fd:FederationMember ;
--- a/dbpedia/vocabularyMappings.nt
+++ b/dbpedia/vocabularyMappings.nt
@@ -1,0 +1,1 @@
+<http://wikidata.org/ontology/birthPlace> <http://www.w3.org/2002/07/owl#sameAs> <http://dbpedia.org/ontology/birthPlace> .
--- a/dbpedia/vocabularyMappings.nt
+++ b/dbpedia/vocabularyMappings.nt
@@ -1,3 +1,3 @@
-<http://dbpedia.org/resource/Adam_Maher> <http://www.w3.org/2002/07/owl#sameAs> <http://data.europa.eu/euodp/jrc-names/Adam_Maher> .
-<http://dbpedia.org/ontology/birthPlace> <http://www.w3.org/2002/07/owl#equivalentProperty> <http://www.wikidata.org/entity/P19> .
-#<http://dbpedia.org/ontology/birthPlace> <http://www.w3.org/2002/07/owl#equivalentProperty> <http://schema.org/birthPlace> .
+<http://data.europa.eu/euodp/jrc-names/Adam_Maher> <http://www.w3.org/2002/07/owl#sameAs> <http://dbpedia.org/resource/Adam_Maher> .
+<http://www.wikidata.org/entity/P19> <http://www.w3.org/2002/07/owl#equivalentProperty> <http://dbpedia.org/ontology/birthPlace> .
+#<http://schema.org/birthPlace> <http://www.w3.org/2002/07/owl#equivalentProperty> <http://dbpedia.org/ontology/birthPlace> .
--- a/dbpedia/vocabularyMappings.nt
+++ b/dbpedia/vocabularyMappings.nt
@@ -1,1 +1,3 @@
-<http://wikidata.org/ontology/birthPlace> <http://www.w3.org/2002/07/owl#sameAs> <http://dbpedia.org/ontology/birthPlace> .
+<http://dbpedia.org/resource/Adam_Maher> <http://www.w3.org/2002/07/owl#sameAs> <http://data.europa.eu/euodp/jrc-names/Adam_Maher> .
+<http://dbpedia.org/ontology/birthPlace> <http://www.w3.org/2002/07/owl#equivalentProperty> <http://www.wikidata.org/entity/P19> .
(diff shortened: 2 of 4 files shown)


src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -200,6 +203,19 @@ protected void addBRTPFServers( final List<String> uris ) {
 			addBRTPFServer(uri);
 	}
 
+	protected void addSPARQLEndpoint( final String sparqlEndpointValue, final String pathToVocabularyMappings ) {
```
chengsijin0817, line 206: Here, we can keep both versions of 'addSPARQLEndpoint' method or combine them into one method and check if the 'pathToVocabularyMappings' parameter is empty. 
@hartig  Which one do you think is better?

dbpedia/vocabularyMappings.nt
```diff
@@ -0,0 +1 @@
+<http://wikidata.org/ontology/birthPlace> <http://www.w3.org/2002/07/owl#sameAs> <http://dbpedia.org/ontology/birthPlace> .
```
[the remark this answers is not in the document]
hartig: In addition to this equivalent-property relationship, maybe you can also add an actual owl:sameAs statement between a Wikidata URI and corresponding DBpedia URI. You should find plenty of them in DBpedia.

ExampleFederation.ttl
```diff
@@ -6,7 +6,9 @@ PREFIX ex:     <http://example.org/>
 ex:dbpediaSPARQL
       a            fd:FederationMember ;
       fd:interface [ a                  fd:SPARQLEndpointInterface ;
-                     fd:endpointAddress <http://dbpedia.org/sparql> ] .
+                     fd:endpointAddress <http://dbpedia.org/sparql> ;
+                     fd:vocabularyMappings "dbpedia/vocabularyMappings.nt"
```
hartig, line 10: Instead of making `fd:vocabularyMappings` a property of the interface of a federation member, I would make it a property of the federation member itself. In other words, for this particular case, `fd:vocabularyMappings` should be a property of `ex:dbpediaSPARQL`.

Additionally, I would name this property `fd:vocabularyMappingsFile` because, instead of connecting the fed.member to the actual vocab.mapping, it connects the fed.member to the file that describes the vocab.mapping.

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -105,7 +103,12 @@ else if ( addr.isURIResource() ) {
 				throw new IllegalArgumentException();
 			}
 
-			addSPARQLEndpoint(addrStr);
+			if( fd.contains(iface, FD.vocabularyMappings) ){
+				final RDFNode path = fd.getRequiredProperty(iface, FD.vocabularyMappings).getObject();
+				addSPARQLEndpoint(addrStr, path.toString());
```
hartig, line 108: Why have that only for SPARQL endpoints and not for any other type of federation member? I think that should be general. To achieve that--in combination with my other comment (about making `fd:vocabularyMappingsFile` a property of the fed.member)--you will probably have to wrap all these loops here into another loop that iterates over all fed.members. Then, you can process the `fd:vocabularyMappingsFile` property first and create a `VocabularyMapping` object from the file (and have that `VocabularyMapping` object be `null` for all the fed.members that do not have a `fd:vocabularyMappingsFile` property).

Another issue: you should explicitly check whether there is indeed a file at the given path.

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
hartig: These should have been merged because there is a lot of code duplication between them. However, if you create the `VocabularyMapping` separately (as suggested in one of my other comments), then this question here becomes irrelevant. Instead, the `addSPARQLEndpoint` function would simply be extended with a second argument for passing the `VocabularyMapping` object (or `null`). And the same would have to be done for the other `add***` functions.

hartig (changes requested):

dbpedia/vocabularyMappings.nt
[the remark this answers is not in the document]
chengsijin0817: I added two actual statements from DBpedia: 
- one owl:equivalentProperty between Wikidata URI and DBpedia, 
- one owl:sameAs between DBpedia and Europa

However, when I try to retrieve data using the matching vocabularies (resources) from Wikidata and Europa, the results are empty. I will check them again.

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -89,147 +88,151 @@ public FederationCatalog getFederationCatalog() {
...
-			final String addrStr;
-			if ( addr.isLiteral() ) {
-				addrStr = addr.asLiteral().getLexicalForm();
-			}
-			else if ( addr.isURIResource() ) {
-				addrStr = addr.asResource().getURI();
-			}
-			else {
-				throw new IllegalArgumentException();
+		final ResIterator isFedMember = fd.listResourcesWithProperty(RDF.type, FD.FederationMember);
```
hartig, line 91: Why do you call this variable `isFedMember`? In particular, what is the meaning of `is`?

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -287,4 +290,14 @@ protected URI verifyExpectedURI( final String uriString ) {
 		return uri;
 	}
 
+	protected boolean verifyValidVocabMappingFile(final String pathString ) {
+		File f = new File(pathString);
```
hartig, line 294:
```suggestion
		final File f = new File(pathString);
```

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -287,4 +290,14 @@ protected URI verifyExpectedURI( final String uriString ) {
...
+		if ( f.exists() ){
+			// TODO: .nt file?
```
hartig, line 296: No need to verify that it is an NTriples file. The Jena code that will later be used to parse this file should be able to figure out which file format it is and, then, use the corresponding parser.

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -287,4 +290,14 @@ protected URI verifyExpectedURI( final String uriString ) {
...
+			return f.isFile();
```
hartig, line 297: If it is not an actual file, you return false here. In this case, with the current implementation that calls this function, there will be no error whatsoever. However, there should be one (similar to the case in which the file/directory does not exist at all (i.e., the else branch below).

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -89,147 +88,151 @@ public FederationCatalog getFederationCatalog() {
...
-
-			final String addrStr;
-			if ( addr.isLiteral() ) {
-				addrStr = addr.asLiteral().getLexicalForm();
-			}
-			else if ( addr.isURIResource() ) {
-				addrStr = addr.asResource().getURI();
+			final Resource iface = fedMember.getProperty(FD.interface_).getResource();
+			// Check the type of interface
+			if ( fd.getRequiredProperty(iface, RDF.type).getObject().equals(FD.SPARQLEndpointInterface) ){
```
hartig, line 107: Instead of repeating `fd.getRequiredProperty(iface, RDF.type).getObject()` in all of these if statements, create a program variable for it. That will improve the readability of the code.

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -89,147 +88,151 @@ public FederationCatalog getFederationCatalog() {
...
 			}
 
-			addGraphQLServer(addrStr);
 		}
 	}
 
 	protected void addSPARQLEndpoints( final List<String> sparqlEndpointValues ) {
 		for ( final String v : sparqlEndpointValues )
-			addSPARQLEndpoint(v);
+			// TODO: extend command line to support connect the endpoint to a file that describe the vocabulary mappings
```
hartig, line 192:
```suggestion
```

No. We should move away from exposing every configuration detail via a separate CLI argument. Instead, the way to specify the federation to be queried (including corresponding vocab.mappings, if any) should be via the federation description file.

hartig (changes requested):
A few minor things remain to be improved.

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
chengsijin0817: This variable is renamed as 'fedMembers'

hartig (approved):
Looks good now. Thanks!

PR #252, opened 2022-10-29
Adds code for join rewriting of Cypher Queries

scferrada:
Hi, I wanted to have a PR with both the join rewriting and the path merging, but I had to add a Visitor for the first one since most `CypherExpression`s are recursive, so this PR is all about making explicit joins implicit.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
@@ -13,18 +13,19 @@ import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.SPARQLStar2CypherTranslator
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherMatchQuery;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.CypherQuery;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.MatchClause;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.UnwindIterator;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.CypherUnionQueryImpl;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.expression.*;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.match.EdgeMatchClause;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.match.NodeMatchClause;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.utils.CypherQueryBuilder;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.utils.CypherQueryCombinator;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.utils.CypherUtils;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.utils.CypherVarGenerator;
 
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
+import java.util.stream.Collectors;
 
 public class SPARQLStar2CypherTranslatorImpl implements SPARQLStar2CypherTranslator {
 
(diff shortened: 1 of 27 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -806,4 +807,85 @@ protected static CypherQuery getVarVarVar(final Node s, final Node p, final Node
...
+    @Override
+    public CypherMatchQuery rewriteJoins(final CypherMatchQuery query) {
+        final CypherQueryBuilder builder = new CypherQueryBuilder();
+        final List<BooleanCypherExpression> variableJoins = query.getConditions().stream()
+                .filter(x -> x instanceof EqualityExpression
+                        && ((EqualityExpression) x).getLeftExpression() instanceof CypherVar
+                        && ((EqualityExpression) x).getRightExpression() instanceof CypherVar
+                        && query.getMatchVars().contains((CypherVar) ((EqualityExpression) x).getLeftExpression())
+                        && query.getMatchVars().contains((CypherVar) ((EqualityExpression) x).getRightExpression()))
+                .collect(Collectors.toList());
```
hartig, line 820: I assume that there can be cases in which the `variableJoins` list created by this statement is empty. In such cases, there is no need to go through the rest of the code. Instead, simply return the given Match query.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -806,4 +807,85 @@ protected static CypherQuery getVarVarVar(final Node s, final Node p, final Node
...
+        final List<MatchClause> toRemove = new ArrayList<>();
+        for (final MatchClause m : matches) {
+            if (m instanceof EdgeMatchClause) {
+                for (final NodeMatchClause node : nodes) {
+                    if (m.isRedundantWith(node))
+                        toRemove.add(node);
+                }
+            }
+        }
+        matches.removeAll(toRemove);
```
hartig, line 849: Similar to my comment above, if the `nodes` list created in lines 838-839 is empty, there is no need to create the `toRemove` list and iterate over all the `matches`.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SPARQLStar2CypherTranslatorImpl.java
```diff
@@ -806,4 +807,85 @@ protected static CypherQuery getVarVarVar(final Node s, final Node p, final Node
...
+        for (final BooleanCypherExpression c : query.getConditions()) {
+            if (variableJoins.contains(c)) continue;
+            builder.add(CypherUtils.replaceVariable(equivalences, c));
+        }
+        for (final UnwindIterator i : query.getIterators()) {
+            builder.add(CypherUtils.replaceVariable(equivalences, i));
+        }
+        for (final AliasedExpression r : query.getReturnExprs()) {
+            builder.add(CypherUtils.replaceVariable(equivalences, r));
+        }
```
hartig, line 837: From a code readability perspective, I would find it more logical to move these part to after the following part which is still about the MATCH clauses.

Perhaps you can also add some comments into the code to describe what each of the next groups of statements is doing (e.g., "rewrite the MATCH clauses by replacing the variables in them", "... and by removing redundant node patterns that are covered by one of the edge patterns").

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/CypherExpression.java
```diff
@@ -16,4 +18,8 @@ public interface CypherExpression {
      */
     Set<CypherVar> getVars();
 
+    /**
+     * accepts a {@link CypherExpressionVisitor} and propagates the visit recursively when required.
+     */
+    void acceptVisitor(final CypherExpressionVisitor visitor);
```
hartig, line 24: The function name that I have used for the visitor pattern in other parts of the code base is `visit`. Unless there is a strong reason for using `acceptVisitor` here, can you please rename this function to `visit` for the sake of consistency.

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/SPARQLStar2CypherTranslatorTest.java
```diff
@@ -884,6 +885,38 @@ public void translateBGPTest() {
                 translation);
     }
 
+    @Test
+    public void translateBGPwithJoinRewriteTest() {
```
hartig, line 889: I think it would be helpful to also add a test that focuses only on the new `rewriteJoin` function; i.e., that test creates a Cypher query with equi-joins, passes that query directly to the `rewriteJoin` function, and then checks that the equi-joins have been rewritten correctly.

hartig (changes requested):
Thanks. Looks good in general, but I have a few suggestions for improvement. See the detailed in-code comments.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/CypherExpression.java
scferrada: I had forgotten we used visitors elsewhere, so I just named the methods as they made more sense to me (i.e.,  the expressions receive a visitor), but I have renamed it to be consistent.

scferrada:
I think everything is addressed now :)

PR #253, opened 2022-11-04
Merge of Path Patterns

scferrada:
This PR introduces a method for the translator object, which takes a set of `NodeMatchClause` and `EdgeMatchClause` and merges them into longer path patterns (`PathMatchPattern`).

To implement this method, all Match clauses are added into a "graph pattern", which is in class `LabeledGraph`. This class provides methods for finding the longest (simple undirected) path in the graph, and removing paths from the graph.

This PR also fixes a bug in the visitor pattern, where objects in lists need to be extracted in the reverse order they were put in the stack.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/impl/match/PathMatchClause.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/impl/match/PathMatchClause.java
@@ -105,10 +105,6 @@ public class PathMatchClause implements MatchClause {
             this.direction = direction;
         }
 
-        public LabeledGraph.Direction getDirection() {
-            return direction;
-        }
-
         @Override
         public boolean equals(Object o) {
             if (this == o) return true;
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/impl/match/PathMatchClause.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/impl/match/PathMatchClause.java
@@ -1,0 +1,84 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.impl.match;
+
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.query.MatchClause;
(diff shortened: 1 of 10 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/query/impl/match/PathMatchClause.java
```diff
@@ -0,0 +1,125 @@
...
+
+            this.left = left;
+            this.edge = edge;
+            this.right = right;
+            this.direction = direction;
+        }
+
+        public LabeledGraph.Direction getDirection() {
+            return direction;
+        }
```
hartig, line 110: I don't think you need that one.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/VariableReplacementVisitor.java
```diff
@@ -168,6 +167,21 @@ public void visitNodeMatch(final NodeMatchClause ex) {
...
+    public void visitPathMatch(PathMatchClause ex) {
+        final Deque<PathMatchClause.EdgePattern> edges = new ArrayDeque<>();
+        for (int i = ex.getEdges().size(); i > 0; i--) {
+            assert stack.peek() instanceof CypherVar;
+            final CypherVar right = (CypherVar) stack.pop();
+            assert stack.peek() instanceof CypherVar;
+            final CypherVar edge = (CypherVar) stack.pop();
+            assert stack.peek() instanceof CypherVar;
+            final CypherVar left = (CypherVar) stack.pop();
+            edges.push(new PathMatchClause.EdgePattern(left, edge, right, ex.getEdges().get(i).getDirection()));
```
hartig, line 180: After making the member variables of `PathMatchClause.EdgePattern` all public, you can change this line to:
```suggestion
            edges.push( new PathMatchClause.EdgePattern(left, edge, right, ex.getEdges().get(i).direction) );
```

hartig (changes requested):

scferrada:
I think all comments have been addressed now @hartig

hartig (approved):
Looks good now.

PR #255, opened 2022-11-23
Fix an issue in result rewriting

chengsijin0817:
Fix an issue in result rewriting, specifically for variables that are bound to RDF literals.

Here is an example query for reproducing the issue:
the execution of this query does not terminate if '[ApplyVocabularyMappings](https://github.com/LiUSemWeb/HeFQUIN/blob/b5ee241745cad089cb4a8e91ee8e6baef73ddf5c/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/LogicalOptimizerImpl.java#L23)' is enabled.
```
SELECT * WHERE {
    SERVICE <http://dbpedia.org/sparql> { 
         <http://data.europa.eu/euodp/jrc-names/Adam_Maher> <http://www.w3.org/2000/01/rdf-schema#label> ?name 
    }
}
```

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
@@ -347,36 +347,37 @@ public class VocabularyMappingImpl implements VocabularyMapping
 		while(i.hasNext()) {
 			final Var v = i.next();
 			final Node n = sm.asJenaBinding().get(v);
-			
+
 			if(!n.isURI()) {
 				for (final BindingBuilder j : bbs) {
 					j.add(v, n);
 				}
 			}
-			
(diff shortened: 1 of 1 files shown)


hartig (approved):
Looks good to me.

@JulietteVV do you approve? (notice that the only thing that is actually changed here is take everything that comes inside the while-loop after the if-block, and wrap that as an else-block)

@chengsijin0817 is this the root cause of the issue that you wanted me to confirm earlier this week?

chengsijin0817:
> 
> @chengsijin0817 is this the root cause of the issue that you wanted me to confirm earlier this week?

Yes, I think so. That issue is resolved by updating this part of the code.

hartig:
Perfect!

hartig:
@JulietteVV please confirm asap that we can merge this PR.

JulietteVV (approved):

hartig:
Thanks all for resolving this issue so quickly!

PR #256, opened 2022-11-24
ApplyVocabularyMapping can now handle plans containing filter operators.

SHSenpai:
Functionality for filter operations in the plan rewrite of ApplyVocabularyMappings. Necessary to get all the tests to run when building the package after uncommenting the lines required for the rewriting package build:
return createQueryOptimizerWithoutOptimization(ctxt); uncommented in HeFQUINEngineConfig.java
heuristics.add( new ApplyVocabularyMappings()); uncommented in LogicalOptimizerImpl.java


hartig:
Simon, thanks a lot for fixing this problem! I see, however, that your fix is suitable only for cases in which the filter conditions would not be affected by the vocabulary mappings. I guess that this is okay for the queries in your evaluation. Therefore, I am fine accepting this PR as is, but with the understanding that it is only addressing half of the problem. For the rest of the problem, I have created issue #257

chengsijin0817 (approved):

PR #265, opened 2022-12-12
ExecOpBindJoinSPARQLwithVALUES

chengsijin0817:
Consider the case where the given solution mapping does not have any join variables with the triple pattern

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUES.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUES.java
@@ -43,8 +43,16 @@ public class ExecOpBindJoinSPARQLwithVALUES extends BaseForExecOpBindJoinSPARQL
 	protected NullaryExecutableOp createExecutableRequestOperator( final Iterable<SolutionMapping> solMaps ) {
 		final Set<Binding> bindings = new HashSet<>();
 		final Set<Var> joinVars = new HashSet<>();
+
+		boolean noJoinVars = false;
 		for ( final SolutionMapping s : solMaps ) {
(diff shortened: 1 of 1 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUES.java
```diff
@@ -55,6 +63,9 @@ protected NullaryExecutableOp createExecutableRequestOperator( final Iterable<So
 			}
 		}
 
+		if (noJoinVars) {
+			return new ExecOpRequestSPARQL( new SPARQLRequestImpl(query), fm, false );
+		}
```
hartig, line 68:
```suggestion
		}

```

src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUES.java
```diff
@@ -43,8 +43,16 @@ public ExecOpBindJoinSPARQLwithVALUES( final SPARQLGraphPattern query, final SPA
...
+
+		boolean noJoinVars = false;
 		for ( final SolutionMapping s : solMaps ) {
 			final Binding b = SolutionMappingUtils.restrict( s.asJenaBinding(), varsInSubQuery );
+			// If there exists a solution mapping that does not have any variables in common with the triple pattern of this operator
+			// retrieve all matching triples of the given query
+			if ( b.isEmpty() ) {
+				noJoinVars = true;
+				break;
+			}
```
hartig, line 55:
```suggestion

			// If there exists a solution mapping that does not have any variables in common with the triple pattern of this operator
			// retrieve all matching triples of the given query
			if ( b.isEmpty() ) {
				noJoinVars = true;
				break;
			}

```

hartig (approved):

Issue #268, opened 2022-12-19
add a new logical operator called 'gpAdd'


hartig:
... and extend `createLogicalAddOpFromLogicalReqOp` of `Worker` in `LogicalOpUtils`

hartig:
Addressed by PR #270

PR #270, opened 2022-12-19
Support for gpAdd and gpOptAdd

hartig:
This PR is meant to address #268

Currently, it is still work in progress. Here is a list of things that I still need to do:

- [x] For all subclasses of `BasePhysicalOpSingleInputJoin`, look at their corresponding executable operators and make sure that they support arbitrary `SPARQLGraphPattern` objects.
- [x] Is there anything to do for `LogicalToPhysicalPlanConverterImpl`?
- [x] Extend `PhysicalPlanFactory` (e.g., `enumeratePlansWithUnaryOpFromReq`).
- [x] Extend `UnionPullUp`.
- [x] Extend `MergeRequests`.
- [x] Extend `CardinalityEstimationImpl`.
- [x] Extend `VarSpecificCardinalityEstimationImpl`.
- [x] Extend the classes in `se.liu.ida.hefquin.engine.queryproc.impl.poptimizer.costmodel`.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
@@ -8,6 +8,7 @@ import se.liu.ida.hefquin.engine.federation.SPARQLEndpoint;
 import se.liu.ida.hefquin.engine.federation.TPFServer;
 import se.liu.ida.hefquin.engine.federation.access.DataRetrievalRequest;
 import se.liu.ida.hefquin.engine.federation.access.impl.req.BGPRequestImpl;
+import se.liu.ida.hefquin.engine.federation.access.impl.req.SPARQLRequestImpl;
 import se.liu.ida.hefquin.engine.federation.access.impl.req.TPFRequestImpl;
 import se.liu.ida.hefquin.engine.federation.access.impl.req.TriplePatternRequestImpl;
 import se.liu.ida.hefquin.engine.query.TriplePattern;
@@ -498,6 +499,20 @@ public class PhysicalPlanFactory
 		final DataRetrievalRequest req;
 		if ( fm.getInterface().supportsBGPRequests() ) {
 			req = new BGPRequestImpl( lop.getBGP() );
+		}
+		else {
+			throw new IllegalArgumentException("Unsupported type of federation member (type: " + fm.getClass().getName() + ").");
+		}
+
+		return createPlanWithRequest( new LogicalOpRequest<>(fm,req) );
+	}
+
+	public static PhysicalPlan extractRequestAsPlan( final LogicalOpGPAdd lop ) {
+		final FederationMember fm = lop.getFederationMember();
+
+		final DataRetrievalRequest req;
+		if ( fm.getInterface().supportsSPARQLPatternRequests() ) {
+			req = new SPARQLRequestImpl( lop.getPattern() );
 		}
 		else {
 			throw new IllegalArgumentException("Unsupported type of federation member (type: " + fm.getClass().getName() + ").");
--- a/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
@@ -541,12 +541,15 @@ public class PhysicalPlanFactory
 		return extractRequestAsPlan( pop.getLogicalOperator() );
 	}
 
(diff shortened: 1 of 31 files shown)


hartig:
@chengsijin0817 all done. Please take a look at the changes. In particular, please double check that the extensions of the classes in `se.liu.ida.hefquin.engine.queryproc.impl.poptimizer.costmodel` correspond to your intention of how the cost model should be extended to cover the new operator. Once you approve this PR, I will merge it.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
```diff
@@ -485,52 +556,101 @@ else if ( lop instanceof LogicalOpBGPAdd ) {
 	public static List<PhysicalPlan> enumeratePlansWithUnaryOpFromReq( final PhysicalOpRequest<?, ?> req,
```
chengsijin0817, line 556: Line 551: the operator LogicalOpGPAdd needs to be considered in this function 
```
public static PhysicalPlan extractRequestAsPlan(final UnaryLogicalOp lop) {
...
else if ( lop instanceof LogicalOpGPAdd ) {
	return extractRequestAsPlan( (LogicalOpGPAdd) lop );
}
...
```

Then, the implementation of 'initiateCostEstimation' in CFRNumberOfTermsShippedInRequests can be simplified.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/costmodel/CFRNumberOfTermsShippedInRequests.java
```diff
@@ -51,6 +51,16 @@ else if ( lop instanceof LogicalOpBGPAdd ) {
...
+		else if ( lop instanceof LogicalOpGPAdd ) {
+			final LogicalOpGPAdd gpAdd = (LogicalOpGPAdd) lop;
+			numberOfTerms = QueryPatternUtils.getNumberOfTermOccurrences( gpAdd.getPattern() );
+
+			final PhysicalPlan subplan = plan.getSubPlan(0);
+			final PhysicalPlan reqGP = PhysicalPlanFactory.extractRequestAsPlan(gpAdd);
+			numberOfJoinVars = ExpectedVariablesUtils.intersectionOfCertainVariables(subplan,reqGP).size();
+
+			futureIntResSize = initiateCardinalityEstimation(subplan);
+		}
```
chengsijin0817, line 63: After updating the helper function 'extractRequestAsPlan' in the class PhysicalPlanFactory, the implementation from line 34 to line 63 can be updated as:
```
if ( lop instanceof UnaryLogicalOp ) {
	SPARQLGraphPattern pattern = null;
	if (lop instanceof LogicalOpTPAdd) {
		pattern = ((LogicalOpTPAdd) lop).getTP();

	} else if (lop instanceof LogicalOpBGPAdd) {
		pattern = ((LogicalOpBGPAdd) lop).getBGP();

	} else if (lop instanceof LogicalOpGPAdd) {
		pattern = ((LogicalOpGPAdd) lop).getPattern();
	}

	numberOfTerms = QueryPatternUtils.getNumberOfTermOccurrences( pattern );

	final PhysicalPlan subplan = plan.getSubPlan(0);
	final PhysicalPlan req = PhysicalPlanFactory.extractRequestAsPlan( (UnaryLogicalOp) lop );
	numberOfJoinVars = ExpectedVariablesUtils.intersectionOfCertainVariables(subplan, req).size();

	futureIntResSize = initiateCardinalityEstimation(subplan);
}
```

chengsijin0817 (commented):
Hi @hartig ,Thank you. I just have one small comment in class 'PhysicalPlanFactory'. 

In addition, I observed some duplicated code when estimating the number of terms and variables for plans with 'LogicalOpBGPAdd' or 'LogicalOpGPAdd' as root operators.
For this case, I tried to simplify the implementation of class 'CFRNumberOfTermsShippedInRequests', can you please check whether it would reduce readability?

hartig:
Thanks for checking and for your suggestion, which makes sense. Can you please push a commit with these two changes to the branch of this PR. When doing so, do not forget to add an else-branch in your proposed simplification to CFRNumberOfTermsShippedInRequests where you throw an exception in case of other UnaryLogicalOp.

chengsijin0817:
> Thanks for checking and for your suggestion, which makes sense. Can you please push a commit with these two changes to the branch of this PR. When doing so, do not forget to add an else-branch in your proposed simplification to CFRNumberOfTermsShippedInRequests where you throw an exception in case of other UnaryLogicalOp.

Done. Please check it ;-) @hartig

hartig:
Looks good. Thanks for the help. I will merge this PR now.

PR #275, opened 2023-01-25
Add a cmd argument: 'printQueryProcTime'

chengsijin0817:
'printQueryProcTime': print out statistics about the query processing time in one line

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
+++ b/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
@@ -169,7 +169,7 @@ public class RunQueryWithoutSrcSel extends CmdARQ
 				final long planningTime = statsAndExceptions.object1.getPlanningTime();
 				final long compilationTime = statsAndExceptions.object1.getCompilationTime();
 				final long executionTime = statsAndExceptions.object1.getExecutionTime();
-				final String queryProcStats = "Overall Query Processing Time, " + overallQueryProcessingTime + ", Query Planning Time, " + planningTime +", Compile Time, " + compilationTime + ", Execution time, " + executionTime + "\n";
+				final String queryProcStats = overallQueryProcessingTime + ", " + planningTime +", " + compilationTime + ", " + executionTime + "\n";
 				System.out.println(queryProcStats);
 			}
 		}
--- a/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
+++ b/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
@@ -169,7 +169,7 @@ public class RunQueryWithoutSrcSel extends CmdARQ
 				final long planningTime = statsAndExceptions.object1.getPlanningTime();
 				final long compilationTime = statsAndExceptions.object1.getCompilationTime();
 				final long executionTime = statsAndExceptions.object1.getExecutionTime();
-				final String queryProcStats = "Overall Query Processing Time: " + overallQueryProcessingTime + " - Query Planning Time: " + planningTime +" - Compile Time: " + compilationTime + " - Execution time: " + executionTime;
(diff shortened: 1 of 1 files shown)


src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
```diff
@@ -158,9 +160,18 @@ protected void exec() {
...
+		     && statsAndExceptions.object1 != null) {
+			if ( contains(argQueryProcStats) ) {
+				StatsPrinter.print(statsAndExceptions.object1, System.err, true);
+			}
+			if ( contains(argOnelineTimeStats) ) {
+				final long overallQueryProcessingTime = statsAndExceptions.object1.getOverallQueryProcessingTime();
+				final long planningTime = statsAndExceptions.object1.getPlanningTime();
+				final long compilationTime = statsAndExceptions.object1.getCompilationTime();
+				final long executionTime = statsAndExceptions.object1.getExecutionTime();
+				final String queryProcStats = "Overall Query Processing Time: " + overallQueryProcessingTime + " - Query Planning Time: " + planningTime +" - Compile Time: " + compilationTime + " - Execution time: " + executionTime;
```
hartig, line 172: Can you please change this such that it really produces a line for a CSV file.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
```diff
@@ -158,9 +160,18 @@ protected void exec() {
...
+			if ( contains(argQueryProcStats) ) {
+				StatsPrinter.print(statsAndExceptions.object1, System.err, true);
+			}
+			if ( contains(argOnelineTimeStats) ) {
+				final long overallQueryProcessingTime = statsAndExceptions.object1.getOverallQueryProcessingTime();
+				final long planningTime = statsAndExceptions.object1.getPlanningTime();
+				final long compilationTime = statsAndExceptions.object1.getCompilationTime();
+				final long executionTime = statsAndExceptions.object1.getExecutionTime();
+				final String queryProcStats = overallQueryProcessingTime + ", " + planningTime +", " + compilationTime + ", " + executionTime + "\n";
+				System.out.println(queryProcStats);
```
hartig, line 173: Just one last thing: You don't need the line break ("\n") at the end of the string because you are using `println(...)` in the next line. Hence, with the current code you may actually get an empty line in between each CSV line.

Here is the fix for this issue, which I will apply myself right away.
```suggestion
				final String queryProcStats = overallQueryProcessingTime + ", " + planningTime +", " + compilationTime + ", " + executionTime + "\n";
				System.out.println(queryProcStats);
```

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
```diff
@@ -42,6 +42,7 @@ public class RunQueryWithoutSrcSel extends CmdARQ
 	protected final ArgDecl argPrintLogicalPlan   = new ArgDecl(ArgDecl.NoValue, "printLogicalPlan");
 	protected final ArgDecl argPrintPhysicalPlan  = new ArgDecl(ArgDecl.NoValue, "printPhysicalPlan");
 	protected final ArgDecl argQueryProcStats = new ArgDecl(ArgDecl.NoValue, "printQueryProcStats");
+	protected final ArgDecl argOnelineTimeStats = new ArgDecl(ArgDecl.NoValue, "printQueryProcTime");
```
hartig, line 45:
```suggestion
	protected final ArgDecl argOnelineTimeStats = new ArgDecl(ArgDecl.NoValue, "printQueryProcMeasurements");
```

hartig (changes requested):
Two more changes that I also will commit right away: the idea of these changes is to change the name of the command line argument from `--printQueryProcTime` to `--printQueryProcMeasurements` because, in the future, we may add other, non-time-related measurements to this CSV line (e.g., result size).

PR #277, opened 2023-01-31
Heuristic-based query reordering

chengsijin0817:
Apply a heuristic for query reordering based on 'JoinAwareWeightedUnboundVariableCount'

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/GreedyBasedReordering.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/GreedyBasedReordering.java
@@ -59,34 +59,25 @@ public class GreedyBasedReordering implements HeuristicForLogicalOptimization {
     }
 
     protected LogicalPlan reorderSubPlans( final LogicalPlan inputPlan ) {
+        // Initialize candidatePlans
         final List<QueryAnalyzer> candidatePlans = new ArrayList<>();
-        final List<QueryAnalyzer> selectedSubPlans = new ArrayList<>();
-
-        // Find the first subQuery and put it into selectedSubPlans
-        double costOfBestSubPlan = Double.MAX_VALUE;
-        QueryAnalyzer bestSubPlan = null;
         for ( int i = 0; i < inputPlan.numberOfSubPlans(); i ++ ) {
             final QueryAnalyzer subPlan = new QueryAnalyzer( inputPlan.getSubPlan(i) );
             candidatePlans.add( subPlan );
-            final double selectivity = estimateSelectivity( selectedSubPlans, subPlan);
-
-            if ( selectivity < costOfBestSubPlan ) {
-                costOfBestSubPlan = selectivity;
-                bestSubPlan = subPlan;
-            }
         }
-        selectedSubPlans.add(bestSubPlan);
-        candidatePlans.remove(bestSubPlan);
 
-        // Find the next subQuery from the remaining subPlans
+        // Find the next subPlan with the lowest cost and put it into selectedSubPlans
+        final List<QueryAnalyzer> selectedSubPlans = new ArrayList<>();
         while ( !candidatePlans.isEmpty() ) {
-            findNextPlan(selectedSubPlans, candidatePlans);
+            final QueryAnalyzer bestSubPlan = findNextPlan(selectedSubPlans, candidatePlans);
+            selectedSubPlans.add(bestSubPlan);
+            candidatePlans.remove(bestSubPlan);
         }
 
         return constructBinaryPlan(selectedSubPlans);
     }
 
-    protected void findNextPlan(final List<QueryAnalyzer> selectedSubPlans, final List<QueryAnalyzer> candidatePlans ) {
+    protected QueryAnalyzer findNextPlan( final List<QueryAnalyzer> selectedSubPlans, final List<QueryAnalyzer> candidatePlans ) {
         double costOfBestSubPlan = Double.MAX_VALUE;
         QueryAnalyzer bestSubPlan = null;
         for ( final QueryAnalyzer subPlan : candidatePlans) {
@@ -96,9 +87,7 @@ public class GreedyBasedReordering implements HeuristicForLogicalOptimization {
                 bestSubPlan = subPlan;
             }
         }
-
-        selectedSubPlans.add(bestSubPlan);
-        candidatePlans.remove(bestSubPlan);
+        return bestSubPlan;
     }
 
     protected LogicalPlan constructBinaryPlan( final List<QueryAnalyzer> selectedSubPlans ) {


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/SubQuery_Analyzer.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class SubQuery_Analyzer {
+    private List<Node> subs, preds, objs;
+
+    public SubQuery_Analyzer( final LogicalPlan lop ) {
+        subs = new ArrayList<>();
+        preds = new ArrayList<>();
+        objs = new ArrayList<>();
+
+        analyze(lop);
```
hartig, line 21: I wouldn't hide the functionality to populate the three lists inside that function because that makes the code less easy to understand. Instead, use the function only to extract/obtain the set of triple patterns and, then, populate the lists directly here in the constructor.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Var_Analyzer.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.apache.jena.graph.Node;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class Var_Analyzer {
+
+//    varsTotal: a set of all variables that have been calculated
```
hartig, line 11: How can "variables [be] calculated"?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Var_Analyzer.java
```diff
@@ -0,0 +1,35 @@
...
+//    Add all bound variables to bindings
```
hartig, line 12: If this set is meant to contain bound variables, then why call it `bindings`?? Call it `boundVariables` instead!

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Var_Analyzer.java
```diff
@@ -0,0 +1,35 @@
...
+    public static int calculateVars( final List<Node> vars, final Set<Node> varsTotal, final Set<Node> bindings ) {
+        varsTotal.addAll( getUniqueNodes(vars) );
```
hartig, line 14: Given that `varsTotal` is a `Set`, there is no need to call `getUniqueNodes` and create yet another `HashSet` there.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Var_Analyzer.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class Var_Analyzer {
+
+//    varsTotal: a set of all variables that have been calculated
+//    Add all bound variables to bindings
+    public static int calculateVars( final List<Node> vars, final Set<Node> varsTotal, final Set<Node> bindings ) {
```
hartig, line 13: It may be very difficult to understand the code that is using this function because the function has the side effect of modifying both `varsTotal` and `bindings`. It would be better to move this code into the place where you call this function.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Join_Analyzer.java
```diff
@@ -0,0 +1,53 @@
...
+        final int chainjoins = countDuplicates(vars_s,vars_o);
+        return chainjoins * J_Tc;
+    }
+
+    public static double getUnusualJoins( final List<Node> vars_s, final List<Node> vars_p, final List<Node> vars_o, final double J_Tu ) {
+        final int unusualjoins = countDuplicates(vars_s,vars_p) + countDuplicates(vars_p,vars_o)+ countDuplicates(vars_p);
+        return unusualjoins * J_Tu;
+    }
+
+    private static int countDuplicates( final List<Node> vars ) {
```
hartig, line 28: This function is difficult to understand for me. Can you please add a JavaDoc comment that specifies what exactly the function is supposed to do.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Join_Analyzer.java
```diff
@@ -0,0 +1,53 @@
...
+        vars.forEach(s-> histogram.put(s, (histogram.containsKey(s) ? histogram.get(s) : -1) +1)
+        );
+        final AtomicInteger dups= new AtomicInteger(0);
+        histogram.keySet().forEach(s -> {
+            if(histogram.get(s) > 0)  dups.set(dups.get() + histogram.get(s));
+        });
+        return dups.get();
+    }
+
+    private static int countDuplicates( final List<Node> vars_a, final List<Node> vars_b ) {
```
hartig, line 39: Same comment here. Please add a JavaDoc comment that specifies what this function does.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Join_Analyzer.java
```diff
@@ -0,0 +1,53 @@
...
+        final int jobjs = countDuplicates(vars_o);
+        return ( jsubs + jobjs ) * J_Ts;
+    }
+
+    public static double getChainJoins( final List<Node> vars_s, final List<Node> vars_o, final double J_Tc ) {
+        final int chainjoins = countDuplicates(vars_s,vars_o);
+        return chainjoins * J_Tc;
+    }
+
+    public static double getUnusualJoins( final List<Node> vars_s, final List<Node> vars_p, final List<Node> vars_o, final double J_Tu ) {
```
hartig, line 23: These three function names are misleading because the functions do not actually return any joins.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
```diff
@@ -0,0 +1,68 @@
...
+    public static final double J_Ts=0.5; // weight for star join
+    public static final double J_Tu=1; // weight for unusual join
+    public static final double J_Tn=0; // no join
+
+    public static double estimate( final List<LogicalPlan> selectedPlans, LogicalPlan lop ) {
+        Set<Node> bindings = new HashSet<>();
+        for ( LogicalPlan plan : selectedPlans ) {
+            final SubQuery_Analyzer analyzer = new SubQuery_Analyzer( plan );
+            bindings = Var_Analyzer.addBinds( analyzer.getSubs(), analyzer.getPreds(), analyzer.getObjs() );
+        }
```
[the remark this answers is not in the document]
chengsijin0817: For points 3 and 4, I have added 'plan' as a member of 'Query_Analyzer', and used 'Query_Analyzer' as elements of 'candidatePlans' and 'selectedSubPlans' to avoid repeatedly creating Query_Analyzer for the same sub-plan. 
(see commit [22b0b7b](https://github.com/LiUSemWeb/HeFQUIN/pull/277/commits/22b0b7bb2bdbced295bb2bbbe3fb2d04de01bf66))

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Join_Analyzer.java
```diff
@@ -0,0 +1,75 @@
...
+public class Join_Analyzer {
+
+    // Consider the subject-subject and object-object joins as star joins
+    public static double getNumOfStarJoins( final List<Node> vars_s, final List<Node> vars_o ){
+        final int jsubs = countDuplicates(vars_s);
+        final int jobjs = countDuplicates(vars_o);
+        return ( jsubs + jobjs );
+    }
+
+    // Consider the object-subject and subject-object joins as chain joins
```
hartig, line 18: What does "consider" mean? What is the purpose of this function/what does it do?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Join_Analyzer.java
```diff
@@ -0,0 +1,75 @@
...
+    public static double getNumOfChainJoins( final List<Node> vars_s, final List<Node> vars_o ) {
+        return countDuplicates(vars_s,vars_o);
+    }
+
+    // Consider subject-predicate, predicate-object, predicate-predicate joins as unusual joins
```
hartig, line 23: What does "consider" mean? What is the purpose of this function/what does it do?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
```diff
@@ -0,0 +1,113 @@
...
+    protected static final double J_Ts = 0.5; // weight for star join
+    protected static final double J_Tu = 1; // weight for unusual join
+
+    public static double estimate( final List<Query_Analyzer> selectedPlans, final Query_Analyzer subPlan ) {
+        // Add bound variables of all selected plans to a set
+        Set<Node> boundVariables = new HashSet<>();
+
+        // This part can be optimized if all formulas need to consider bound variables:
+        // adding to this set directly when adding a new subPlan to 'selectedPlans'
+        for ( Query_Analyzer plan : selectedPlans ) {
```
hartig, line 38:
```suggestion
        for ( final Query_Analyzer plan : selectedPlans ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
```diff
@@ -0,0 +1,113 @@
...
+            boundVariables = addBinds( plan.getSubs(), plan.getPreds(), plan.getObjs() );
```
hartig, line 39: This line is overriding `boundVariables` for every `plan`. I don't think that this is what you want to do. I suggest remove the `addBinds` function and move the code from that function directly into this for loop here.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        return calculateCost( boundVariables, subPlan );
+    }
+
+    // Formula (7) in paper "Heuristics-based Query Reordering for Federated Queries in SPARQL 1.1 and SPARQL-LD"
+    private static double calculateCost( final Set<Node> bindings, final Query_Analyzer subPlan ) {
```
hartig, line 46: The variable name `bindings` is misleading.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
```diff
@@ -0,0 +1,113 @@
...
+        final int totalSubs = calculateVars(vars_s, varsTotal, boundVariables);
+        final int totalObjs = calculateVars(vars_o, varsTotal, boundVariables);
+        final int totalPreds = calculateVars(vars_p, varsTotal, boundVariables);
+
+        return calculateTripleWeights(totalSubs, totalPreds, totalObjs);
+    }
+
+    /**
+     * Creates a collection with all variables that have been currently calculated
+     * and adds the bound variables to bindings
```
hartig, line 72: It is still not clear what is meant by "variables that have been [...] calculated."

Additionally, there is nothing called `bindings` here.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param vars A list of variables, can be subs, preds or objects
+     * @param varsTotal A list of all variables that have been currently calculated in this subquery
+     * @param boundVariables All bound variables (including variables in selected plans and calculated part of this subquery)
```
hartig, line 75: Also here, I don't understand what you mean by "calculated". Do you want to say "counted"?

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
chengsijin0817: Hi @hartig , for the implementation of cost calculation (using formulas), since some details are missing in the paper so I use some methods from the corresponding [GitHub repository](https://github.com/TYannakis/SPARQL-LD-Query-Optimizer/blob/14902192461e41df34335e67f5c9998935b7ad9d/src/main/java/ics/forth/formulas/GreedyJWUVC.java#L200) of that paper. 

In particular, I directly used the implementation for this function of that repository. But after carefully reading it, I realized that the parameter "varsTotal" is not needed so I have removed it.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/utils/Join_Analyzer.java
```diff
@@ -0,0 +1,100 @@
...
+        for ( final Node a: vars_a ) {
+            final int count_a = map.containsKey(a)? (map.get(a) + 1) : 1;
+            map.put( a, count_a );
+        }
+
+        int sum = 0;
+        for ( final Node b: vars_b ) {
+            if ( map.containsKey(b) && map.get(b) > 0 ) {
+                sum += 1;
+                map.put( b, map.get(b)-1 );
```
hartig, line 93: Instead of doing three accesses to the map based on the same key, you can reduce this to only one access:
```suggestion
            final Integer bCounter = map.get(b);
            if ( bCounter != null && bCounter > 0 ) {
                sum++;
                map.put( b, bCounter-1 );
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/formula/JoinAwareWeightedUnboundVariableCount.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param boundVariables All bound variables (including variables in selected plans and counted part of this subquery)
+     * @return The number of unbounded variables
+     */
+    private static int countUnboundVars( final List<Node> vars, final Set<Node> boundVariables ) {
+        final Set<Node> uniqueVars = new HashSet<>(vars);
+        uniqueVars.removeAll(boundVariables);
+
+        boundVariables.addAll( uniqueVars );
+        return uniqueVars.size();
+    }
```
hartig, line 77: I am still feeling a bit uncomfortable with the fact that this function modifies the given `boundVariables` set as an (undocumented) side effect of producing the return value. In fact, I believe that this can even have a (perhaps undesired or desired??) effect on the result of the `weightedUnboundVarsCount` function that uses this function here. I mean, depending on whether you call this function first for the subjects and then for the objects or the other way around, you may get different counts. Isn't this a problem?

Also, the documentation comment for this function is still not sufficiently clear. I mean, read the sentence that you wrote ("For a given list of variables, count the number of unbound variables") and then think that I give you a list of variables and ask you to tell me how many of them are unbound. You would probably ask: what do you mean by "unbound"? You may also ask: what should I do with a variable that occurs multiple times in the given list and that is "unbound"--should I count that variable only ones or multiple times?

hartig (changes requested):
I few more comments that I would like you to address. Additionally, I would feel more comfortable if there were unit tests for the functions in `Join_Analyzer` and for `weightedUnboundVarsCount` in `JoinAwareWeightedUnboundVariableCount`.

PR #281, opened 2023-03-07
Parallelism based cost model

chengsijin0817:


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/cardinality/VarSpecificCardinalityEstimationImpl.java
```diff
@@ -165,6 +165,35 @@ protected CompletableFuture<Integer> _initiateUnionCardinalityEstimation(
 
 	}
 
+	protected CompletableFuture<Integer> _initiateMultiwayUnionCardinalityEstimation(
+			final List<PhysicalPlan> subPlans,
+			final Var v )
+	{
+		final Set<Var> allJoinVars = ExpectedVariablesUtils.intersectionOfAllVariables( subPlans.toArray(new PhysicalPlan[0]) );
+
+		if ( allJoinVars.contains(v) ) {
```
hartig, line 174: I am not sure but I believe that this is wrong. Variable `v` is a join variable not only if it appears in _all_ of the subplans under the union but also if it appears in only one of these subplans.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/cardinality/VarSpecificCardinalityEstimationImpl.java
```diff
@@ -95,6 +88,13 @@ else if ( rootOp instanceof LogicalOpJoin ) {
 		else if ( rootOp instanceof LogicalOpUnion ) {
 			return _initiateUnionCardinalityEstimation( plan.getSubPlan(0), plan.getSubPlan(1), v );
 		}
+		else if ( rootOp instanceof LogicalOpMultiwayUnion ) {
+			final List<PhysicalPlan> subPlans = new ArrayList<>();
+			for( int i = 0; i < plan.numberOfSubPlans(); i++ ) {
+				subPlans.add( plan.getSubPlan(i) );
+			}
```
hartig, line 95: I suggest to avoid creating and populating this extra list by, instead, simply passing the `plan` directly to the helper function `_initiateMultiwayUnionCardinalityEstimation`. The for-loops in that function can then iterate over the subplans in the way you do it here (i.e., based on `int i` and calling `plan.getSubPlan(i)`).

hartig (changes requested):

hartig (approved):
Good!

PR #283, opened 2023-03-09
Cardinality based greedy join plan optimizer

chengsijin0817:
Implements a query optimizer that builds left-deep query plans, for which it uses a greedy approach to determine the join order based on cardinality estimation, and then choose the physical algorithm according to the estimated number of requests to execute the join.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
@@ -140,22 +140,26 @@ public class CardinalityBasedGreedyJoinPlanOptimizerImpl extends JoinPlanOptimiz
 
                     final FederationMember fm = flattenRequestMemPairs.get(index).getMember();
                     planWithStatistics.addNumOfAccess( accessNumForReq( resps[index].getCardinality(), fm ) );
+                    planWithStatistics.addFederationMembers( new ArrayList<>( Arrays.asList(fm) ) );
 
                     index++;
                 }
                 else if (pop instanceof PhysicalOpBinaryUnion || pop instanceof PhysicalOpMultiwayUnion) {
                     int sumCard = 0, sumAccess = 0;
+                    final List<FederationMember> fms = new ArrayList<>();
                     for ( int count = 0; count < plan.numberOfSubPlans(); count++ ) {
                         sumCard += resps[index].getCardinality();
 
                         final FederationMember fm = flattenRequestMemPairs.get(index).getMember();
                         sumAccess += accessNumForReq( resps[index].getCardinality(), fm );
+                        fms.add( fm );
 
                         index++;
                     }
 
                     planWithStatistics.addCandidate( sumCard );
                     planWithStatistics.addNumOfAccess( sumAccess );
+                    planWithStatistics.addFederationMembers( fms );
                 }
                 else
                     throw new IllegalArgumentException("Unsupported type of subquery in source assignment (" + pop.getClass().getName() + ")");
@@ -192,13 +196,16 @@ public class CardinalityBasedGreedyJoinPlanOptimizerImpl extends JoinPlanOptimiz
          * For the comparison, only keeping the second element is enough.
          */
         protected PhysicalPlanWithStatistics decidePhysicalAlgorithm( final PhysicalPlanWithStatistics currentPlan, final PhysicalPlanWithStatistics nextPlan ) {
-            final double blockSize = 30;
-            final double accNumSHJ = // currentPlan.getNumOfAccess() +
-                                        nextPlan.getNumOfAccess();
-            final double accNumBJ = // currentPlan.getNumOfAccess() +
-                                       currentPlan.getCandidate() / blockSize;
+            final double accNumSHJ = nextPlan.getNumOfAccess();
 
-            PhysicalPlan newPlan = null;
+            final List<FederationMember> fms = nextPlan.getFederationMembers();
+            double accNumBJ = 0;
+            for ( final FederationMember fm: fms ) {
+                double blockSize = setBlockSize(fm);
+                accNumBJ += currentPlan.getCandidate() / blockSize;
+            }
+
+            PhysicalPlan newPlan;
             if ( accNumSHJ <= accNumBJ ) {
                 newPlan = PhysicalPlanFactory.createPlanWithJoin(currentPlan.plan, nextPlan.plan);
             }
@@ -228,6 +235,24 @@ public class CardinalityBasedGreedyJoinPlanOptimizerImpl extends JoinPlanOptimiz
             else
                 throw new IllegalArgumentException("Unsupported type of federation member: " + fm.getClass().getName() );
         }
+
+        /**
+         * The block size (number of bindings can be attached) depends on the type of interface
+         */
+        protected double setBlockSize( final FederationMember fm ) {
+            if ( fm instanceof SPARQLEndpoint ){
+                return 50;
+            }
+            else if ( fm instanceof BRTPFServer ){
+                return 30;
+            }
+            else if ( fm instanceof TPFServer ){
+                return 1;
+            }
+            else
+                throw new IllegalArgumentException("Unsupported type of federation member: " + fm.getClass().getName() );
+        }
+
     }
         
 }
@@ -236,6 +261,7 @@ class PhysicalPlanWithStatistics {
     public final PhysicalPlan plan;
     protected int candidate;
     protected int numOfAccess;
+    protected List<FederationMember> fms;
 
     public PhysicalPlanWithStatistics( final PhysicalPlan plan ){
         this.plan = plan;
@@ -243,6 +269,10 @@ class PhysicalPlanWithStatistics {
 
     public void addCandidate( final int candidate ){
         this.candidate = candidate;
+    }
+
+    public void addFederationMembers( final List<FederationMember> fms ){
+        this.fms = fms;
     }
 
     public void addNumOfAccess( final int numOfAccess ){
@@ -255,6 +285,10 @@ class PhysicalPlanWithStatistics {
 
     public int getNumOfAccess( ){
         return numOfAccess;
+    }
+
+    public List<FederationMember> getFederationMembers( ){
+        return fms;
     }
 
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
@@ -204,7 +204,7 @@ public class CardinalityBasedGreedyJoinPlanOptimizerImpl extends JoinPlanOptimiz
                 accNumBJ += currentPlan.getCandidate() / blockSize;
             }
 
-            PhysicalPlan newPlan;
+            final PhysicalPlan newPlan;
             if ( accNumSHJ <= accNumBJ ) {
                 newPlan = PhysicalPlanFactory.createPlanWithJoin(currentPlan.plan, nextPlan.plan);
             }
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
@@ -264,23 +264,15 @@ public class CardinalityBasedGreedyJoinPlanOptimizerImpl extends JoinPlanOptimiz
 
     class PhysicalPlanWithStatistics {
         public final PhysicalPlan plan;
-        protected int cardinality;
-        protected int numOfAccess;
-        protected List<FederationMember> fms;
-
-        public PhysicalPlanWithStatistics( final PhysicalPlan plan ){
-            this.plan = plan;
-        }
+        protected final int cardinality;
+        protected final int numOfAccess;
+        protected final List<FederationMember> fms;
 
         public PhysicalPlanWithStatistics( final PhysicalPlan plan, final List<FederationMember> fms, final int cardinality, final int numOfAccess ){
             this.plan = plan;
             this.fms = fms;
             this.cardinality = cardinality;
             this.numOfAccess = numOfAccess;
-        }
-
-        public void setCardinality( final int cardinality ){
-            this.cardinality = cardinality;
         }
 
         public int getCardinality( ){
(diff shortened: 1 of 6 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	protected class GreedyConstructionAlgorithm implements EnumerationAlgorithm {
+        protected final List<PhysicalPlan> subplans;
+
+        public GreedyConstructionAlgorithm( final List<PhysicalPlan> subplans ) {
+            this.subplans = subplans;
+        }
+
+        public PhysicalPlan getResultingPlan() throws PhysicalOptimizationException {
```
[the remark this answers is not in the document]
hartig: Additionally, it would be very helpful for the future if you add a few comments into the code of this function to give our future selves or other readers of this code an idea of what is going on here. I mean, you don't need to comment every single line of code but each of the relevant steps within this functions should have a comment. As an example, you can take a look at the `apply` function of the `UnionPullUp` class.

For instance, for the first four lines below, you can add a comment such as: "Create a list of RequestMemberPair objects for all request operators inside the given list of subplans, where the list of these objects is ordered in the order in which the request operators can be found by a depth-first traversal of the subplans."

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+                resps = FederationAccessUtils.performCardinalityRequests(fedAccessMgr, flattenRequestMemPairs.toArray(new RequestMemberPair[0]));
+            } catch (final FederationAccessException e) {
+                throw new RuntimeException("Issuing a cardinality request caused an exception.", e);
+            }
+
+            final List<PhysicalPlanWithStatistics> subPlansWithStatistics = associateCardWithSubPlans( flattenRequestMemPairs, resps);
+
+            PhysicalPlanWithStatistics currentPlan = null;
+            for ( int k = 0; k < subplans.size(); k++ ) {
+                final PhysicalPlanWithStatistics nextPlan = chooseNextSubPlanBasedOnCard(subPlansWithStatistics);
```
hartig, line 62: From this code, it is not immediately obvious that all that happens here is that the subplans are simply picked in order of their respective cardinality estimates. I would like that to be more obvious from the code itself.

You can simply sort the `subPlansWithStatistics` using the `sort` method of `Collections` with a `Comparator<PhysicalPlanWithStatistics>` that compares the `PhysicalPlanWithStatistics` objects based on their cardinality.

After that, you only need to iterate over that sorted list, which you can do with an `Iterator<PhysicalPlanWithStatistics>` where the first call of `next()` is used to initialize `currentPlan` and then you do `decidePhysicalAlgorithm` inside a while loop.

That would significantly simplify the code here and, also, make it easier to understand what is going on here. Additionally, that would avoid having to repeatedly go over all remaining elements of `subPlansWithStatistics` over and over again inside `chooseNextSubPlanBasedOnCard`, and it also avoids having to remove every `nextPlan` from `subPlansWithStatistics`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+         * - BJ: Card(T1)/PageSize + Card(T1)/blockSize
+         * For the comparison, only keeping the second element is enough.
+         */
+        protected PhysicalPlanWithStatistics decidePhysicalAlgorithm( final PhysicalPlanWithStatistics currentPlan, final PhysicalPlanWithStatistics nextPlan ) {
+            final double accNumSHJ = nextPlan.getNumOfAccess();
+
+            final List<FederationMember> fms = nextPlan.getFederationMembers();
+            double accNumBJ = 0;
+            for ( final FederationMember fm: fms ) {
+                double blockSize = setBlockSize(fm);
```
hartig, line 204: The function name is misleading. You are not actually setting anything here. Better name the function `determineBlockSize`.

Additionally, `blockSize` can be `final`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+         * - SHJ: Card(T1)/PageSize + Card(T2)/PageSize
+         * - BJ: Card(T1)/PageSize + Card(T1)/blockSize
+         * For the comparison, only keeping the second element is enough.
+         */
+        protected PhysicalPlanWithStatistics decidePhysicalAlgorithm( final PhysicalPlanWithStatistics currentPlan, final PhysicalPlanWithStatistics nextPlan ) {
+            final double accNumSHJ = nextPlan.getNumOfAccess();
+
+            final List<FederationMember> fms = nextPlan.getFederationMembers();
+            double accNumBJ = 0;
+            for ( final FederationMember fm: fms ) {
```
hartig, line 203:
```suggestion
            double accNumBJ = 0;
            for ( final FederationMember fm: nextPlan.getFederationMembers() ) {
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+                double blockSize = setBlockSize(fm);
+                accNumBJ += currentPlan.getCandidate() / blockSize;
```
hartig, line 205: "candidate"?? You really need to rename that function!

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+            }
+
+            PhysicalPlan newPlan;
```
hartig, line 208:
```suggestion
            final PhysicalPlan newPlan;
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+                }
+            }
+
+            return bestPlanWithCandidate;
+        }
+
+        /**
+         * The physical algorithm is determined based on the estimated number of requests to execute the join
+         * - SHJ: Card(T1)/PageSize + Card(T2)/PageSize
+         * - BJ: Card(T1)/PageSize + Card(T1)/blockSize
```
hartig, line 195: In addition to mentioning the formulas here, can you please also reference the corresponding formula in the paper.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+
+            PhysicalPlan newPlan;
+            if ( accNumSHJ <= accNumBJ ) {
+                newPlan = PhysicalPlanFactory.createPlanWithJoin(currentPlan.plan, nextPlan.plan);
+            }
+            else {
+                newPlan = PhysicalPlanFactory.createPlanWithDefaultUnaryOpIfPossible(currentPlan.plan, nextPlan.plan);
+            }
+
+            final PhysicalPlanWithStatistics newPlanWithStatistics = new PhysicalPlanWithStatistics( newPlan );
```
hartig, line 216: When you change this part based on my earlier comment (about passing all statistic values directly to the constructor of `PhysicalPlanWithStatistics`), you do not need to create a new list of fed.members but instead pass `null` as this list to the constructor, but add a comment saying that the list of fed.members will not be accessed anymore for this new `PhysicalPlanWithStatistics` object that you are creating here.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+
+            final int joinCardinality = Math.min( currentPlan.getCandidate(), nextPlan.getCandidate() );
```
hartig, line 218: Can you please add a comment here that refers to the corresponding part/formula of the paper, where they say that they use min for this.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,291 @@
...
+            else
+                throw new IllegalArgumentException("Unsupported type of federation member: " + fm.getClass().getName() );
+        }
+
+        /**
+         * The block size (number of bindings can be attached) depends on the type of interface
+         */
+        protected double setBlockSize( final FederationMember fm ) {
+            if ( (fm instanceof SPARQLEndpoint) || (fm instanceof BRTPFServer) ){
+                return 30;
```
hartig, line 244: Also here, instead of hard-coding the constant directly in the code, use the corresponding `static` member variable that we already have somewhere else.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,302 @@
...
+        public PhysicalPlanWithStatistics( final PhysicalPlan plan, final List<FederationMember> fms, final int cardinality, final int numOfAccess ){
+            this.plan = plan;
+            this.fms = fms;
+            this.cardinality = cardinality;
+            this.numOfAccess = numOfAccess;
+        }
+
+        public void setCardinality( final int cardinality ){
+            this.cardinality = cardinality;
+        }
```
hartig, line 287:
```suggestion
        protected final int cardinality;
        protected final int numOfAccess;
        protected final List<FederationMember> fms;

        public PhysicalPlanWithStatistics( final PhysicalPlan plan, final List<FederationMember> fms, final int cardinality, final int numOfAccess ){
            this.plan = plan;
            this.fms = fms;
            this.cardinality = cardinality;
            this.numOfAccess = numOfAccess;
        }
```

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,350 @@
...
+            final List<PhysicalPlanWithStatistics> subPlansContainsVars = new ArrayList<>();
+            for ( final PhysicalPlanWithStatistics subplan: subPlansWithStatistics ) {
+                if ( ! ExpectedVariablesUtils.intersectionOfAllVariables( vars, subplan.plan.getExpectedVariables() ).isEmpty() ){
+                    subPlansContainsVars.add(subplan);
+                }
+            }
+
+            if ( !subPlansWithStatistics.isEmpty() && subPlansContainsVars.isEmpty() ){
+                // Independent subplans exist. In this case, choose the subplan with the lowest candidate (from remaining subPlans) as candidate plan
+                return Arrays.asList( chooseFirstSubPlan(subPlansWithStatistics) );
```
hartig, line 255: No, this doesn't belong into this function. Move it into the loop from where you call this function.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,350 @@
...
+            // To achieve this, we implement a custom comparator and create a PriorityQueue to store candidate subPlans in cardinality order.
+            final Comparator<PhysicalPlanWithStatistics> orderBasedOnCard = new Comparator<PhysicalPlanWithStatistics>() {
+                @Override
+                public int compare( PhysicalPlanWithStatistics o1, PhysicalPlanWithStatistics o2 ) {
+                    return Integer.compare( o1.getCardinality(), o2.getCardinality() );
+                }
+            };
+            final PriorityQueue<PhysicalPlanWithStatistics> orderedCandidateSubPlans = new PriorityQueue<>(orderBasedOnCard);
+
+            for ( int i = 1; i < subplans.size(); i ++ ){
```
hartig, line 93: This has confused me a lot (Why is the iteration starting from index 1 here? is it because a first subplan was already picked? But that subplan would not necessarily be the first subplan (with index 0)!).

So, I had to think twice about this. I understand now that this is correct, but a much more obvious and less confusing thing would be to have a while loop here, with the while condition that at least one of the following lists is not empty yet: `orderedCandidateSubPlans` and `orderedCandidateSubPlans`.
```suggestion
            while ( ! orderedCandidateSubPlans.isEmpty() && ! orderedCandidateSubPlans.isEmpty() ) {
```

From the code readability perspective, I would find it more natural to

hartig (changes requested):
Almost there now ;-)

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedGreedyJoinPlanOptimizerImpl.java
chengsijin0817: It does make more sense! Just for confirmation, do you mean to refer to these lists: 'orderedCandidateSubPlans' and 'subPlansWithStatistics'?

hartig: Yes. These are the two lists that I meant. Sorry for the typo! (must have been a copy & paste mistake)

hartig (approved):
Looks good now!

PR #285, opened 2023-03-13
Avoid counting duplicates in cost model

chengsijin0817:
When aggregating cost values in the cost model, avoid counting duplicates by skipping the subquery that has already been visited.


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/costmodel/CostFunctionForPlan.java
```diff
@@ -10,5 +11,5 @@
  */
 public interface CostFunctionForPlan
 {
-	CompletableFuture<Integer> initiateCostEstimation( PhysicalPlan plan );
+	CompletableFuture<Integer> initiateCostEstimation( final Set<PhysicalPlan> visitedPlans, final PhysicalPlan plan );
```
hartig, line 14: Please add a JavaDoc comment to describe what this function is supposed to do and, in particular, what the  given `visitedPlans` have to do with it (i.e., what the purpose of `visitedPlans` is).

hartig (changes requested):

PR #290, opened 2023-04-25
Fix an issue in QueryPatternUtils

chengsijin0817:
Fix an issue: Class org.apache.jena.sparql.syntax.ElementGroup cannot be cast to class org.apache.jena.sparql.syntax.ElementPathBlock (see the detailed description below):
https://github.com/MaastrichtU-IDS/federatedQueryKG/issues/6


hartig:
This fix does not only fix https://github.com/MaastrichtU-IDS/federatedQueryKG/issues/6 but also https://github.com/MaastrichtU-IDS/federatedQueryKG/issues/7 and https://github.com/MaastrichtU-IDS/federatedQueryKG/issues/8

PR #293, opened 2023-05-02
Extend the implementation of 'ApplyVocabularyMappings'

chengsijin0817:
Extend the implementation of the class 'ApplyVocabularyMappings' to generate query plans that fully consider all possible vocabulary mapping, or generate a compressed version of the query plan by configuring 'compressVocabularyRewriting'.

In order to generate a compressed version of a query plan, check if adding an l2g operator over a request is necessary by examining the form of the graph pattern. 
(Note the current implementation assumes that only concepts and roles are considered in vocabulary mapping.)

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
@@ -3,7 +3,9 @@ package se.liu.ida.hefquin.engine.queryproc.impl.loptimizer;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 
+import se.liu.ida.hefquin.engine.data.VocabularyMapping;
 import se.liu.ida.hefquin.engine.data.mappings.VocabularyMappingUtils;
 import se.liu.ida.hefquin.engine.federation.FederationMember;
 import se.liu.ida.hefquin.engine.federation.SPARQLEndpoint;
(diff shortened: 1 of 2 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -29,36 +31,14 @@
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithNaryRootImpl;
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithNullaryRootImpl;
 import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithUnaryRootImpl;
+import se.liu.ida.hefquin.engine.queryplan.utils.LogicalOpUtils;
 
 public class ApplyVocabularyMappings implements HeuristicForLogicalOptimization {
+	boolean compressVocabularyRewriting;
```
hartig, line 37:
```suggestion
	protected final boolean compressVocabularyRewriting;

```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -67,14 +47,44 @@ public static LogicalPlan rewriteToUseLocalVocabulary( final LogicalPlan inputPl
...
-			} else {
+
+			// For a compressed version of query plan, check if it is necessary to add the l2g operator over a request based on the form of the graph pattern.
+			// The current implementation assumes that only concepts and roles are being considered in vocabulary mapping.
+			boolean rewriteNeeded = false;
+			if ( compressVocabularyRewriting ) {
+				final Set<TriplePattern> tps = LogicalOpUtils.getTriplePatternsOfReq(requestOp);
+				if ( !tps.isEmpty() ) {
+					for ( final TriplePattern tp : tps ) {
+						// If any triple pattern is in the form of (-, ?p, -) or (-, rdfs:type, ?o),
```
hartig, line 58:
```suggestion
						// If any triple pattern is in the form of (-, ?p, -) or (-, rdf:type, ?o),

```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -67,14 +47,44 @@ public static LogicalPlan rewriteToUseLocalVocabulary( final LogicalPlan inputPl
...
 		if (inputPlan.getRootOperator() instanceof LogicalOpRequest) {
 			final LogicalOpRequest<?,?> requestOp = (LogicalOpRequest<?,?>) inputPlan.getRootOperator();
-			if(requestOp.getFederationMember().getVocabularyMapping() != null) { // If fm has a vocabulary mapping vm
-				final LogicalOpLocalToGlobal l2g = new LogicalOpLocalToGlobal(requestOp.getFederationMember().getVocabularyMapping());
-				final LogicalPlan rw = rewriteToUseLocalVocabulary(inputPlan);
-				return new LogicalPlanWithUnaryRootImpl(l2g,rw);
-			} else {
+
+			// For a compressed version of query plan, check if it is necessary to add the l2g operator over a request based on the form of the graph pattern.
+			// The current implementation assumes that only concepts and roles are being considered in vocabulary mapping.
```
hartig, line 52:
```suggestion
			// The current implementation assumes that only concepts and roles are being considered in vocabulary mapping
			// and that the data of the federation members is only instance data (i.e., properties
			// appear only in the predicate position of triples and classes appear only in the
			// object position of rdf:type triples).
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -67,14 +47,44 @@ public static LogicalPlan rewriteToUseLocalVocabulary( final LogicalPlan inputPl
...
+			boolean rewriteNeeded = false;
+			if ( compressVocabularyRewriting ) {
+				final Set<TriplePattern> tps = LogicalOpUtils.getTriplePatternsOfReq(requestOp);
+				if ( !tps.isEmpty() ) {
+					for ( final TriplePattern tp : tps ) {
+						// If any triple pattern is in the form of (-, ?p, -) or (-, rdfs:type, ?o),
+						// the intermediate results might need to be rewritten. This requires adding a L2G operator over the request.
+						if ( tp.asJenaTriple().getPredicate().isVariable()
+								|| (tp.asJenaTriple().getPredicate().hasURI("<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>")
```
hartig, line 61: Instead of hard-coding the URI of rdf:type as a string, please use the class `org.apache.jena.vocabulary.RDF`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -67,14 +47,44 @@ public static LogicalPlan rewriteToUseLocalVocabulary( final LogicalPlan inputPl
...
-				return new LogicalPlanWithUnaryRootImpl(l2g,rw);
-			} else {
+
+			// For a compressed version of query plan, check if it is necessary to add the l2g operator over a request based on the form of the graph pattern.
+			// The current implementation assumes that only concepts and roles are being considered in vocabulary mapping.
+			boolean rewriteNeeded = false;
+			if ( compressVocabularyRewriting ) {
+				final Set<TriplePattern> tps = LogicalOpUtils.getTriplePatternsOfReq(requestOp);
+				if ( !tps.isEmpty() ) {
+					for ( final TriplePattern tp : tps ) {
```
hartig, line 57: It would be better to take this whole code block that is checking whether `rewriteNeeded` should be true and move it to a separate function which you can then simply call from here.

hartig: Additionally, this function should be called only after making sure that there is a vocabulary mapping. (currently, your code is always checking whether `rewriteNeeded` should be true, even in cases in which the fed.member of the request op. doesn't have a vocab.mapping).

hartig (changes requested):

hartig (approved):
Looks much cleaner now.

PR #294, opened 2023-05-02
Reduce number of vocabulary rewriting

chengsijin0817:
Implement a plan optimizer that determines join ordering by considering vocabulary mapping and cardinality. Reduce the number of vocabulary rewriting by eliminating some l2g and g2l when the same vocabulary mapping is used.


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,267 @@
...
+            }
+
+            return subPlansWithStatistics;
+        }
+
+        /**
+         * Construct a physical plan using symmetric hash join:
+         * If both currentPlan and nextPlan have PhysicalOpLocalToGlobal as root operator, and use the same vocabulary mapping,
+         * the physical plan can be simplified as:
+         * join( l2g^{vm}(p1), l2g^{vm}(p2) ) --> l2g^{vm}( join (p1, p2) )
```
hartig, line 187: I don't think that this rewriting is correct in general. Here is a counterexample: Assume a vocab.mapping with

Local1  rdfs:subClassOf  Global
Local2  rdfs:subClassOf  Global

Assume two solution mappings:

m1 = { ?x -> Local1 }
m2 = { ?x -> Local2 }

where m1 comes from p1 and m2 comes from p2.

When applying the vocab.mapping directly to m1 and m2 (before doing the join), then the rewritten solution mappings are:

m1' = { ?x -> Global }
m2' = { ?x -> Global }

These are compatible and, thus, they join up with one another.

In contrast, when doing the join first, m1 and m2 are not compatible and, thus, they cannot be joined.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,267 @@
...
+        /**
+         * @param resps A list of cardinality, which is ordered in the order in which the request operators can be found by a depth-first traversal of the subplans.
+         * @return A list of PhysicalPlanWithStatistics, with each object corresponding to a subplan in the 'subplans' list.
+         * Each PhysicalPlanWithStatistics object contains the physical plan for the subplan, along with its cardinality,
+         * a list of relevant federations members, and estimated number of access to federation members.
+         *
+         * During a depth-first traversal of the subplans, increase the index by 1 when visiting a request operator:
+         * If the root operator of the subplan is a Request operator, get the corresponding cardinality from resps with the same index;
+         * If the subPlan with a Union of requests, calculate the total cardinality by summing up the cardinality of individual requests.
+         * (See page 1052 in the paper[1])
```
hartig, line 128: [1] ?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,267 @@
...
+            try {
+                resps = FederationAccessUtils.performCardinalityRequests(fedAccessMgr, reqOpsOfAllSubPlans);
+            } catch (final FederationAccessException e) {
+                throw new PhysicalOptimizationException("Issuing a cardinality request caused an exception.", e);
+            }
+
+            // Then, annotate each subplan (in the 'subplans') with all statistics relevant for the join planning
+            // ( including cardinality, number of access, and a list of relevant federation members )
+            final List<PhysicalPlanWithStatistics> subPlansWithStatistics = associateCardWithSubPlans(resps);
+
```
hartig, line 75: This looks very familiar. Don't we have this code (and also the following lines) somewhere else already? If so, can you try to extract the common parts into a shared superclass method? Or some helper class?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,267 @@
...
+         * @return A list of PhysicalPlanWithStatistics, with each object corresponding to a subplan in the 'subplans' list.
+         * Each PhysicalPlanWithStatistics object contains the physical plan for the subplan, along with its cardinality,
+         * a list of relevant federations members, and estimated number of access to federation members.
+         *
+         * During a depth-first traversal of the subplans, increase the index by 1 when visiting a request operator:
+         * If the root operator of the subplan is a Request operator, get the corresponding cardinality from resps with the same index;
+         * If the subPlan with a Union of requests, calculate the total cardinality by summing up the cardinality of individual requests.
+         * (See page 1052 in the paper[1])
+         */
+        protected List<PhysicalPlanWithStatistics> associateCardWithSubPlans(
```
hartig, line 130: Is this the same function as in `CardinalityBasedGreedyJoinPlanOptimizerImpl`? If so, can you reuse it instead of copying the code?

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
chengsijin0817: Thank you very much for your explanation. Indeed! This rewriting is incorrect. 
I have removed it.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
chengsijin0817: I have extracted the common parts into a superclass: 'CardinalityBasedJoinPlanOptimizerBase'

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public EnumerationAlgorithm initializeEnumerationAlgorithm( final List<PhysicalPlan> subplans ) {
+
+        return new CardinalityBasedJoinPlanOptimizerBase.GreedyConstructionAlgorithm(subplans) {
+            @Override
+            protected PhysicalPlan determineJoinOrderAndConstructPlan( final PhysicalPlanWithStatistics firstSubPlan, final List<PhysicalPlanWithStatistics> subPlansWithStatistics, final Comparator<PhysicalPlanWithStatistics> orderBasedOnCard) {
+                final PriorityQueue<PhysicalPlanWithStatistics> orderedCandidateSubPlans = new PriorityQueue<>(orderBasedOnCard);
+                PhysicalPlanWithStatistics nextSelectedSubPlan = firstSubPlan;
```
hartig, line 37: Shouldn't this better be called `previousSelectedSubPlan`?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/CardinalityBasedJoinPlanOptimizerBase.java
```diff
@@ -0,0 +1,262 @@
...
+                final PhysicalOperator pop = subplan.getRootOperator();
+
+                PhysicalPlanWithStatistics planWithStatistics = null;
+                if ( pop instanceof PhysicalOpRequest
+                        || (pop instanceof PhysicalOpFilter && subplan.getSubPlan(0).getRootOperator() instanceof PhysicalOpRequest)) {
+                    final int cardinality = resps[index].getCardinality();
+                    final FederationMember fm = reqOpsOfAllSubPlans.get(index).getFederationMember();
+                    final int numOfAccess = accessNumForReq(cardinality, fm);
+
+                    planWithStatistics = new PhysicalPlanWithStatistics( subplan, Arrays.asList(fm), null, cardinality, numOfAccess );
```
hartig, line 165: Shouldn't the vocab.mapping of the fed.member (if any) be passed to the constructor as well?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+                PhysicalPlan currentJoinPlan = nextSelectedSubPlan.plan;
+
+                while ( !subPlansWithStatistics.isEmpty() ){
+                    // Identify subplans that have join variables with the selected subplan as new candidateSubPlans.
+                    List<PhysicalPlanWithStatistics> tmpCandidateSubPlans = getSubPlansContainVars( currentJoinPlan.getExpectedVariables(), subPlansWithStatistics );
+                    if( tmpCandidateSubPlans.isEmpty() ){
+                        tmpCandidateSubPlans = subPlansWithStatistics;
+                    }
+
+                    // Keep candidate SubPlans that use the same vocabulary mapping with the previous selected SubPlan.
```
hartig, line 47: What if the previously selected subplan has multiple vocabulary mappings associated with it? Would that still be okay here?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+                    orderedCandidateSubPlans.addAll(candidateSubPlans);
+
+                    // select the first element (with the lowest cardinality) from orderedCandidateSubPlans as the nextSelectedSubPlan
+                    // Remove the selected subPlan from subPlansWithStatistics and clear orderedCandidateSubPlans for the next loop
+                    nextSelectedSubPlan = orderedCandidateSubPlans.poll();
+                    subPlansWithStatistics.remove(nextSelectedSubPlan);
+                    orderedCandidateSubPlans.clear();
+
+                    // Construct a physical plan using bind join
+                    currentJoinPlan = constructPlanWithBJ( currentJoinPlan, nextSelectedSubPlan.plan );
```
hartig, line 64: Related to my previous comment about calling `nextSelectedSubPlan` `previousSelectedSubPlan`, here it makes sense to call this program variable `nextSelectedSubPlan`. Therefore, for the sake of the readability of the code, I would use `nextSelectedSubPlan` as a `final` variable here inside the loop and, then, at the end of the loop assign `nextSelectedSubPlan` to `previousSelectedSubPlan`.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Iterate through the remaining subplans and selected those that have the same vocabulary mapping in the given list of vocabulary mappings.
+     */
+    protected List<PhysicalPlanWithStatistics> getSubPlansWithSameVoc( final List<VocabularyMapping> vms, final List<PhysicalPlanWithStatistics> subPlansWithStatistics ) {
+        final List<PhysicalPlanWithStatistics> subPlansWithSameVoc = new ArrayList<>();
+        for ( final PhysicalPlanWithStatistics subplan: subPlansWithStatistics ) {
+            final List<VocabularyMapping> vmsNext = subplan.getVocabularyMappings();
+            vms.retainAll(vmsNext);
```
hartig, line 79: This changes `vms` which is not what you want. Instead, can you swap it (`vmsNext.retainAll(vms)`)?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        return subPlansWithSameVoc;
+    }
+
+    /**
+     * Construct a physical plan using bind join when it is possible:
+     * If both currentPlan and nextPlan have PhysicalOpLocalToGlobal as root operator, and use the same vocabulary mapping,
+     * the physical plan can be simplified as when unary operator can be applied:
+     * tpAdd^{tp}_{fm}(g2l^{vm}(l2g^{vm}(p))) --> tpAdd^{tp}_{fm}(p)
```
hartig, line 92: While this equivalence is correct, the function is not just doing that but, instead, it is actually converting a binary join into a tpAdd. I mean, in contrast to the formula given here, there is no g2l operator anywhere in the two arguments (`currentPlan` and `nextPlan`).

Since the function is actually converting a binary join into a tpAdd, it is not correct in general. Instead, it has still the same issue as mentioned in the comments of my previous review for this PR. In the same way, the equivalences that you have in Proposition 1 of the paper draft are also not correct in general. They only hold for specific vocabulary mappings.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
hartig: I take back what I said about the equivalences in the paper draft. Given that they are only for triple patterns that cannot produce solution mappings with IRIs of classes or properties, they are correct. However, I would like both the equivalences and the implementation to be less restrictive. In particular, the equivalences can also be used for vocab.mappings that do not use subclass-based rules, in which case the patterns of the query can be anything. Should I put what I mean into the paper draft?

chengsijin0817: Hi Olaf, I would appreciate it if you could put your intended meaning in the paper draft. 
Thanks.

hartig: Done now. Sorry for the delay! Take a look and email me if you have questions or comments about it.

chengsijin0817: Hi Olaf, I sincerely appreciate you taking time out of your tight schedule to add text. The text described is very clear, and I have only a minor question. 
Specifically, I am uncertain if the first equivalence of Proposition 4 is necessary because it can be further simplified to join(\varphi, \varphi'). In other words, it appears to imply the same rule as the second proposition.

Regarding the implementation of plan rewriting, I have not found a good way to verify if every rule in the vocabulary mapping is either an equivalence on concepts or an equivalence on roles. Do you have any suggestions?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
chengsijin0817: I suppose the previously selected sub-plan has multiple vocabulary mappings only when this sub-plan has union as the root operator. Given that these candidate plans will be utilized to determine the join order, selecting a sub-plan that has some common vocabulary mapping may increase the likelihood of creating a plan with tpAdd that facilitates a reduction in plan rewriting operations. What do you think?

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/simple/ReduceVocRewritingJoinPlanOptimizerImpl.java
hartig: > I am uncertain if the first equivalence of Proposition 4 is necessary because it can be further simplified to join(\varphi, \varphi'). In other words, it appears to imply the same rule as the second proposition.

Indeed, that equivalence is redundant because of the second equivalence. Great observation! I have removed it.

> I have not found a good way to verify if every rule in the vocabulary mapping is either an equivalence on concepts or an equivalence on roles.

The issue is in the `VocabularyMappingImpl` class, right? I suggest that you extend the `VocabularyMapping` interface with a `boolean`-typed function called `isEquivalenceOnly()`. Then, in the `VocabularyMappingImpl` class, you implement this function by checking whether the `Graph vocabularyMapping` does not contain triples with predicates rdfs:subClassOf, rdfs:subPropertyOf, owl:inverseOf, owl:unionOf, and owl:intersectionOf. Use the corresponding Java classes for the RDFS and the OWL vocabulary (as done in other parts of the file).

hartig:
@chengsijin0817 What is the status of this PR? I see that the last thing that happened to it is that you added some commits, but since you didn't explicitly request my review again, I was assuming that you are not done yet with addressing my previous review. Are you?

chengsijin0817:
> @chengsijin0817 What is the status of this PR? I see that the last thing that happened to it is that you added some commits, but since you didn't explicitly request my review again, I was assuming that you are not done yet with addressing my previous review. Are you?

Hi @hartig, I think this branch (implementation of a physical plan optimizer) was discarded after we created a join order optimizer for logical plans (i.e., [CardinalityBasedJoinOrderingWithRequests](https://github.com/LiUSemWeb/HeFQUIN/blob/main/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/CardinalityBasedJoinOrderingWithRequests.java)) and implemented different vocabulary-related rewriting rules as separate heuristics that can be applied in [LogicalOptimizerImpl](https://github.com/LiUSemWeb/HeFQUIN/blob/main/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/LogicalOptimizerImpl.java). The related heuristics are:
[RemoveUnnecessaryL2gAndG2l](https://github.com/LiUSemWeb/HeFQUIN/blob/main/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveUnnecessaryL2gAndG2l.java) 
[PullUpLtgOverUnion](https://github.com/LiUSemWeb/HeFQUIN/blob/main/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/PullUpLtgOverUnion.java)
[PullUpLtgOverJoin](https://github.com/LiUSemWeb/HeFQUIN/blob/main/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/PullUpLtgOverJoin.java)
[PushJoinUnderUnionWithRequests](https://github.com/LiUSemWeb/HeFQUIN/blob/main/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/PushJoinUnderUnionWithRequests.java)
[RemovePairsOfG2lAndL2g](https://github.com/LiUSemWeb/HeFQUIN/blob/main/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemovePairsOfG2lAndL2g.java)

hartig:
> I think this branch (implementation of a physical plan optimizer) was discarded after ...

In other words, it can be safely removed? Please confirm again and I will remove it.

chengsijin0817:
> > I think this branch (implementation of a physical plan optimizer) was discarded after ...
> 
> In other words, it can be safely removed? Please confirm again and I will remove it.

Yes, I confirm that this branch can be safely removed. Thanks.

hartig:
Thanks for the confirmation. I am closing this PR and removing its branch now.

PR #296, opened 2023-05-02
Extend cost model to consider vocabulary mapping

chengsijin0817:
Now applying vocabulary mappings is supported by all available algorithms (except for the evolutionary algorithm)

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/LogicalOptimizerImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/LogicalOptimizerImpl.java
@@ -35,7 +35,7 @@ public class LogicalOptimizerImpl implements LogicalOptimizer
 		//// Uncomment the following line to apply vocabulary mappings of
 		//// federation members during query planning. Note that applying
 		//// vocabulary mappings is not supported by the evolutionary algorithm
-		//// currently since the rewriting rules needs to be extended to consider
+		//// currently since the rewriting rules need to be extended to consider
 		//// operators PhysicalOpLocalToGlobal and PhysicalOpGlobalToLocal.
 		//heuristics.add( new ApplyVocabularyMappings());
 	}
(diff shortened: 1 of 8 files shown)


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/LogicalOptimizerImpl.java
```diff
@@ -34,8 +34,8 @@ public LogicalOptimizerImpl() {
 
 		//// Uncomment the following line to apply vocabulary mappings of
 		//// federation members during query planning. Note that applying
-		//// vocabulary mappings is only supported by the naive algorithm
-		//// currently since the cost model needs to be extended to consider
+		//// vocabulary mappings is not supported by the evolutionary algorithm
+		//// currently since the rewriting rules needs to be extended to consider
```
hartig, line 38:
```suggestion
		//// currently since the rewriting rules need to be extended to consider
```

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/poptimizer/costmodel/CFRNumberOfTermsShippedInRequests.java
```diff
@@ -33,20 +33,17 @@ public CompletableFuture<Integer> initiateCostEstimation( final PhysicalPlan pla
...
 			} else if (lop instanceof LogicalOpBGPAdd) {
 				pattern = ((LogicalOpBGPAdd) lop).getBGP();
 
-			} else if (lop instanceof LogicalOpGPAdd) {
+			} else {
 				pattern = ((LogicalOpGPAdd) lop).getPattern();
 			}
-			else {
-				throw createIllegalArgumentException(lop);
-			}
```
hartig, line 49: Please revert back to the original version of this part -- just as an additional safety measure in case we change something here later.

hartig (changes requested):
Two very minor things.

PR #297, opened 2023-05-10
Extend the VocabularyMapping

chengsijin0817:
Extend the VocabularyMapping interface with a boolean-typed function called isEquivalenceOnly()

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
@@ -3,6 +3,7 @@ package se.liu.ida.hefquin.engine.data;
 import java.util.Set;
 
 import se.liu.ida.hefquin.engine.query.SPARQLGraphPattern;
+import se.liu.ida.hefquin.engine.query.SPARQLUnionPattern;
 import se.liu.ida.hefquin.engine.query.TriplePattern;
 
 public interface VocabularyMapping
@@ -40,5 +41,8 @@ public interface VocabularyMapping
 	 * given solution mapping.
 	 */
 	Set<SolutionMapping> translateSolutionMappingFromGlobal( SolutionMapping sm );
+
+
+	boolean isEquivalenceOnly();
 
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
@@ -43,6 +43,10 @@ public interface VocabularyMapping
 	Set<SolutionMapping> translateSolutionMappingFromGlobal( SolutionMapping sm );
 
 
+	/**
+	 * Checks whether the vocabulary mapping only contains equivalence rules, e.g., owl:equivalentClass, owl:equivalentProperty, owl:sameAs
+	 * Return ture if the vocabulary mapping consists solely of equivalence rules, false otherwise.
+	 */
 	boolean isEquivalenceOnly();
 
 }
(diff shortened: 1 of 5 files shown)


src/main/java/se/liu/ida/hefquin/engine/data/VocabularyMapping.java
```diff
@@ -41,4 +42,7 @@
 	 */
 	Set<SolutionMapping> translateSolutionMappingFromGlobal( SolutionMapping sm );
 
+
+	boolean isEquivalenceOnly();
```
hartig, line 46: Please add a JavaDoc comment that specifies what this function is about.

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -431,30 +431,31 @@ public Set<SolutionMapping> translateSolutionMappingFromGlobal( final SolutionMa
...
-						if(!k.isEmpty()) {
-							translationCopy.addAll(k.snapshot());
+			else {
+				final Set<Node> bindingTranslation = translateBindingFromGlobal(n);
+				if (bindingTranslation.size() > 1) {
+					final Set<BindingBuilder> bbsCopy = new HashSet<>();
+
+					for (final Node j : bindingTranslation) {
+						for (final BindingBuilder k : bbs) {
+							BindingBuilder translationCopy = BindingBuilder.create();
```
hartig, line 441:
```suggestion
							final BindingBuilder translationCopy = BindingBuilder.create();
```

src/main/java/se/liu/ida/hefquin/engine/data/impl/VocabularyMappingImpl.java
```diff
@@ -465,7 +466,24 @@ public Set<SolutionMapping> translateSolutionMappingFromGlobal( final SolutionMa
 		}
 		return results;
 	}
-	
+
+	@Override
+	public boolean isEquivalenceOnly() {
+		for (final Triple m : getMappings(Node.ANY, Node.ANY, Node.ANY)){
```
hartig, line 472: Calling the `getMappings` function is very inefficient, in particular if you call it with all three positions unbound. The issue is that this function then simply copies the complete graph into another `Set` and returns that. (But, of course, that's not your fault!)  In fact, we should get rid of this function in this class. In the meantime, however, it is better not to use it. Instead, the iterator-based approach as that function uses internally can be used directly here.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/VocabularyMappingWrappingImpl.java
```diff
@@ -72,4 +72,10 @@ public Set<SolutionMapping> translateSolutionMappingFromGlobal( final SolutionMa
 		}
 		return result;
 	}
+
+	@Override
+	public boolean isEquivalenceOnly() {
+		return false;
```
hartig, line 78: No. Instead of simply returning false, throw an UnsupportedOperationException.

src/main/java/se/liu/ida/hefquin/engine/federation/catalog/FederationDescriptionReader.java
```diff
@@ -174,7 +175,12 @@ protected VocabularyMapping parseVocabMapping( final Resource fm, final Model fd
 
 			final String path = pathToMappingFile.toString();
 			if ( verifyValidVocabMappingFile(path) ) {
-				return new VocabularyMappingImpl(path);
+				VocabularyMapping vm = vocabMappingByPath.get( path );
+				if ( vm == null ) {
+					vm = new VocabularyMappingImpl(path);
+					vocabMappingByPath.put( path, vm );
+				}
+				return vm;
```
hartig, line 183: Good idea!!

hartig (changes requested):
Just a few minor things.

hartig (approved):

PR #298, opened 2023-05-22
Implementation of schema mapping

chengsijin0817:


hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,286 @@
...
 import se.liu.ida.hefquin.engine.query.TriplePattern;
+import se.liu.ida.hefquin.engine.query.impl.SPARQLUnionPatternImpl;
+import se.liu.ida.hefquin.engine.query.impl.TriplePatternImpl;
+import se.liu.ida.hefquin.engine.utils.Pair;
 
 public class SchemaMappingImpl implements SchemaMapping
 {
-	// TODO: define the interface TermMapping and implementing classes for the various options
-	//protected final Map<Node, Set<TermMapping>> g2lMap = new HashMap<>();
+	protected final Graph mappingDescription;
```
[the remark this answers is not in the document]
chengsijin0817: Hello Olaf, I kept this member variable as it is used in the function 'getConnectedMapping' for retrieving all local terms of mapping rules that are associated with the 'unionOf'.

For instance, in the Jena.graph, the RDF triple "ex:n owl:unionOf (ex:f ex:g)." is parsed as follows:
```
 { 
 e6127389-5f6f-4173-a1ff-8aac98ad63bd @rdf:rest f4152bb4-8a81-4896-9d02-a4dacc7bbed8; 
 e6127389-5f6f-4173-a1ff-8aac98ad63bd @rdf:first ex:f; 
 f4152bb4-8a81-4896-9d02-a4dacc7bbed8 @rdf:rest rdf:nil; 
 f4152bb4-8a81-4896-9d02-a4dacc7bbed8 @rdf:first ex:g; 
 ex:n @owl:unionOf e6127389-5f6f-4173-a1ff-8aac98ad63bd; 
 }
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,286 @@
...
+	 * Only apply equivalent rules
+	 */
+	public Set<Node> applyMapToGlobalURI( final Node node, final Map<Node, Set<TermMapping>> vocMap ) {
+		final Set<TermMapping> termMappings = vocMap.get(node);
+		final Set<Node> results = new HashSet<>();
+		if ( termMappings != null ) {
+			for (TermMapping mappingTypes : termMappings) {
+				if ( mappingTypes.getTypeOfRule().equals(OWL.equivalentClass.asNode()) || mappingTypes.getTypeOfRule().equals(OWL.equivalentProperty.asNode()) ) {
+					results.addAll(mappingTypes.getTranslatedTerms());
+				}
```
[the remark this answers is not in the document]
chengsijin0817: Hi, I have fixed the issue regarding cases where no terms are added to the results (in [commit](https://github.com/LiUSemWeb/HeFQUIN/pull/298/commits/086062958db6bf390603ce9f2342ceaf9c694306)).

To avoid this issue, I added the original binding to the results if no mapping rules can be applied. 
In addition, I didn't include an else branch since the g2lMap contains all different types of term mappings. If a mapping rule does not fall within the equivalence types, it should be skipped rather than triggering an exception.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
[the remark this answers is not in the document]
hartig: I don't understand. `getConnectedMapping` is a helper function that you use only (!) when parsing the RDF description to populate the data structures in the constructor. You can simply pass the `Graph` object to `getConnectedMapping`.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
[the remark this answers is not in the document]
hartig: > In addition, I didn't include an else branch since the g2lMap contains all different types of term mappings. If a mapping rule does not fall within the equivalence types, it should be skipped rather than triggering an exception.

I see your reasoning. However, the only cases in which this whole function (`applyMapToGlobalURI`)  should be called (according to the rest of your implementation) is from `applyMapToSolutionMapping` with `useInverse=false`. And that one should be called only from the G2L operator, and the G2L operator is safe only for equivalence-only vocab.mappings. Therefore, I was suggesting to throw the exception as an additional safety measure, just to make sure that we will see error messages if we use these functions in the future in a different way than currently anticipated.

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,271 @@
...
-	// TODO: define the interface TermMapping and implementing classes for the various options
-	//protected final Map<Node, Set<TermMapping>> g2lMap = new HashMap<>();
+	protected final Graph mappingDescription;
+	protected final Map<Node, Set<TermMapping>> g2lMap = new HashMap<>();
+	protected final Map<Node, Set<TermMapping>> l2gMap = new HashMap<>();
+	boolean isEquivalenceOnly = true;
 
 	public SchemaMappingImpl( final Graph mappingDescription ) {
+		this.mappingDescription = mappingDescription;
 		parseMappingDescription(mappingDescription);
```
hartig, line 34:
```suggestion
	protected final boolean isEquivalenceOnly;

	public SchemaMappingImpl( final Graph mappingDescription ) {
		this.mappingDescription = mappingDescription;
		this.isEquivalenceOnly = parseMappingDescription(mappingDescription);
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,271 @@
...
 	}
 
-	protected void parseMappingDescription( final Graph mappingDescription ) {
-		// TODO: implement this initialization function
+	protected boolean parseMappingDescription( final Graph mappingDescription ) {
```
hartig, line 37:
```suggestion
	protected boolean parseMappingDescription( final Graph mappingDescription ) {
		boolean isEquivalenceOnly = true;
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,271 @@
...
+		// Populate l2gMap
+		for ( final Node localTerm: local ) {
+			Set<TermMapping> l2gtermMappings = l2gMap.get(localTerm);
+
+			if (l2gtermMappings == null) {
+				l2gtermMappings = new HashSet<>();
+				l2gMap.put(localTerm, l2gtermMappings);
+			}
+
+			TermMappingImpl globalTerms = new TermMappingImpl(mappingType, global);
```
hartig, line 93:
```suggestion
			final TermMappingImpl globalTerms = new TermMappingImpl(mappingType, global);
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,271 @@
...
+					tmpBindings.add( newBinding.build() );
+				}
+			}
+			resultsBindings.clear();
+			resultsBindings.addAll( tmpBindings );
+			tmpBindings.clear();
+		}
+
+		final Set<SolutionMapping> resultsSolMaps = new HashSet<>();
+		for ( Binding b: resultsBindings ){
```
hartig, line 191:
```suggestion
		for ( final Binding b: resultsBindings ){
```

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,273 @@
...
+		final Set<TermMapping> termMappings = l2gMap.get(node);
+		final Set<Node> results = new HashSet<>();
+		if ( termMappings != null ) {
+			for ( final TermMapping mappingTypes : termMappings ) {
+				results.addAll( mappingTypes.getTranslatedTerms() );
+			}
+		}
+		else {
+			results.add(node);
+		}
```
hartig, line 206:
```suggestion
		if ( termMappings == null ) {
			return Collections.singleton(node);
		}

		final Set<Node> results = new HashSet<>();
		for ( final TermMapping mappingTypes : termMappings ) {
			results.addAll( mappingTypes.getTranslatedTerms() );
		}
```

src/main/java/se/liu/ida/hefquin/engine/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -1,44 +1,273 @@
...
+				}
+				else  {
+					throw new IllegalArgumentException();
+				}
+			}
+		}
+
+		if ( results.isEmpty() ){
+			results.add(node);
+		}
```
hartig, line 230:
```suggestion
		if ( termMappings == null ) {
			return Collections.singleton(node);
		}

		final Set<Node> results = new HashSet<>();
		for ( final TermMapping mappingTypes : termMappings ) {
			if (    mappingTypes.getTypeOfRule().equals(OWL.equivalentClass.asNode())
			     || mappingTypes.getTypeOfRule().equals(OWL.equivalentProperty.asNode()) ) {
				results.addAll( mappingTypes.getTranslatedTerms() );
			}
			else  {
				throw new IllegalArgumentException();
			}
		}
```

hartig (approved):

PR #302, opened 2023-05-23
Remove unnecessary l2g and g2l operators

chengsijin0817:


src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveUnnecessaryL2gAndG2l.java
```diff
@@ -0,0 +1,109 @@
...
+					if( !currentTPs.isEmpty() && previousTPs != null && !currentTPs.equals( previousTPs) ) {
+						throw new IllegalArgumentException("UNION is not added as a result of source selection");
+					}
+					previousTPs = currentTPs;
+				}
+				else {
+					throw new IllegalArgumentException("Unsupported type of subquery under UNION (" + subLop.getClass().getName() + ")");
+				}
+			}
+			return previousTPs;
```
hartig, line 99: I don't understand this part. Why don't you simply call this function recursively for all subplans under the union (and collect all the triple patterns in a set that can then be returned)?

This rewriting rule can be general, for any type of FedQPL expression and not just source assignments.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveUnnecessaryL2gAndG2l.java
```diff
@@ -0,0 +1,109 @@
...
+		}
+		else if( lop instanceof LogicalOpFilter || lop instanceof LogicalOpLocalToGlobal || lop instanceof LogicalOpGlobalToLocal ) {
+			return extractTPs( plan.getSubPlan(0) );
+		}
+		else {
```
hartig, line 104: Somewhat related to my other comment, this rewriting rule should be general. Hence, you should include else-if-branches also for the other types of logical operators.

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveUnnecessaryL2gAndG2l.java
```diff
@@ -0,0 +1,109 @@
...
+			if ( tp.asJenaTriple().getPredicate().equals(RDF.Nodes.type)
+					&& tp.asJenaTriple().getObject().isVariable() ) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	protected static Set<TriplePattern> extractTPs( final LogicalPlan plan ) {
+		final LogicalOperator lop = plan.getRootOperator();
```
hartig, line 76: I suggest to rename this variable to `rootOp`.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveUnnecessaryL2gAndG2l.java
```diff
@@ -73,36 +72,23 @@ protected boolean checkIfL2gOrG2lNeeded( final LogicalPlan plan ){
...
-					final Set<TriplePattern> currentTPs = LogicalOpUtils.getTriplePatternsOfReq( (LogicalOpRequest<?, ?>) subLop);
-					if( !currentTPs.isEmpty() && previousTPs != null && !currentTPs.equals( previousTPs) ) {
-						throw new IllegalArgumentException("UNION is not added as a result of source selection");
-					}
-					previousTPs = currentTPs;
-				}
-				else {
-					throw new IllegalArgumentException("Unsupported type of subquery under UNION (" + subLop.getClass().getName() + ")");
-				}
+		else if ( rootOp instanceof BinaryLogicalOp || rootOp instanceof NaryLogicalOp) {
```
hartig, line 80: It would be better to list the Java class of each operator explicitly. That way, we are sure to throw the IllegalArgumentException whenever we add new types of operators and can then think about whether we need to do something special for them here.

hartig (changes requested):

hartig (approved):
Great. Thanks!

PR #306, opened 2023-08-22
Adds CLI to send BGPs to a Neo4j Server

scferrada:
This does add some extra code which might be unwanted (e.g., changes in the way queries are printed specially, and some tests).
It also changes a Node type for a Var type everywhere in the translation, which I don't remember why was needed.

I'll use the review tool to do some clean up, and then will assign @hartig to review.


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/utils/LabeledGraph.java
```diff
@@ -189,6 +189,13 @@ private Path longestStartingFrom(final CypherVar start) {
                 candidate = new Path(start, currentEdge);
             }
             else if (! visitedEdges.contains(currentEdge.id)) {
+                /*if (adjacencyLists.get(candidate.path.get(candidate.path.size()-1).target).stream().noneMatch(
+                        e -> e.edge.equals(currentEdge.edge) && e.target.equals(currentEdge.target)
+                )) {
+                    System.out.println("ERROR: " + currentEdge);
+                    System.out.println("node:" + candidate.path.get(candidate.path.size()-1).target);
+                    System.out.println("doesnt connect with: " + currentEdge.target);
+                }*/
```
scferrada, line 198:
```suggestion
```

scferrada:
Okay, the Node->Var migration was to avoid doing several castings.

This PR also adds a CLI to run Cypher queries over Neo4J.

src/main/java/se/liu/ida/hefquin/cli/modules/ModQuery.java
```diff
@@ -8,7 +8,7 @@
 public class ModQuery extends ModQueryIn
 {
     public ModQuery() {
-        super( Syntax.syntaxSPARQL_11 );
+        super( Syntax.syntaxARQ );
```
hartig, line 11:
```suggestion
        super( Syntax.syntaxARQ ); // we need syntaxARQ instead of syntaxSPARQL_11 in order to get SPARQL-star features
```

hartig (approved):
Looks good. Thanks!

PR #308, opened 2023-09-15
Add flexible LPG2RDF Configuration

BentolhodaH:


src/main/java/se/liu/ida/hefquin/cli/RunBGPOverNeo4j.java
```diff
@@ -56,6 +57,8 @@ public class RunBGPOverNeo4j extends CmdARQ
 	protected final ArgDecl argNoMerge   = new ArgDecl(ArgDecl.NoValue, "disablePathMerging");
 	protected final ArgDecl argSuppressResultPrintout = new ArgDecl(ArgDecl.NoValue, "suppressResultPrintout");
 
+	protected final String LABEL = "http://www.example.org/2000/01/rdf-schema#label";
```
hartig, line 60: That's the wrong URI. Please use exactly the one that was/is used in `DefaultConfiguration.java`. Moreover, this member variable can also be `static` now (in addition to being `final`).

hartig (changes requested):
Thanks for the PR. There are a few things that I would like you to change: see my comments in the "Files changed" tab of this PR. Most of these comments (all except for one) contain explicit change suggestions. You can directly commit this suggestions within the "Files changed" view (i.e., instead of having to reproduce them in your local code first). Only the other one comment is something you need to do in your local code (and remember: before you do that, do a `git pull` to pull the changes from the change suggestions after you have committed them; and later after doing the changes for the one other comment locally on your computer, you have to commit these changes locally and then push the branch again, so that these changes also show up here in the PR).

BentolhodaH:
Thanks for your comments.
I changed all of them.

hartig (approved):

PR #310, opened 2023-09-20
Encapsulate map/unmap Node

BentolhodaH:
Encapsulate map/unmap Node.
Add unit test for NodeMappingToUTIsImpl and NodeMappingToURIsImpl class.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
@@ -23,12 +23,12 @@ public class LPG2RDFConfigurationImpl implements LPG2RDFConfiguration {
 
     @Override
     public Node mapNode(final LPGNode node) {
-        return this.nodeMapping.mapNode(node);
+        return nodeMapping.map(node);
     }
 
     @Override
     public LPGNode unmapNode(final Node node) {
-        return this.unmapNode(node);
+        return nodeMapping.unmap(node);
     }
 
     @Override
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
@@ -8,7 +8,6 @@ import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
 public class LPG2RDFConfigurationImpl implements LPG2RDFConfiguration {
 
     protected final String NS = "https://example.org/";
-    protected final String NODE = "node/";
     protected final String NODELABEL = "label/";
     protected final String RELATIONSHIP = "relationship/";
     protected final String PROPERTY = "property/";
@@ -100,6 +99,6 @@ public class LPG2RDFConfigurationImpl implements LPG2RDFConfiguration {
 
     @Override
     public boolean mapsToNode(final Node n) {
-        return n.isURI() && n.getURI().startsWith(NS + NODE);
+        return nodeMapping.isPossibleResult(n);
     }
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
@@ -14,25 +14,21 @@ public class LPG2RDFConfigurationImpl implements LPG2RDFConfiguration {
     protected final String PROPERTY = "property/";
 
     protected final Node label;
+    protected final NodeMapping nodeMapping;
 
-    public LPG2RDFConfigurationImpl(final Node label){
+    public LPG2RDFConfigurationImpl(final Node label, final NodeMapping nodeMapping){
         this.label = label;
+        this.nodeMapping = nodeMapping;
     }
-
 
     @Override
     public Node mapNode(final LPGNode node) {
-        return NodeFactory.createURI(NS + NODE + node.getId());
+        return this.nodeMapping.mapNode(node);
     }
 
     @Override
     public LPGNode unmapNode(final Node node) {
-        if (!node.isURI())
-            throw new IllegalArgumentException("Default configuration only accepts URI Node mappings");
-        if (!node.getURI().startsWith(NS + NODE))
-            throw new IllegalArgumentException("The provided URI is not mapping a Node");
-        final String id = node.getURI().replaceAll(NS + NODE, "");
-        return new LPGNode(id, "", null);
+        return this.unmapNode(node);
     }
 
     @Override
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMapping.java
@@ -28,9 +28,9 @@ public interface NodeMapping {
 
 
     /**
-     *Check if the given RDF term is in the image of this node mapping and,
-     *thus, may be one of the RDF terms returned by the {@link #map(LPGNode)}
-     *function for some LPG node.
+     * Check if the given RDF term is in the image of this node mapping and,
+     * thus, may be one of the RDF terms returned by the {@link #map(LPGNode)}
+     * function for some LPG node.
      */
     boolean isPossibleResult(Node node);
 }
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMapping.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMapping.java
@@ -1,0 +1,11 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import org.apache.jena.graph.Node;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
+
+public interface NodeMapping {
+
+    public Node mapNode(final LPGNode node);
+
+    public LPGNode unmapNode(final Node node);
+}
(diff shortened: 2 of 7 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMappingToURIsImpl.java
```diff
@@ -0,0 +1,26 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
+
+public class NodeMappingToURIsImpl implements NodeMapping{
+
+    protected final String NS = "https://example.org/";
+    protected final String NODE = "node/";
```
hartig, line 10: Can you please make just one string out of these two and, then, add a constructor to this class via which this string can be passed to the class (i.e., similar to how the `label` IRI is passed to `LPG2RDFConfigurationImpl`).

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMappingToURIsImpl.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Override
+    public Node mapNode(final LPGNode node) {
+        return NodeFactory.createURI(NS + NODE + node.getId());
+    }
+
+    @Override
+    public LPGNode unmapNode(final Node node) {
+        if (!node.isURI())
+            throw new IllegalArgumentException("LPG2RDF configuration only accepts URI Node mappings");
```
hartig, line 20: Please change the message of this exception in a way similar to what I suggested for the blank-node-related case in the other class.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMappingToURIsImpl.java
```diff
@@ -0,0 +1,26 @@
...
+        if (!node.getURI().startsWith(NS + NODE))
+            throw new IllegalArgumentException("The provided URI is not mapping a Node");
```
hartig, line 22:
```suggestion
            throw new IllegalArgumentException("The given IRI (" + node.getURI() + ") is not in the image of this node mapping.");
```

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/NodeMappingToBNodesImplTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.Test;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl.NodeMapping;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl.NodeMappingToBNodesImpl;
+
+import static org.junit.Assert.*;
+
+public class NodeMappingToBNodesImplTest {
+    protected final String NS = "https://example.org/";
+    protected final String NODE = "node/";
```
hartig, line 14: You can concatenate these two into a single string and, then, move it into the one test function that uses it (i.e., no need to have this as a member variable in this class).

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/NodeMappingToURIsImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.LPGNode;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl.NodeMapping;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl.NodeMappingToURIsImpl;
+import org.apache.jena.graph.Node;
+
+import static org.junit.Assert.*;
+
+public class NodeMappingToURIsImplTest {
+    protected final String NS = "https://example.org/";
+    protected final String NODE = "node/";
```
hartig, line 14: You can concatenate these two into a single string.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
```diff
@@ -14,25 +14,21 @@ public class LPG2RDFConfigurationImpl implements LPG2RDFConfiguration {
...
+    public LPG2RDFConfigurationImpl(final Node label, final NodeMapping nodeMapping){
         this.label = label;
+        this.nodeMapping = nodeMapping;
     }
 
-
     @Override
     public Node mapNode(final LPGNode node) {
-        return NodeFactory.createURI(NS + NODE + node.getId());
+        return this.nodeMapping.mapNode(node);
```
hartig, line 26: No need for `this.`
```suggestion
        return nodeMapping.mapNode(node);
```

hartig (changes requested):
Looks already very good! I just have several minor requests for changes.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
```diff
@@ -14,25 +14,21 @@ public class LPG2RDFConfigurationImpl implements LPG2RDFConfiguration {
     protected final String PROPERTY = "property/";
 
     protected final Node label;
+    protected final NodeMapping nodeMapping;
```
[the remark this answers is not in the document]
BentolhodaH: I think variable NODE is required because it is used in mapsToNode method.

BentolhodaH:
I have applied all the comments except the last one.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationImpl.java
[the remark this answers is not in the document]
hartig: Oh, right. I didn't remember we had such a function as well.

Now, this means the new `NodeMapping` interface that you have created for this PR should actually have a third function.

This function may be called `isPossibleResult` and the JavaDoc for it can say something like: "Returns true if the given RDF term is in the image of this node mapping and, thus, may be one of the RDF terms returned by the {@link #map(LPGNode)} function."

Once you have added this function, you can use it to implement the `mapsToNode` method.

Then, you will also need to add an implementation for this new function to each of the two classes that implement the `NodeMapping` interface. The implementation for the IRI-version can be like the current content of the `mapsToNode` method and the implementation for the blank node-version just needs to check that the given RDF term is a blank node.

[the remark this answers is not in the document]
BentolhodaH: I added the isPossibleResult method.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMapping.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Returns the LPG node that corresponds to the given RDF term.
+     * It applies the inverse of this node mapping to the given RDF term (which is assumed to be a blank node or an IRI) in order to obtain the corresponding LPG node.
+     */
+    LPGNode unmap(Node node);
+
+
+    /**
+     *Check if the given node is in the image of corresponding
+     * node mapping or not.
```
hartig, line 32:
```suggestion
     * Check if the given RDF term is in the image of this node mapping and,
     * thus, may be one of the RDF terms returned by the {@link #map(LPGNode)}
     * function for some LPG node.
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMappingToBNodesImpl.java
```diff
@@ -0,0 +1,26 @@
...
+    @Override
+    public LPGNode unmap(final Node node) {
+        if (!node.isBlank())
+            throw new IllegalArgumentException("The given RDF term (" + node.toString() + ") is not a blank node.");
+        final String id = node.getBlankNodeId().toString();
+        return new LPGNode(id, "", null);
+    }
+
+    @Override
+    public boolean isPossibleResult(Node node) {
```
hartig, line 23:
```suggestion
    public boolean isPossibleResult(final Node node) {
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMappingToURIsImpl.java
```diff
@@ -0,0 +1,34 @@
...
+        if (!node.isURI())
+            throw new IllegalArgumentException("The given RDF term (" + node.toString() + ") is not an URI node.");
+        if (!node.getURI().startsWith(NSNODE))
+            throw new IllegalArgumentException("The given IRI (" + node.getURI() + ") is not in the image of this node mapping.");
+        final String id = node.getURI().replaceAll(NSNODE, "");
+        return new LPGNode(id, "", null);
+    }
+
+    @Override
+    public boolean isPossibleResult(Node node) {
```
hartig, line 31:
```suggestion
    public boolean isPossibleResult(final Node node) {
```

hartig (changes requested):
There are a few more change requests that you can apply directly from the Github page about this PR.

Additionally, you forgot to change the implementation of the `mapsToNode` method. That method can now simply use the new `isPossibleResult` function of the `NodeMapping` given to `LPG2RDFConfigurationImpl`.

Finally, once you have changed the implementation of the `mapsToNode` method, you can remove `NODE` from `LPG2RDFConfigurationImpl` (line 11).

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/NodeMapping.java
```diff
@@ -28,8 +28,9 @@ public interface NodeMapping {
 
 
     /**
-     *Check if the given node is in the image of corresponding
-     * node mapping or not.
+     *Check if the given RDF term is in the image of this node mapping and,
+     *thus, may be one of the RDF terms returned by the {@link #map(LPGNode)}
+     *function for some LPG node.
```
hartig, line 33:
```suggestion
     * Check if the given RDF term is in the image of this node mapping and,
     * thus, may be one of the RDF terms returned by the {@link #map(LPGNode)}
     * function for some LPG node.
```

hartig (approved):
Thanks!

PR #311, opened 2023-10-03
Add LPG2RDFConfigurationReader

BentolhodaH:
Add LPG2RDFConfigurationReader
Add test for LPG2RDFConfigurationReader

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
@@ -40,7 +40,7 @@ public class LPG2RDFConfigurationReader {
         }
         final Resource lpg2rdfConfig = lpg2rdfConfigs.next();
         if(lpg2rdfConfigs.hasNext()){
-            throw new IllegalArgumentException("More than one instance of LPGtoRDFConfiguration.");
+            throw new IllegalArgumentException("More than one instance of LPGtoRDFConfiguration!");
         }
 
         final String validDataType = lpg2Rdf.getNsPrefixURI("xsd").concat("anyURI");
@@ -64,7 +64,7 @@ public class LPG2RDFConfigurationReader {
         }
         final RDFNode labelPredicateObj = labelPredicateIterator.next().getObject();
         if(labelPredicateIterator.hasNext()){
-            throw new IllegalArgumentException("More than one labelPredicate.");
+            throw new IllegalArgumentException("More than one labelPredicate!");
         }
         if (!labelPredicateObj.asLiteral().getDatatypeURI().equals(validDataType)){
             throw new IllegalArgumentException("labelPredicate is invalid!");
@@ -90,7 +90,7 @@ public class LPG2RDFConfigurationReader {
             }
             final RDFNode prefixOfIRIObj = prefixOfIRIsIterator.next().getObject();
             if(prefixOfIRIsIterator.hasNext()){
-                throw new IllegalArgumentException("More than one prefixOfURI");
+                throw new IllegalArgumentException("More than one prefixOfURI!");
             }
 
             final String prefixOfIRIUrl = prefixOfIRIObj.asLiteral().getString();
@@ -100,7 +100,7 @@ public class LPG2RDFConfigurationReader {
             nodeMapping = new NodeMappingToBNodesImpl();
         }
         else {
-            throw new IllegalArgumentException("Type of Resource in NodeMapping is not defined.");
+            throw new IllegalArgumentException("NodeMapping type (" + nodeMappingResourceType + ") is unexpected!");
         }
         return nodeMapping;
     }
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
@@ -1,0 +1,112 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import org.apache.jena.graph.NodeFactory;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ResourceRequiredException;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.ResIterator;
+import org.apache.jena.rdf.model.Resource;
+import org.apache.jena.riot.RDFDataMgr;
+import org.apache.jena.vocabulary.RDF;
+import se.liu.ida.hefquin.vocabulary.LPG2RDF;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class LPG2RDFConfigurationReader {
+
+    public static LPG2RDFConfigurationReader instance = new LPG2RDFConfigurationReader();
+
+    protected LPG2RDFConfigurationReader(){}
+
+    public static LPG2RDFConfigurationImpl readFromFile( final String filename ) {
+        return instance.parseLPG2RDFConf(filename);
+    }
+
+    public static LPG2RDFConfigurationImpl readFromModel(final Model lpg2Rdf ) {
+        return instance.parseLPG2RDFConf(lpg2Rdf);
+    }
+
+    public LPG2RDFConfigurationImpl parseLPG2RDFConf( final String filename ) {
+        final Model lpg2Rdf = RDFDataMgr.loadModel(filename);
+        return parseLPG2RDFConf(lpg2Rdf);
+    }
+
+    public LPG2RDFConfigurationImpl parseLPG2RDFConf(final Model lpg2Rdf ) {
+        final NodeMapping nodeMapping;
+        final String label;
+
+        final ResIterator lpg2rdfConfigs = lpg2Rdf.listResourcesWithProperty(RDF.type, LPG2RDF.LPGtoRDFConfiguration);
+
+        if (!lpg2rdfConfigs.hasNext()){
+            throw new IllegalArgumentException("LPGtoRDFConfiguration is required!");
+        }
+        else {
+            final Resource lpg2rdfConfig = lpg2rdfConfigs.next();
+
+            verifyLPG2RDFConfiguration(lpg2rdfConfig);
+
+            if(!lpg2rdfConfig.hasProperty(LPG2RDF.labelPredicate)){
+                throw new IllegalArgumentException("labelPredicate is required!");
+            }
+            final RDFNode labelPredicateObj = lpg2rdfConfig.getProperty(LPG2RDF.labelPredicate).getObject();
+            String url;
+            try {
+                url = labelPredicateObj.asResource().getURI();
+            }
+            catch (ResourceRequiredException re){
+                throw new IllegalArgumentException("labelPredicate is invalid!");
+            }
+            label = verifyExpectedURI(url).toString();
+
+            final Resource nodeMappingResource = lpg2rdfConfig.getProperty(LPG2RDF.nodeMapping).getResource();
+            final RDFNode nodeMappingResourceType = lpg2Rdf.getRequiredProperty(nodeMappingResource, RDF.type).getObject();
+
+            if ( nodeMappingResourceType.equals(LPG2RDF.IRIBasedNodeMapping) ) {
+                if(!nodeMappingResource.hasProperty(LPG2RDF.prefixOfIRIs)){
+                    throw new IllegalArgumentException("prefixOfIRIs is required!");
+                }
+                final RDFNode prefixOfIRIObj = nodeMappingResource.getProperty(LPG2RDF.prefixOfIRIs).getObject();
+                final String prefixOfIRIUrl = verifyExpectedURI(prefixOfIRIObj.asResource().getURI()).toString();
+                nodeMapping = new NodeMappingToURIsImpl(prefixOfIRIUrl);
+            }
+            else if ( nodeMappingResourceType.equals(LPG2RDF.BNodeBasedNodeMapping) ) {
+                nodeMapping = new NodeMappingToBNodesImpl();
+            }
+            else {
+                throw new IllegalArgumentException("Type of Resource in NodeMapping is not defined.");
+            }
+        }
+        return new LPG2RDFConfigurationImpl(NodeFactory.createURI(label), nodeMapping);
+    }
+
+
+    /**
+     * Verifies that the given string represents an HTTP URI
+     * or an HTTPS URI and, if so, returns that URI.
+     */
+    protected URI verifyExpectedURI(final String uriString ) {
+        final URI uri;
+        try {
+            uri = new URI(uriString);
+            if (    ! uri.isAbsolute()
+                    || (uri.getScheme().equals("http") && uri.getScheme().equals("https")) ) {
+                throw new IllegalArgumentException( "The following URI is of an unexpected type; it should be an HTTP URI or an HTTPS URI: " + uriString );
+            }
+        }
+        catch ( final URISyntaxException ex ) {
+            throw new IllegalArgumentException("URI parse exception (" + ex.getMessage() + ")");
+        }
+
+        return uri;
+    }
+
+    protected void verifyLPG2RDFConfiguration(final Resource lpg2rdfConf){
+        if(!lpg2rdfConf.hasProperty(LPG2RDF.nodeMapping)){
+            throw new IllegalArgumentException("nodeMapping is required!");
+        }
+        else if(!lpg2rdfConf.hasProperty(LPG2RDF.labelPredicate)){
+            throw new IllegalArgumentException("labelPredicate is required!");
+        }
+    }
+}
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
@@ -81,11 +81,17 @@ public class LPG2RDFConfigurationReader {
     }
 
     public NodeMapping getNodeMapping(final Model lpg2Rdf, final Resource lpg2rdfConfig){
-        if(!lpg2rdfConfig.hasProperty(LPG2RDF.nodeMapping)){
+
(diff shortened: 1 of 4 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,112 @@
...
+
+            verifyLPG2RDFConfiguration(lpg2rdfConfig);
+
+            if(!lpg2rdfConfig.hasProperty(LPG2RDF.labelPredicate)){
+                throw new IllegalArgumentException("labelPredicate is required!");
+            }
+            final RDFNode labelPredicateObj = lpg2rdfConfig.getProperty(LPG2RDF.labelPredicate).getObject();
+            String url;
+            try {
+                url = labelPredicateObj.asResource().getURI();
```
hartig, line 55: Given that the domain of the `labelPredicate` property are literals of datatype `xsd:anyURI`, doing `getResource()` here cannot work.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,112 @@
...
+            }
+            catch (ResourceRequiredException re){
+                throw new IllegalArgumentException("labelPredicate is invalid!");
+            }
+            label = verifyExpectedURI(url).toString();
```
hartig, line 60: Move the declaration of the `label` variable to here. I mean, this line should be:
```suggestion
            final String label = verifyExpectedURI(url).toString();
```

hartig: Also, calling `verifyExpectedURI` here is not necessary. The label predicate URI in an LPG-to-RDF configuration can be an arbitrary URI; it does not need to be an HTTP or HTTPS URI.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,112 @@
...
+
+            final Resource nodeMappingResource = lpg2rdfConfig.getProperty(LPG2RDF.nodeMapping).getResource();
+            final RDFNode nodeMappingResourceType = lpg2Rdf.getRequiredProperty(nodeMappingResource, RDF.type).getObject();
+
+            if ( nodeMappingResourceType.equals(LPG2RDF.IRIBasedNodeMapping) ) {
+                if(!nodeMappingResource.hasProperty(LPG2RDF.prefixOfIRIs)){
+                    throw new IllegalArgumentException("prefixOfIRIs is required!");
+                }
```
hartig, line 68: In addition to checking that the `NodeMapping` instance in the RDF description has a `perfixOfIRIs` property, also check that it does not have more than one such property.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,112 @@
...
+                final RDFNode prefixOfIRIObj = nodeMappingResource.getProperty(LPG2RDF.prefixOfIRIs).getObject();
+                final String prefixOfIRIUrl = verifyExpectedURI(prefixOfIRIObj.asResource().getURI()).toString();
```
hartig, line 70: Similar issues as above for the label predicate URI:
i) the object of triples with predicate `prefixOfIRIs` are literals with datatype `xsd:anyURI` and, thus, `asResource()` cannot work here,
ii) no need to verify that the extracted IRI prefix is an HTTP or and HTTPS URI.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,112 @@
...
+                nodeMapping = new NodeMappingToURIsImpl(prefixOfIRIUrl);
+            }
+            else if ( nodeMappingResourceType.equals(LPG2RDF.BNodeBasedNodeMapping) ) {
+                nodeMapping = new NodeMappingToBNodesImpl();
+            }
+            else {
+                throw new IllegalArgumentException("Type of Resource in NodeMapping is not defined.");
```
hartig, line 77: To make it easier for users, increase the information content of this error message by including--in parenthesis within the message--the URI of `nodeMappingResourceType`).

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,112 @@
...
+                    || (uri.getScheme().equals("http") && uri.getScheme().equals("https")) ) {
+                throw new IllegalArgumentException( "The following URI is of an unexpected type; it should be an HTTP URI or an HTTPS URI: " + uriString );
+            }
+        }
+        catch ( final URISyntaxException ex ) {
+            throw new IllegalArgumentException("URI parse exception (" + ex.getMessage() + ")");
+        }
+
+        return uri;
+    }
```
hartig, line 102: As per my comments above, I don't think a helper function with this functionality is needed in this class.

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReaderTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Test
+    public void LPG2RDFConfigWithIRIBasedNodeMapping() {
+        final String turtle =
+                "PREFIX lr:     <http://www.example.org/se/liu/ida/hefquin/lpg2rdf#>\n"
+              + "PREFIX ex:     <http://example.org/>\n"
+              + "\n"
+              + "ex:LPGtoRDFConfig\n"
+              + "   a  lr:LPGtoRDFConfiguration ;\n"
+              + "   lr:labelPredicate  <http://www.w3.org/2000/01/rdf-schema#label> ;\n"
```
hartig, line 25: That's not right. The object of this triple cannot be the URI but it must be a literal that has the URI as its string value and that has the datatype `xsd:anyURI`. Hence, this line should look as follows (but then you also need to introduce the `xsd:` prefix in the header of this snippet of Turtle).
```suggestion
              + "   lr:labelPredicate  "http://www.w3.org/2000/01/rdf-schema#label"^^xsd:anyURI ;\n"
```

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReaderTest.java
```diff
@@ -0,0 +1,144 @@
...
+              + "   lr:nodeMapping  ex:IRINodeMapping ."
+              + "\n"
+              + "ex:IRINodeMapping\n"
+              + "   a  lr:IRIBasedNodeMapping ;\n"
+              + "   lr:prefixOfIRIs  <https://example.org/node/> .";
```
hartig, line 30: Same issue here, the object of this triple must be a literal:
```suggestion
              + "   lr:prefixOfIRIs  "https://example.org/node/"^^xsd:anyURI .";
```

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReaderTest.java
hartig: Same issue then also below in the other tests.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
BentolhodaH: This comment is not clear to me, and I am not able to get your point, can you provide more information about this comment?

hartig: The if-else block before considers two types of things: instances of the RDF class `IRIBasedNodeMapping` and instances of the RDF class `BNodeBasedNodeMapping`. This exception here is for things that are not in any of these two classes (at least not according to an explicit `rdf:type` triple in the given RDF description). So, in the given RDF description they may have `rdf:type` triple(s) with some other class URIs, which was probably a mistake by the user when he/she wrote the RDF description. Now, to make it easier for the user to understand where and what the mistake in the RDF description is, the error message should also mention the (unexpected/unsupported) class URI of the corresponding `rdf:type` triple in the given RDF description.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,107 @@
...
+        final RDFNode labelPredicateObj = labelPredicateIterator.next().getObject();
+        if(labelPredicateIterator.hasNext()){
+            throw new IllegalArgumentException("More than one labelPredicate!");
+        }
+        if (!labelPredicateObj.asLiteral().getDatatypeURI().equals(validDataType)){
+            throw new IllegalArgumentException("labelPredicate is invalid!");
+        }
+        final String label = labelPredicateObj.asLiteral().getString();
+
+        return NodeFactory.createURI(label);
```
hartig, line 74: Does this work even if the string value of the `labelPredicate` property is not actually a URI? For instance, if it is an arbitrary string such as "Hello World!" ?

I really don't know. So, I am just curious. Maybe you can check this with a unit test. If it works with arbitrary strings, that's not good and you should incorporate a check of the string first (i.e., to check that really represents a URI). To do this check you can use the Java class `URI` (in package `java.net`), which has a static function called `create`. Use this function and see whether it throws an exception.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public NodeMapping getNodeMapping(final Model lpg2Rdf, final Resource lpg2rdfConfig){
+        final NodeMapping nodeMapping;
+        if(!lpg2rdfConfig.hasProperty(LPG2RDF.nodeMapping)){
+            throw new IllegalArgumentException("nodeMapping is required!");
+        }
+
+        final Resource nodeMappingResource = lpg2rdfConfig.getProperty(LPG2RDF.nodeMapping).getResource();
```
hartig, line 83: What if `lpg2rdfConfig` has multiple `nodeMapping` properties?

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,107 @@
...
+        final RDFNode nodeMappingResourceType = lpg2Rdf.getRequiredProperty(nodeMappingResource, RDF.type).getObject();
+
+        if ( nodeMappingResourceType.equals(LPG2RDF.IRIBasedNodeMapping) ) {
+            final StmtIterator prefixOfIRIsIterator = nodeMappingResource.listProperties(LPG2RDF.prefixOfIRIs);
+            if(!prefixOfIRIsIterator.hasNext()){
+                throw new IllegalArgumentException("prefixOfIRIs is required!");
+            }
+            final RDFNode prefixOfIRIObj = prefixOfIRIsIterator.next().getObject();
+            if(prefixOfIRIsIterator.hasNext()){
+                throw new IllegalArgumentException("More than one prefixOfURI!");
```
hartig, line 93:
```suggestion
                throw new IllegalArgumentException("An instance of IRIBasedNodeMapping has more than one prefixOfIRIs property.");
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            final String prefixOfIRIUrl = prefixOfIRIObj.asLiteral().getString();
```
hartig, line 96: First, you need to check that `prefixOfIRIObj` is indeed a literal, and you should also check the datatype URI of this literal.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,107 @@
...
+            nodeMapping = new NodeMappingToURIsImpl(prefixOfIRIUrl);
```
hartig, line 97: You can directly do the following here.
```suggestion
            return new NodeMappingToURIsImpl(prefixOfIRIUrl);
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+        else if ( nodeMappingResourceType.equals(LPG2RDF.BNodeBasedNodeMapping) ) {
+            nodeMapping = new NodeMappingToBNodesImpl();
```
hartig, line 100: Similarly here:
```suggestion
            return new NodeMappingToBNodesImpl();
```

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public NodeMapping getNodeMapping(final Model lpg2Rdf, final Resource lpg2rdfConfig){
+        final NodeMapping nodeMapping;
+        if(!lpg2rdfConfig.hasProperty(LPG2RDF.nodeMapping)){
+            throw new IllegalArgumentException("nodeMapping is required!");
+        }
+
+        final Resource nodeMappingResource = lpg2rdfConfig.getProperty(LPG2RDF.nodeMapping).getResource();
+        final RDFNode nodeMappingResourceType = lpg2Rdf.getRequiredProperty(nodeMappingResource, RDF.type).getObject();
```
hartig, line 84: What if `nodeMappingResource` has multiple rdf:type statements? In particular, it may be possible that some RDF description of an LPG-to-RDF configuration describes a node mapping and says that it is of rdf:type `NodeMapping` (the superclass) and of rdf:type `IRIBasedNodeMapping`, or of rdf:type `NodeMapping` and of rdf:type `BNodeBasedNodeMapping`. In this case it may happen that your call of the `getRequiredProperty` function here ends up picking the rdf:type triple with class `NodeMapping` and then you throw the exception below, but the RDF description is not actually wrong.

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReaderTest.java
```diff
@@ -0,0 +1,147 @@
...
+        final LPGNode node = new LPGNode("0", "", null);
+        final Node resultNode = lpg2RDFConfiguration.mapNode(node);
+        assertNotNull(resultNode);
+        assertTrue(resultNode.isBlank());
+        assertEquals(resultNode.getBlankNodeId().toString(), "0");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void InvalidLabelLPG2RDFConfigWithIRIBasedNodeMapping() {
+        final String turtle =
```
hartig, line 83: Can you please include a comment within this function to briefly describe what exactly is illegal in the given RDF description (i.e., to give other developers an idea of what exactly the test is checking).

hartig: Same for the other unit tests below.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
BentolhodaH: I have checked this. It generated no errors with strings like 'Hello World!'. I made changes based on the URI class. Now, 'Hello World!' is unacceptable, but a string like 'test' would be acceptable. So, if I consider it a URI, the current implementation seems correct. However, if I want to assume 'test' is an invalid URI, I should convert it into a URL and check it by calling `toURL().` Please let me know if the `current` implementation is correct.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/LPG2RDFConfigurationReader.java
hartig: @BentolhodaH I think that this comment is not addressed yet.

BentolhodaH: I forgot to address this comment earlier, but I did consider it. In this case, it chooses the first option. Another solution is to check it similarly to how I check LPG2RDFConfiguration. I meant to ask about which one is in your mind, but I forgot. I am so sorry about that.

hartig: Check that there is only one, similar to how you do it for `labelPredicate`.

hartig:
Looks good now. Thanks!

PR #315, opened 2023-11-02
Add RegexBasedEdgeLabelMapping and its tests

BentolhodaH:
Add RegexBasedEdgeLabelMapping and its tests.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/RegexBasedEdgeLabelMappingToURIsImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/RegexBasedEdgeLabelMappingToURIsImpl.java
@@ -1,0 +1,38 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+
+public class RegexBasedEdgeLabelMappingToURIsImpl implements EdgeLabelMapping {
+
+    protected final String NSRELATIONSHIP;
+    protected final String regex;
+
+
+    public RegexBasedEdgeLabelMappingToURIsImpl(final String regex, final String NSRELATIONSHIP){
+        this.NSRELATIONSHIP = NSRELATIONSHIP ;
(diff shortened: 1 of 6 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/RegexBasedEdgeLabelMappingToURIsImpl.java
```diff
@@ -0,0 +1,38 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+
+public class RegexBasedEdgeLabelMappingToURIsImpl implements EdgeLabelMapping {
```
hartig, line 6: You can make this a subclass that extends `EdgeLabelMappingToURIsImpl`. Then, you don't need to repeat the implementation of the last two methods here, and the `map` method can be implemented to call `map` of the superclass.

src/test/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/RegexBasedEdgeLabelMappingToURIsImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.junit.Assert.*;
+
+public class RegexBasedEdgeLabelMappingToURIsImplTest {
+    protected final String NSRELATIONSHIP = "https://example2.org/test/";
+    protected final String regex = "^\\w+";
+
+    protected final EdgeLabelMapping edgeLabelMapping = new RegexBasedEdgeLabelMappingToURIsImpl(regex, NSRELATIONSHIP);
+
+    @Test
+    public void mapEdgeLabel() {
```
hartig, line 16: Can you please also include a test in which the given edge label does not match the regular expression and, thus, an exception is thrown.

hartig (changes requested):

hartig (approved):
Looks all good now. Thanks!

I will merge this PR.

The next step will then be "SingleEdgeLabelMapping" or whatever we called it.

PR #316, opened 2023-11-06
Add SingleEdgeLabelMapping and its tests

BentolhodaH:
Add SingleEdgeLabelMapping and its tests.

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SingleEdgeLabelMappingToURIsImpl.java
+++ b/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SingleEdgeLabelMappingToURIsImpl.java
@@ -1,0 +1,37 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+
+public class SingleEdgeLabelMappingToURIsImpl implements EdgeLabelMapping {
+
+    protected final String label;
+    protected final String iri;
+
+    public SingleEdgeLabelMappingToURIsImpl(final String label, final String iri){
+        this.label=label;
+        this.iri=iri;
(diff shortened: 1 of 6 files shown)


src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SingleEdgeLabelMappingToURIsImpl.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public SingleEdgeLabelMappingToURIsImpl(final String label, final String iri){
+        this.label=label;
+        this.iri=iri;
+    }
+
+    @Override
+    public Node map(final String label) {
+        if (label.equals(this.label)) {
+            return NodeFactory.createURI(iri);
```
hartig, line 19: Create this Node object already in the constructor. That way, it will not be created every time this function is called.

hartig (changes requested):
Just one small thing.

src/main/java/se/liu/ida/hefquin/engine/wrappers/lpgwrapper/impl/SingleEdgeLabelMappingToURIsImpl.java
```diff
@@ -0,0 +1,39 @@
+package se.liu.ida.hefquin.engine.wrappers.lpgwrapper.impl;
+
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+
+public class SingleEdgeLabelMappingToURIsImpl implements EdgeLabelMapping {
+
+    protected final String label;
+    protected final String iri;
```
hartig, line 9: You don't need this member variable anymore. In the `isPossibleResult` function you can get the IRI from the new `node` member variable instead.

hartig (changes requested):

hartig (approved):
Looks good now. Thanks!

PR #322, opened 2023-12-03
Creating structure of MaterializeRDFViewOfLPG (Dummy Version)

BentolhodaH:


hartig (approved):
I am approving this PR and will merge it. However, there are issues with it that you need to focus on next. In particular, the issues are the following.

I don't think there is a need for the `query` member variable.

It will be more efficient if the program uses (at least) two Cypher queries, one that retrieves all nodes of the LPG and one that retrieves all edges.

For the purpose of this program, the Record2SolutionMappingTranslator cannot be used. Instead, the triples for the resulting RDF graph need to be produced directly from the records of the Cypher query results, in a way as defined in the research paper. Yet, for the next task, focus only on executing the two aforementioned Cypher queries.

PR #323, opened 2023-12-12
Get data from neo4j and translate data

BentolhodaH:
I have created two Cypher queries to fetch data from Neo4j and translate the results. However, results handling has remained. 
Could you please take a look at the implementation? 
I am not quite sure about the way to return the results.
Please guide me on how to handle this. 
Thanks!

### DIFF
--- a/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewOfLPG.java
+++ b/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewOfLPG.java
@@ -27,6 +27,7 @@ import se.liu.ida.hefquin.engine.federation.access.impl.reqproc.Neo4jRequestProc
 import se.liu.ida.hefquin.engine.federation.access.impl.reqproc.Neo4jRequestProcessorImpl;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.Record2SolutionMappingTranslator;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.conf.LPG2RDFConfiguration;
+import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.PropertyMap;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.RecordEntry;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.TableRecord;
 import se.liu.ida.hefquin.engine.wrappers.lpgwrapper.data.impl.*;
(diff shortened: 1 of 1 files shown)


hartig (changes requested):
I think you have over-complicated things.

As the Cypher query for retrieving the data about the nodes in the LPG, it should be possible to simply use the query
```
MATCH (n) RETURN n
```
The reason why I think this should be enough is that I believe our internal code that handles the execution of this query and that produces the `RecordsResponse` object based on the result will produce the `RecordsResponse` object such that it contains `LPGNodeValue` objects for each matching LPG node in the query result, and these `LPGNodeValue` objects should contain a `LPGNode` object with all the relevant data of the LPG node (namely, the ID of the node, the label, and the properties). Notice that, to get the `LPGNodeValue` objects, you will have to cast the `Value` object that you get from each `RecordEntry`. Can you try with your small example Neo4j database to see whether the `LPGNode` objects really contain all the relevant data?

If this works, then the Cypher query for retrieving all the relevant data about the edges in the LPG can be as simple as the following.
```
MATCH (n1)-[e]->(n2) RETURN ID(n1) AS nid1, ID(n2) AS nid2, e
```

You do not need to implement result translation in this PR. Just implement two dummy functions that iterate over the `TableRecord` objects in the `RecordsResponse`  for each of the two queries and print to stdout the content of the `RecordEntry` objects (after casting these values to the correct Java class, which would be either `LPGNodeValue`, `LPGEdgeValue`, or `LiteralValue`).

hartig (changes requested):
> Regarding printing in stdout, I am unsure if my implementation is correct. [...] due to the toString() function behavior, I got just the ID as the printed results. [...] please explain a bit more about how the expected output should be.

Well, within the for loop in the `printNodesOutput` function you can access each of the `LPGNode` objects from each of the `LPGNodeValue` objects and, then, get all of the data of such an LPG node from the `LPGNode` object and print each of these things separately (the ID, the label, each of the properties). Similarly, within the for loop in the `printEdgesOutput` function you can to the same for each of the `LPGEdge` objects.

BentolhodaH:
Hi Olaf,

I tried to implement result printing. I realized that the data is not complete and the node label and relationship for edges are missed. So, I modified the queries a bit to bring these two.

hartig (approved):
Looks okay now. But please post what the print out of the program looks like when you apply it for your small example LPG.

BentolhodaH:
```
3,label,Person
3,name,Charlie Sheen
4,label,Movie
4,title,Wall Street
5,label,Person
5,name,Oliver Stone
0,3,ACTED_IN,4
0,role,Bud Fox
1,5,DIRECTED,4
```

hartig:
Thanks for posting the printout!

Issue #328, opened 2024-06-24
RunQueryWithoutSrcSel lists wrong CLI parameter 


keski:
`RunQueryWithoutSrcSel.java` lists an old CLI paramater when calling `help`:
```java
@Override
protected String getSummary() {
       return getCommandName()+" --query=<query> --considerSPARQLEndpoint=<endpoint URI>";
}
```

should be:
```java
@Override
protected String getSummary() {
       return getCommandName()+" --query=<query> --federationDescription=<federation description>";
}
```

hartig:
Closed by PR #329

PR #329, opened 2024-06-24
resolves issue #328

keski:


hartig (commented):
Thanks!

hartig:
By the way, whenever I merge a PR, I am also deleting the corresponding branch right away. This is to avoid ending up with a lot of branches that we don't know anything about anymore in half a year.

PR #331, opened 2024-06-25
resolves issue ##330

keski:


src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -21,13 +21,14 @@ public void registerWith( final CmdGeneral cmdLine ) {
 	}
 
 	@Override
-	public void processArgs( final CmdArgModule cmdLine ) {
+    public void processArgs(final CmdArgModule cmdLine) {
```
hartig, line 24: Please revert the change to this line. You can do so directly by applying the edit suggestion below.

Two things about this:
1. Please don't change from tab-based indentation to space-based indentation. In particular not if you do it in only one line without changing it anywhere else in the file because that makes the file look like a mess if I view it with a different tab size than you (but also don't do it in general--I prefer tab-based).
2. Please don't remove the spaces inside the parenthesis if there are such spaces everywhere else in the file. That makes it look a bit messy because of the inconsistency across the file. If you start a new file and prefer no such spaces there, then that's fine--as long as you keep it consistent within the file.

```suggestion
	public void processArgs( final CmdArgModule cmdLine ) {
```

src/main/java/se/liu/ida/hefquin/cli/modules/ModFederation.java
```diff
@@ -21,13 +21,14 @@ public void registerWith( final CmdGeneral cmdLine ) {
...
 		if ( cmdLine.contains(fedDescrDecl) ) {
 			final String fedDescrFilename = cmdLine.getValue(fedDescrDecl);
 			cat = FederationDescriptionReader.readFromFile(fedDescrFilename);
+		} else {
```
hartig, line 28: I prefer to have `else` in a new line. I know that this makes the code a little less compact, but I find it easier to read.

```suggestion
		}
		else {
```

hartig (changes requested):
Just some style things.

Issue #332, opened 2024-06-25
add a basic web service


keski:
Currently, HeFEQUIN provides no easy way of running a service that allows queries to be issued from the web. This issue is about creating a basic server implementation that provides REST-based API based on the SPARQL protocol over HTTP. The service should include the following: 1) an endpoint that accepts SPARQL queries over HTTP and returns results back to the client, and 2) the federation query engine (HeFEQUIN) configured when the service starts up (i.e., users should not provide their own federation descriptions).

The following REST API should be supported in the initial version of the service:

<details>
<summary>
<code>POST</code><code><b>/sparql</b></code> (execute a SPARQL query)
</summary>

##### Parameters

> | name      |  type     | data type               | description                                                           |
> |-----------|-----------|-------------------------|-----------------------------------------------------------------------|
> | query     |  required | string                  | The SPARQL query string                                               |
> | format    |  required | string                  | The desired response format (e.g., JSON, XML, CSV)                    |

##### Responses

> | http code     | content-type                      | response                                                            |
> |---------------|-----------------------------------|---------------------------------------------------------------------|
> | 200           | application/sparql-results+json   | The query results in JSON format                                     |
> | 200           | application/sparql-results+xml    | The query results in XML format                                      |
> | 200           | text/csv                          | The query results in CSV format                                      |
> | 200           | text/tsv                          | The query results in TSV format                                      |
> | 415           | application/json                  | {"code":"415", "message":"Unsupported Media Type"}                               |

##### Example cURL

> ```bash
> curl -X POST -H "Content-Type: application/x-www-form-urlencoded" --data "query=SELECT+*+WHERE+%7B+%3Fs+%3Fp+%3Fo+%7D" http://localhost:3331/sparql
> ```

</details>

<details>
 <summary>
<code>GET</code><code><b>/sparql</b></code> (execute a SPARQL query via URL parameters)
</summary>

##### Parameters

> | name      |  type     | data type               | description                                                           |
> |-----------|-----------|-------------------------|-----------------------------------------------------------------------|
> | query     |  required | string                  | The SPARQL query string                                               |
> | format    |  required | string                  | The desired response format (e.g., JSON, XML, CSV)                    |

##### Responses

> | http code     | content-type                      | response                                                            |
> |---------------|-----------------------------------|---------------------------------------------------------------------|
> | 200           | application/sparql-results+json   | The query results in JSON format                                     |
[...]

PR #338, opened 2024-06-27
Basic web service

keski:
Provides a basic server implementation that keeps a `HeFQUINEngine` instance running in the background. The service provides a REST API based on the SPARQL HTTP protocol. Tests are included for POST and GET requests and common media types. Resolves issue #332.


hartig:
The automatic (Github Actions based) check of building this PR using Maven has failed. Looking at the error messages, it seems that the Spring Boot classes are not in the build path.

hartig:
After pulling the branch of this PR to my laptop, I just tried `mvn package` and it runs through without any problems. In fact, it even runs all our unit tests, plus the 15 new tests added as part of this PR.

So, the Github Action that runs the automatic check of the PR on Github doesn't seem to pick up the pom file within this PR. I will take a look.

Also, can you please remind me of how to run the service after the successful build.

hartig:
@keski  I now tried `mvn clean package` on my laptop and it worked as well (i.e., with running the unit tests).

hartig:
@keski Compiling of the automatic build for this PR is still failing. To make sure that it really picks up the POM file in the PR, I changed the HeFQUIN version in that POM file. The build log shows the new version but it fails nonetheless. Can you please take a look at the log and check whether you see something strange.

keski:
@hartig This branch has now been refactored completely. The new web service (servlet) is compiled as a war file during normal build and has been confirmed to be work on Jetty and Tomcat. All tests are passing.

hartig:
Great! I can confirm that it compiles (`mvn package`) on my computer.

Two questions:
1. How do I start the service?
2. Once it is started, I can (currently) access it only via `curl` (or some other direct HTTP interaction) but not in a Web browser, right?

src/main/resources/DefaultEngineConf.ttl
hartig, line 1: Just a heads up: During the summer I have done some work on the RDF-based configuration framework of HeFQUIN such that it is possible now to pass lists (`java.util.List` objects) as arguments to a constructor, where such lists can be described in the RDF-based configuration file with all of their elements. This was needed to be able to list all the heuristics to be considered by the heuristics-based logical optimizer. Based on this new functionality, I have adapted `ExampleEngineConf.ttl` in the root directory. These changes should be carried over into this `DefaultEngineConf.ttl` file. Yet, I propose to do it after this PR has been merged because this PR does not contain the code needed to be able to read the extended configuration vocabulary.

keski:
> 1. How do I start the service?

The war file needs to be deployed in a servlet container, such as Jetty or Tomcat. For example, using docker we can run it using:
```bash
$ mvn clean package
$ docker run --rm \
  -p 8080:8080 \
  -v ./target/HeFQUIN-0.0.4-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war \
  tomcat:latest
```
(This is similar to the approach used in #339)

But providing a way to run it directly seems really useful... I've just pushed an update so that we can now use Jetty as an embedded server (deploying as war still works). Now the embedded server can also be run using:
```bash
$ java -cp target/HeFQUIN-0.0.4-SNAPSHOT.jar se.liu.ida.hefquin.service.HeFQUINServer
```

> Once it is started, I can (currently) access it only via curl (or some other direct HTTP interaction) but not in a Web browser, right?

That is correct. At the start page (e.g., `http://localhost:8080`) there is a description of the API/protocol. It is of course possible to run `GET` queries directly in the browser but we currently don't provide an actual query editor.

hartig:
Thanks for the clarification and for the update to use Jetty as an embedded server!

I confirm that running
```
java -cp target/HeFQUIN-0.0.4-SNAPSHOT.jar se.liu.ida.hefquin.service.HeFQUINServer
```
works and starts up the service.

Just one remaining observation: After starting the service and going to http://localhost:8080 in my browser, instead of seeing the `index.html` page with the API description, I see a page saying
```
{"error": "SPARQL query is missing or empty"}
```
Yet, that's a minor thing. If you have an immediate idea how to fix this issue, please implement it still in this PR. Otherwise, I am fine with merging the PR as is. Let me know.

keski:
Oh, I see. I forgot to add a link the html file in `HeFQUINServer.java`. I've now updated it so that it instead uses the `web.xml` so that it behaves the same way as the deployed war.

src/main/java/se/liu/ida/hefquin/service/HeFQUINServer.java
```diff
@@ -0,0 +1,23 @@
...
+		server.start();
+		server.join();
+	}
+
+	public static Server run( final int port ) {
+		final Server server = new Server( port );
+
+        WebAppContext webAppContext = new WebAppContext();
+		webAppContext.setResourceBase("src/main/webapp");
+        server.setHandler(webAppContext);
```
hartig, line 19: Something is wrong with the indentation here.

hartig:
```suggestion
		final WebAppContext webAppContext = new WebAppContext();
		webAppContext.setResourceBase("src/main/webapp");
		server.setHandler(webAppContext);
```

hartig (approved):
Perfect. Works now.

PR #340, opened 2024-06-28
Re implement logical plan walker

huanyu-li:


src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalOperator.java
```diff
@@ -21,4 +21,6 @@ public interface LogicalOperator
 	 * Returns an identifier of this operator, which should be unique for all the operators within the same plan.
 	 */
 	int getID();
+	
+	String printString(String indentStr);
```
hartig, line 25: This should not be part of the operators themselves. Instead, the decision of how exactly each operator is presented should be part of the plan printer. So, please remove this function here (and for the purpose of this particular PR, the plan printer can still just use the `toString()` function of the operators; a later PR can then change this behavior by doing something nicer looking directly in the plan printer).

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpMultiwayJoin.java
```diff
@@ -47,7 +47,12 @@ public void visit( final LogicalPlanVisitor visitor ) {
 
 	@Override
 	public String toString(){
-		return "> mj (" + getID() + ")";
+		return "mj (" + getID() + ")";
+	}
+	
+	@Override
+	public String printString(String identStr) {
+		return identStr + "mj" + " (" + getID() + ")";
```
hartig, line 55: As per my previous comment, remove this function again from all the classes.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+import se.liu.ida.hefquin.engine.utils.IndentingPrintStream;
+
+public class TextBasedLogicalPlanPrinterImpl2 implements LogicalPlanPrinter
+{
+	// The string represents '|  '.
+	protected final String levelIndentBase = "\u2502  ";
+	// The string represents '├──'.
+	protected final String nonLastChildIndentBase = "\u251C\u2500\u2500";
+	// The string represents '└──'.
+	protected final String lastChildIndentBase = "\u2514\u2500\u2500";
```
hartig, line 18: You can make these `static` as well. No need to allocate extra memory for every instance of this class.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+		//final LogicalPlanVisitor afterVisitor = new MyAfterVisitor(iOut);
+		//LogicalPlanWalker.walk(plan, beforeVisitor, afterVisitor);
+		//iOut.flush();
+		
+		planWalk(plan, 0, 0, 0, iOut);
+		iOut.flush();
+		
+	}
+	
+	public String getIndentLevelString(int planNumber, int planLevel, int numberOfSiblings) {
```
hartig, line 33: All the arguments should be marked as `final`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+				indentLevelString += lastChildIndentBase;
+			}
+			else {
+				indentLevelString = "";
+			}
+		}
+		return indentLevelString;
+	}
+	
+	public void planWalk( final LogicalPlan plan, int planNumber, int planLevel, int numberOfSiblings, IndentingPrintStream out) {
```
hartig, line 52: All the arguments should be marked as `final`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+		//System.out.println(getIndentLevelString2(planLevel, planNumber, numberOfSiblings) + plan.getRootOperator().toString() +" "+planLevel + " " + planNumber + " "+ numberOfSiblings);
+		//System.out.println(getIndentLevelString2(planLevel, planNumber, numberOfSiblings) + plan.getRootOperator().toString() );
+		//System.out.println(plan.getRootOperator().printString(getIndentLevelString2(planLevel, planNumber, numberOfSiblings)));
+		out.append( plan.getRootOperator().printString(getIndentLevelString(planNumber, planLevel, numberOfSiblings)) );
+		out.append(System.lineSeparator());
```
hartig, line 57: Be consistent in terms of how you use spaces within such function calls. Otherwise the code looks messy. For instance, in the first of these two lines, you have a space before the first argument and a space after the last argument, whereas the second line does not have such spaces.

My rule of thumb for these cases: If all the arguments are only program variables (i.e., no function calls themselves and no use of the dot (`.`) operator), then I omit these spaces. As soon as one of the arguments is not just a program variable, I put these spaces (and then always both at the beginning and at the end, as you did in the first of these two lines here). Hence, the first of these two lines satisfies this rule but the second one does not.

Also, try to avoid nesting of function calls to a level of two or more. That makes the code difficult to read, and step-by-step debugging also becomes more difficult to keep track of where the execution currently is.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+				indentLevelString = "";
+			}
+		}
+		return indentLevelString;
+	}
+	
+	public void planWalk( final LogicalPlan plan, int planNumber, int planLevel, int numberOfSiblings, IndentingPrintStream out) {
+		//System.out.println(getIndentLevelString2(planLevel, planNumber, numberOfSiblings) + plan.getRootOperator().toString() +" "+planLevel + " " + planNumber + " "+ numberOfSiblings);
+		//System.out.println(getIndentLevelString2(planLevel, planNumber, numberOfSiblings) + plan.getRootOperator().toString() );
+		//System.out.println(plan.getRootOperator().printString(getIndentLevelString2(planLevel, planNumber, numberOfSiblings)));
```
hartig, line 55: I assume you will clean up these things before the PR is ready to be merged.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+		out.append( plan.getRootOperator().printString(getIndentLevelString(planNumber, planLevel, numberOfSiblings)) );
+		out.append(System.lineSeparator());
+		planLevel += 1;
```
hartig, line 58:
```suggestion
		planLevel++;
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+		for ( int i = 0; i < plan.numberOfSubPlans(); ++i ) {
+			planWalk( plan.getSubPlan(i), i, planLevel, plan.numberOfSubPlans(), out );
+		}
+
```
hartig, line 62:
```suggestion
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,314 @@
...
+		out.append(System.lineSeparator());
+		planLevel += 1;
+		for ( int i = 0; i < plan.numberOfSubPlans(); ++i ) {
+			planWalk( plan.getSubPlan(i), i, planLevel, plan.numberOfSubPlans(), out );
+		}
+		
+	}
+
+
+	protected  class MyBeforeVisitor implements LogicalPlanVisitor {
```
hartig, line 66: This is not used anymore, right? In this case, you can remove it.

hartig (changes requested):

huanyu-li:
The last three commits are for addressing above comments.

src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpRequest.java
```diff
@@ -61,14 +61,11 @@ public void visit( final LogicalPlanVisitor visitor ) {
 	public String toString(){
 		final int codeOfReq = req.toString().hashCode();
 		final int codeOfFm = fm.getInterface().toString().hashCode();
```
hartig, line 63: As these two variables are not used anymore in the rest of this function, you can remove them.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,63 @@
...
+			}
+			else {
+				indentLevelString = "";
+			}
+		}
+		return indentLevelString;
+	}
+	
+	
+	public void planWalk( final LogicalPlan plan, final int planNumber, final int planLevel, final int numberOfSiblings, final IndentingPrintStream out) {
```
hartig, line 47: Please add a Javadoc comment that described at least what the arguments of this method are about.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,63 @@
...
+		String indentLevelString = getIndentLevelString(planNumber, planLevel, numberOfSiblings);
```
hartig, line 48: Please make it a habit to mark local variables as `final` if they are not meant to be reassigned later in the code. Following this practice has two advantages: i) it makes it easier to read the code (because one realizes that this variable is not going to change anymore) and ii) it helps to avoid bugs (because the compiler would now complain if we reassign a variable that is not supposed to be reassigned). Hence, in this particular case, the change is:
```suggestion
		final String indentLevelString = getIndentLevelString(planNumber, planLevel, numberOfSiblings);
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,63 @@
...
+		if (planNumber < numberOfSiblings-1) {
+			indentLevelString += nonLastChildIndentBase ;
+		}
+		else {
+			if (numberOfSiblings > 0) {
+				indentLevelString += lastChildIndentBase;
+			}
+			else {
+				indentLevelString = "";
+			}
```
hartig, line 41: When would that case ever happen? Is that a special case that occurs only for the root operator of the plan to be printed? If so, please add a comment into the else-branch to say so (such that we do not wonder again the next time we look at this piece of code).

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,63 @@
...
+	public String getIndentLevelString(final int planNumber, final int planLevel, final int numberOfSiblings) {
+		String indentLevelString = "";
+		for ( int i = 1; i < planLevel; i++ ) {
+			indentLevelString += levelIndentBase;
+		}
+		if (planNumber < numberOfSiblings-1) {
+			indentLevelString += nonLastChildIndentBase ;
+		}
+		else {
+			if (numberOfSiblings > 0) {
```
hartig, line 36: To make the code easier to read, you can unnest these nested if statements; i.e., use `else if (...)` here.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,63 @@
...
+	// The string represents '|  '.
+	private static String levelIndentBase = "\u2502  ";
+	// The string represents '├──'.
+	private static String nonLastChildIndentBase = "\u251C\u2500\u2500";
+	// The string represents '└──'.
+	private static String lastChildIndentBase = "\u2514\u2500\u2500";
+		
+	@Override
+	public void print( final LogicalPlan plan, final PrintStream out ) {
+		final IndentingPrintStream iOut = new IndentingPrintStream(out);
```
hartig, line 22: Is it still necessary to wrap the given `PrintStream` into an `IndentingPrintStream`? I don't think you are using the functionality of `IndentingPrintStream` anywhere here, are you?

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,62 @@
...
+			indentLevelString += nonLastChildIndentBase ;
+		}
+		else if (numberOfSiblings > 0) {
+			indentLevelString += lastChildIndentBase;
+		}
+		else {
+			// This is only for the root operator.
+			indentLevelString = "";
+		}
+		return indentLevelString;
```
hartig, line 37: Since the case of `numberOfSiblings == 0` is a special case (only for the root operator), I would handle it outside of the regular code path that is used for the subplans. This is to improve readability.
Additionally, I would include an additional check to make sure the special case is indeed invoked only for the root operator.
In summary, I would change the code of this function as follows.
```suggestion
		if ( numberOfSiblings == 0 ) {
			// This is only for the root operator of the overall plan to be printed.
			assert planLevel == 0;
			return "";
		}

		String indentLevelString = "";
		for ( int i = 1; i < planLevel; i++ ) {
			indentLevelString += levelIndentBase;
		}

		if ( planNumber < numberOfSiblings-1 ) {
			return indentLevelString + nonLastChildIndentBase ;
		}
		else {
			return indentLevelString + lastChildIndentBase;
		}
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,62 @@
...
+	 * @param out The print stream that will print a plan.
+	 */
+	public void planWalk( final LogicalPlan plan, final int planNumber, final int planLevel, final int numberOfSiblings, final PrintStream out) {
+		final String indentLevelString = getIndentLevelString(planNumber, planLevel, numberOfSiblings);
+		if (planNumber < numberOfSiblings-1) {
+			out.append( indentLevelString + plan.getRootOperator().toString() );
+		}
+		else {
+			out.append( indentLevelString + plan.getRootOperator().toString() );
+		}
```
hartig, line 55: Why do you have if/else here? The code in both branches is the same.

hartig (changes requested):
Just two more (minor) things.

hartig (approved):
Looks good now!

hartig:
I will merge this PR now. The next task then is to extend the new `TextBasedLogicalPlanPrinterImpl2` class that you introduce in this PR, by adding the functionality to print logical plans as shown in the mockup.
image

For that implementation you should not anymore use the `toString()` functions of the operators.

PR #352, opened 2024-08-22
Re implement physical plan printer

huanyu-li:


huanyu-li:
I see there is already a `BaseForPhysicalOps.java` in the code having `getID()`. Should we just use this one?

In addition, is the following way for printing lop information for a physical operator OK?
```
FILTERBindJoin (3) 
|    - bgpAdd (ID)  
|         - fm (ID) SPARQL endpoint at https://query.wikidata.org/sparql )
|         - pattern (HASHCODE) (bgp ?cc  <http://www.w3.org/2000/01/rdf-schema#label>  ?o .)
└──  req (0)	 
                - fm (HASHCODE) SPARQL endpoint at http://dbpedia.org/sparql)	
                - pattern (HASHCODE) (bgp ?c  <http://www.w3.org/2002/07/owl#sameAs>  ?cc .
```

hartig:
> I see there is already a `BaseForPhysicalOps.java` in the code having `getID()`. Should we just use this one?

Yes. In Malaga I asked you to already add this function to the physical operators as well, anticipating that we would need it in the same way as we needed it for the logical operators.

> In addition, is the following way for printing lop information for a physical operator OK? [...]

Thanks for asking! I would prefer:
```
FILTERBindJoin (3)
|  - lop (ID) bgpAdd
|  - fm (ID) SPARQL endpoint at https://query.wikidata.org/sparql
|  - pattern (HASHCODE) (bgp ?cc  <http://www.w3.org/2000/01/rdf-schema#label>  ?o .)
|
└──  req (0)	 
       - fm (HASHCODE) SPARQL endpoint at http://dbpedia.org/sparql
       - pattern (HASHCODE) (bgp ?c  <http://www.w3.org/2002/07/owl#sameAs>  ?cc .)
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,149 @@
...
+
+public class TextBasedPhysicalPlanPrinterImpl2 implements PhysicalPlanPrinter
+{
+	// The string represents '|   '.
+	private static String levelIndentBase = "\u2502   ";
+	// The string represents '├── '.
+	private static String nonLastChildIndentBase = "\u251C\u2500\u2500 ";
+	// The string represents '└── '.
+	private static String lastChildIndentBase = "\u2514\u2500\u2500 ";
+	private static String spaceBase = "    ";
```
hartig, line 21: Assuming these things are the same as in `TextBasedLogicalPlanPrinterImpl2`, I propose to create a base class for both of the cases and move these common things to that base class (in which case you will need to change `private` to `protected`). The base class may be called `BaseForTextBasedPlanPrinters`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,149 @@
...
+		
+	@Override
+	public void print( final PhysicalPlan plan, final PrintStream out ) {
+		planWalk(plan, 0, 0, 1, out, "");
+		out.flush();
+	}
+
+	public String getIndentLevelString(final int planNumber, final int planLevel, final int numberOfSiblings, final String upperRootOpIndentString) {
```
hartig, line 29: This functions seems to be something that can be moved to a common base class as well.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,149 @@
...
+					return indentLevelString + lastChildIndentBase;
+				}
+			}
+			else {
+				return indentLevelString;
+			}
+		}
+	}
+	
+	public String getIndentLevelStringForDetail(final int planNumber, final int planLevel, final int numberOfSiblings, final String indentLevelString) {
```
hartig, line 72: ... and so does this.

And also the functions that print the detailed information about fed.members / about patterns.

huanyu-li:
When printing logical operators as part in a physical plan, do we just need to print unary logical operators?

hartig:
I just wanted to say 'yes', but then I thought that it may sometimes be convenient to see for which, say, logical union op and particular physical union op has been created. In this sense, my answer now is 'no'. Better print the logical ops for every type of physical op.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,354 @@
...
+	protected void printOperatorInfoForRequest( final PhysicalOpRequest op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		final LogicalOpRequest lop = op.getLogicalOperator();
+		final DataRetrievalRequest req = lop.getRequest();
+		out.append( indentLevelString + "req (" + lop.getID() + ")" );
+		out.append( System.lineSeparator() );
+		printFederationMember( lop.getFederationMember(), indentLevelStringForOpDetail, out );
+		out.append( indentLevelStringForOpDetail + "  - pattern (" + req.hashCode() +  ") (" + req.toString() + ")" );
+		out.append( System.lineSeparator() );
+		out.append( indentLevelStringForOpDetail );
+		out.append( System.lineSeparator() );
```
hartig, line 181: Why these?

And why is there now `printLogicalOpForPhysicalOp` in this function?

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,354 @@
...
+	
+	protected void printOperatorInfoForBinaryUnion ( final PhysicalOpBinaryUnion op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "binaryUnion (" + op.getID() + ") " );
+		printLogicalOpForPhysicalOp( op.getLogicalOperator(), out, indentLevelString, indentLevelStringForOpDetail );
+	}
+	
+	protected void printOperatorInfoForBindJoin ( final PhysicalOpBindJoin op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "bindJoin (" + op.getID() + ") " );
+		printLogicalOpForPhysicalOp( op.getLogicalOperator(), out, indentLevelString, indentLevelStringForOpDetail );
+	}
```
hartig, line 109: As `PhysicalOpBindJoin` is a physical operator for `LogicalOpTPAdd`, `LogicalOpTPOptAdd`, and the likes, it has a fed.member and a query pattern as well. Hence, the `- fm ...` and `- pattern ...` lines should be printed here as well.

Same comment for the other bind-join variations (`PhysicalOpBindJoinWithFILTER`, etc.)

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,354 @@
...
+
+	protected void printOperatorInfoForOpNaiveNestedLoopsJoin ( final PhysicalOpNaiveNestedLoopsJoin op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "naiveNestedLoop (" + op.getID() + ") " );
+		printLogicalOpForPhysicalOp( op.getLogicalOperator(), out, indentLevelString, indentLevelStringForOpDetail );
+	}
+
+	protected void printOperatorInfoForOpParallelMultiLeftJoin ( final PhysicalOpParallelMultiLeftJoin op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "parallelMultiLeftJoin (" + op.getID() + ") " );
+		out.append( System.lineSeparator() );
+		//printLogicalOpForPhysicalOp( op.getLogicalOperator(), out, indentLevelString, indentLevelStringForOpDetail );
```
hartig, line 169: Remove if this is not needed here.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,354 @@
...
+		}
+		else if ( lop instanceof LogicalOpGPOptAdd) {
+			printLogicalOpGPOptAddForPhysicalOp( (LogicalOpGPOptAdd) lop, out, indentLevelStringForOpDetail );
+		}
+		else if ( lop instanceof LogicalOpFilter ) {
+			printLogicalOpFilterForPhysicalOp( (LogicalOpFilter) lop, out, indentLevelStringForOpDetail );
+		}
+		else if ( lop instanceof LogicalOpGlobalToLocal ) {
+			printLogicalOpGlobalToLocalForPhysicalOp( (LogicalOpGlobalToLocal) lop, out, indentLevelStringForOpDetail );
+		}
```
hartig, line 220: All of this stuff with all these additional `printLogicalOp...ForPhysicalOp` functions is too complicated. Why not simply a function that returns the short string name for each of the logical operators (e.g., `"tpAdd"` for `LogicalOpTPAdd`) and then a generic function that puts that short string name together with the rest of the things needed for the `- lop` line (where this rest is only the string `- lop` and the string with the ID of the log.op. in parenthesis).

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,354 @@
...
+		else {
+			throw new IllegalArgumentException( "Unexpected logical operator type: " + lop.getClass().getName() );
+		}
+	} 
+	
+	protected void printLogicalOpTpAddForPhysicalOp (final LogicalOpTPAdd lop, final PrintStream out, final String indentLevelStringForOpDetail) {
+		out.append( indentLevelStringForOpDetail + singleBase + "  - lop (" + lop.getID() + ") tpAdd" );
+		out.append( System.lineSeparator() );
+		printFederationMember( lop.getFederationMember(), indentLevelStringForOpDetail + singleBase, out );
+		printSPARQLGraphPattern( lop.getTP(), indentLevelStringForOpDetail + singleBase, out );
```
hartig, line 242: These things don't belong into the function that is supposed to print the `- lop` line.

hartig (changes requested):
While I didn't manage to look at the whole PR, here are already a first bunch of comments that I would like you to address.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
huanyu-li: These two lines are for giving an extra space line after "-pattern".

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
huanyu-li: In `printLogicalOpForPhysicalOp`, I have a check on if the logical operator is `tpAdd`, `tpOptAdd`, `BGPAdd`, `BGPOptAdd`, `GPAdd`, `GPOptAdd`. Therefore `- fm ...` and `- pattern ...` lines will be printed.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
huanyu-li: I don't understand the meaning here.  I have the following output for example from this piece of code after printing `lop ... tpAdd`.

```
indexNestedLoop (4) 
        │  - lop (7) tpAdd
        │  - fm (1) TPFInterface server at http://fragments.dbpedia.org/2016-04/en
        │  - pattern (-803321067) (?P2  <http://www.global-example#researchTopic>  ?R)
        │
        └── req (2)
              - fm (1) TPFInterface server at http://fragments.dbpedia.org/2016-04/en
              - pattern (2003851450) (?P2  a  <http://www.global-example#Professor>)
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
huanyu-li: I'll update as you suggest.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/BaseForTextBasedPlanPrinters.java
```diff
@@ -0,0 +1,101 @@
...
+	protected void printFederationMember( final FederationMember fm, final String indentLevelStringForOpDetail, final PrintStream out ) {
+		out.append( indentLevelStringForOpDetail + "  - fm (" + fm.getInterface().getID() + ") " + fm.getInterface().toString() );
+		out.append( System.lineSeparator() );
+	}
+	
+	protected void printSPARQLGraphPattern (final SPARQLGraphPattern gp, final String indentLevelStringForOpDetail, final PrintStream out ) {
+		out.append( indentLevelStringForOpDetail + "  - pattern (" + gp.hashCode() +  ") (" + gp.toString() + ")" );
+		out.append( System.lineSeparator() );
+		out.append( indentLevelStringForOpDetail );
+		out.append( System.lineSeparator() );
```
hartig, line 99: Why these two lines? It may not always be the case that the `- pattern` line is the last of the detailed information lines. Whatever function uses this function here is the one that can decide whether an empty line should be printed below the `- pattern` line.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl2.java
```diff
@@ -181,19 +105,19 @@ else if ( rootOp instanceof LogicalOpUnion ) {
...
 		out.append( indentLevelString + "bgpAdd (" + op.getID() + ") " );
 		out.append( System.lineSeparator() );
-		printFederationMember( op.getFederationMember(), indentLevelStringForOpDetail + levelIndentBase, out );
-		printSPARQLGraphPattern( op.getBGP(), indentLevelStringForOpDetail + levelIndentBase, out );
-		out.append( indentLevelStringForOpDetail + levelIndentBase );
-		out.append( System.lineSeparator() );
+		printFederationMember( op.getFederationMember(), indentLevelStringForOpDetail + singleBase, out );
+		printSPARQLGraphPattern( op.getBGP(), indentLevelStringForOpDetail + singleBase, out );
+		//out.append( indentLevelStringForOpDetail + singleBase );
+		//out.append( System.lineSeparator() );
```
hartig, line 111: Not needed anymore? Then, remove.

However, I guess that these two code lines are indeed needed after the fixing the issue that have pointed out in my comment above about the `printSPARQLGraphPattern` function.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,334 @@
...
+			throw new IllegalArgumentException( "Unexpected operator type: " + rootOp.getClass().getName() );
+		}
+		for ( int i = 0; i < plan.numberOfSubPlans(); ++i ) {
+			planWalk( plan.getSubPlan(i), i, planLevel+1, plan.numberOfSubPlans(), out, indentLevelString );
+		}
+	}
+	
+	protected void printOperatorInfoForBinaryUnion ( final PhysicalOpBinaryUnion op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "binaryUnion (" + op.getID() + ") " );
+		printLogicalOperator( nameOfLogicalOp(op.getLogicalOperator()), op.getLogicalOperator().getID(), out, indentLevelStringForOpDetail );
```
hartig, line 105: I would change the signature of `printLogicalOperator` such that you can pass the logical operator and instead of passing both the short name and the ID of it. Then, `printLogicalOperator` takes care of obtaining the short name and the ID. This would reduce the amount of source code here in all of these `printOperatorInfoFor...` functions, which should make the code a bit more readable.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,334 @@
...
+			printSPARQLGraphPattern( ((LogicalOpBGPOptAdd) lop).getBGP(), indentLevelStringForOpDetail + singleBase, out );
+		}
+		else if (lop instanceof LogicalOpGPAdd ) {
+			printFederationMember( ((LogicalOpGPAdd) lop).getFederationMember(), indentLevelStringForOpDetail + singleBase, out );
+			printSPARQLGraphPattern( ((LogicalOpGPAdd) lop).getPattern(), indentLevelStringForOpDetail + singleBase, out );
+		}
+		else if (lop instanceof LogicalOpGPOptAdd ) {
+			printFederationMember( ((LogicalOpGPOptAdd) lop).getFederationMember(), indentLevelStringForOpDetail + singleBase, out );
+			printSPARQLGraphPattern( ((LogicalOpGPOptAdd) lop).getPattern(), indentLevelStringForOpDetail + singleBase, out );
+		}
```
hartig, line 149: Move this code into a separate helper function that you can call from each of the `printOperatorInfoFor...` functions in which you currently have a copy of this code block.

One more thing that I have come across recently that makes writing such code with `instanceof` checks followed by a casting operation much easier and more readable: Since Java 14 it is possible to shorten, say, the last else-if block here as follows.
```java
else if ( lop instanceof LogicalOpGPOptAdd addOp ) {
	printFederationMember( addOp.getFederationMember(), indentLevelStringForOpDetail + singleBase, out );
	printSPARQLGraphPattern( addOp.getPattern(), indentLevelStringForOpDetail + singleBase, out );
}
```
Notice the new variable called `addOp` introduced right after the `instanceof` test. This variable is assigned with the same Java object as `lop`, but automatically casted to `LogicalOpGPOptAdd`. As a consequence, the explicit casting inside the else-if block can be removed and the code is less complex.

Look at https://www.baeldung.com/java-pattern-matching-instanceof if you want to read a bit more about this new Java feature.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,334 @@
...
+	
+	protected void printOperatorInfoForRequest( final PhysicalOpRequest op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		final LogicalOpRequest lop = op.getLogicalOperator();
+		final DataRetrievalRequest req = lop.getRequest();
+		out.append( indentLevelString + "req (" + op.getID() + ")" );
+		out.append( System.lineSeparator() );
+		printFederationMember( lop.getFederationMember(), indentLevelStringForOpDetail, out );
+		out.append( indentLevelStringForOpDetail + "  - pattern (" + req.hashCode() +  ") (" + req.toString() + ")" );
+		out.append( System.lineSeparator() );
+	}
```
hartig, line 274: You should do `printLogicalOperator` in this function as well.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,334 @@
...
+	
+	protected void printOperatorInfoForOpSymmetricHashJoin ( final PhysicalOpSymmetricHashJoin op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "symmetricHashJoin (" + op.getID() + ") " );
+		printLogicalOperator( nameOfLogicalOp(op.getLogicalOperator()), op.getLogicalOperator().getID(), out, indentLevelStringForOpDetail );
+	}
+	
+	protected String nameOfLogicalOp ( final LogicalOperator lop) {
```
hartig, line 281: You can move this function to the base class and, then, use it in the plan printer for logical plans as well. This way we will be guaranteed to have consistency regarding the short names used for the logical operators.

hartig (changes requested):

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/BaseForTextBasedPlanPrinters.java
hartig: > Do you mean, let the function calling printSPARQLGraphPattern decide whether to print an extra line?

Yes, exactly.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
hartig: Huanyu, I have fixed the issue. That is, I have increased the Java version to be used, both within our Maven POM file and our configuration file for the CI workflow. This means that you may have to refresh the project in your IDE. For instance, in Eclipse I had to right-click on the HeFQUIN project in the "Package Explorer" tab and the select "Refresh" in the pop-up menu.
/cc @keski

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,293 @@
...
+		final LogicalOperator lop = op.getLogicalOperator();
+		printLogicalOperatorBase( "  - lop ", lop, out, indentLevelStringForOpDetail + singleBase );
+		out.append( System.lineSeparator() );
+		if ( lop instanceof LogicalOpTPAdd addOp ) {
+			printFederationMember( addOp.getFederationMember(), indentLevelStringForOpDetail + singleBase, out );
+		}
+		else if ( lop instanceof LogicalOpTPOptAdd addOp ) {
+			printFederationMember( addOp.getFederationMember(), indentLevelStringForOpDetail + singleBase, out );
+		}
+		printSPARQLGraphPattern( ((LogicalOpTPAdd) lop).getTP(), indentLevelStringForOpDetail + singleBase, out );
```
hartig, line 115: Why don't you use `printOperatorInfoFmAndPattern` here?

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,293 @@
...
+		out.append( indentLevelStringForOpDetail + singleBase );
+		out.append( System.lineSeparator() );
+	}
+	
+	protected void printOperatorInfoForBindJoinWithFILTER ( final PhysicalOpBindJoinWithFILTER op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "FILTERBindJoin (" + op.getID() + ") " );
+		out.append( System.lineSeparator() );
+		printLogicalOperatorBase( "  - lop ", op.getLogicalOperator(), out, indentLevelStringForOpDetail + singleBase );
+		out.append( System.lineSeparator() );
+	}
```
hartig, line 125: You should use `printOperatorInfoFmAndPattern` here as well.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,293 @@
...
+		out.append( System.lineSeparator() );
+		printOperatorInfoFmAndPatternWithoutOpt( lop, out, indentLevelStringForOpDetail );
+	}
+	
+	protected void printOperatorInfoForOpFILTER ( final PhysicalOpFilter op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "filter (" + op.getID() + ") " );
+		out.append( System.lineSeparator() );
+		final LogicalOperator lop = op.getLogicalOperator();
+		printLogicalOperatorBase( "  - lop ", lop, out, indentLevelStringForOpDetail + singleBase );
+		out.append( " filterExpressions (" + ((LogicalOpFilter) lop).getFilterExpressions().toString() + ")" );
```
hartig, line 150: You can do the cast to `LogicalOpFilter` directly for `lop`. That makes hte code a bit more readable.
```suggestion
		final LogicalOpFilter lop = (LogicalOpFilter) op.getLogicalOperator();
		printLogicalOperatorBase( "  - lop ", lop, out, indentLevelStringForOpDetail + singleBase );
		out.append( " filterExpressions (" + lop.getFilterExpressions().toString() + ")" );
```

hartig: ... same in the cases of `LogicalOpGlobalToLocal` and `LogicalOpLocalToGlobal`.

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,293 @@
...
+		out.append( System.lineSeparator() );
+		printLogicalOperatorBase( "  - lop ", op.getLogicalOperator(), out, indentLevelStringForOpDetail + singleBase );
+		out.append( System.lineSeparator() );
+	}
+	
+	protected void printOperatorInfoForOpIndexNestedLoopsJoin ( final PhysicalOpIndexNestedLoopsJoin op, final PrintStream out, final String indentLevelString, final String indentLevelStringForOpDetail ) {
+		out.append( indentLevelString + "indexNestedLoop (" + op.getID() + ") " );
+		out.append( System.lineSeparator() );
+		final LogicalOperator lop = op.getLogicalOperator();
+		printLogicalOperatorBase( "  - lop ", op.getLogicalOperator(), out, indentLevelStringForOpDetail + singleBase );
```
hartig, line 181:
```suggestion
		final LogicalOperator lop = op.getLogicalOperator();
		printLogicalOperatorBase( "  - lop ", lop, out, indentLevelStringForOpDetail + singleBase );
```

src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedPhysicalPlanPrinterImpl2.java
```diff
@@ -0,0 +1,293 @@
...
+		final LogicalOpRequest lop = op.getLogicalOperator();
+		final DataRetrievalRequest req = lop.getRequest();
+		out.append( indentLevelString + "req (" + op.getID() + ")" );
+		out.append( System.lineSeparator() );
+		printFederationMember( lop.getFederationMember(), indentLevelStringForOpDetail, out );
+		out.append( indentLevelStringForOpDetail + "  - pattern (" + req.hashCode() +  ") (" + req.toString() + ")" );
+		out.append( System.lineSeparator() );
+		printLogicalOperatorBase( "  - lop ", lop, out, indentLevelStringForOpDetail );
+		out.append( System.lineSeparator() );
+	}
```
hartig, line 224: In this one, the order of `- fm`+`- pattern` vs `- lop` is different than in all the other cases. Keep it consistent.

hartig (changes requested):

hartig (approved):
Thanks for implementing these additional changes. I will merge the PR, and this wraps up the work on plan printing for now.

PR #354, opened 2024-08-22
339 provide docker image for hefquin

keski:


example/ExampleEngineConf.ttl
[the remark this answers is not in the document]
hartig: As a related comment, we should try to minimize the versions of this type of file that we have in the repo. Ideally, we have only one of them.

example/example-config.properties
hartig, line 1: I have essentially the same comment for this file. Notice that my issue is primarily with the name of the directory (`example`) which I wouldn't use to put files for creating Docker images.

hartig:
> What about dropping the `example/` directory [...]

We may want to have an `example/` directory to provide a few example queries, but that's a different topic of course. So, yes, for the purpose of this PR, introducing this directory doesn't seem to make much sense.

> [...] and replacing the contents of the docker-compose file with:
>
> ```yml
> ...
> 
> volumes:
> # Use custom config files
> #- <PATH_TO_CONFIG_FILE>:/usr/local/tomcat/webapps/ROOT/config.properties
> #- <PATH_TO_ENGINE_CONFIG_FILE>:/usr/local/tomcat/webapps/ROOT/DefaultEngineConf.ttl
> #- <PATH_TO_FEDERATION_CONFIG_FILE>:/usr/local/tomcat/webapps/ROOT/DefaultFedConf.ttl
>``` 

Is it possible to replace each of these `<PATH_TO_...>` by the respective path to the corresponding file in the `resources` subdirectory? (but still have these lines commented out)
If yes, then I would prefer that. Otherwise, I fine with this proposal.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '3.8'
+
+services:
+  hefquin:
+    build:
+      context: .
+      #args:
+      #  TAG: docker-test
```
hartig, line 8:
```suggestion
      ### If you want to build the Docker image for a specific version
      ### of HeFQUIN, uncomment the following two lines and replace
      ### 'docker-test' by the release tag of that version.
      #args:
      #  TAG: docker-test
```

keski:
Yes, pointing to the resource sub-directory works fine for now. @hartig

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      context: .
+      ### If you want to build the Docker image for a specific version
+      ### of HeFQUIN, uncomment the following two lines and replace
+      ### 'docker-test' by the release tag of that version.
+      #args:
+      #  TAG: docker-test
+    ports:
+      - "8080:8080"
+    volumes:
+     # Use custom config files
```
hartig, line 15:
```suggestion
     ### If you want to use configuration files other than the default ones
     ### or a federation description other than the default one, uncomment
     ### the following three lines and change their respective first parts
     ### to refer to the files that you want to use.
```

hartig:
Robin, assuming you are okay with my new change suggestion and commit it, is the PR ready to be merged then?

PR #369, opened 2024-10-03
add workflow to publish docker image 

keski:
This workflow is triggered when a release is published, and builds and publishes the corresponding docker image in `ghcr.io` with the tag `latest` as well as any tags provided for the release.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,10 +15,19 @@ jobs:
     steps:
     # Checkout the repository
     - name: Checkout repository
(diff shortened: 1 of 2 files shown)


keski:
There was a lot of trial and error to get this working correctly. In the end, it turns out that a workflow in a feature branch does not checkout the main branch by default, which made me question my sanity for a while... Anyways, all the commits should be squashed for readability.

hartig (approved):

hartig:
Thanks! I have squashed the commits during merge.

PR #371, opened 2024-10-18
Maven central repo requirements

keski:
First version of metadata required in pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <artifactId>hefquin</artifactId>
   <version>0.0.4-SNAPSHOT</version>
   <packaging>pom</packaging>
-  
+
   <modules>
     <module>hefquin-base</module>
     <module>hefquin-engine</module>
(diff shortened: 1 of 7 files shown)


keski:
This also updates artifactId to lowercase.

pom.xml
```diff
@@ -16,12 +16,44 @@
...
+      <name>Olaf Hartig</name>
+      <email>olaf.hartig@liu.se</email>
+      <organization>Linköping University</organization>
+      <organizationUrl>https://liu.se/</organizationUrl>
+    </developer>
+  </developers>
+
+  <!-- SCM (Source Code Management) -->
+  <scm>
+    <connection>scm:git:git://github.com/LiUSemWeb/HeFQUIN.git</connection>
```
hartig, line 44: You have `:git:` twice here. Is that a typo?

pom.xml
```diff
@@ -16,12 +16,44 @@
...
+    <developerConnection>scm:git:ssh://github.com:LiUSemWeb/HeFQUIN.git</developerConnection>
+    <url>http://github.com/LiUSemWeb/HeFQUIN/tree/main</url>
```
hartig, line 46: Is there a reason why you didn't simply put `http://github.com/LiUSemWeb/HeFQUIN`? (I'm just curious)

Also, shouldn't it be `https://` instead of `http://`?

pom.xml
```diff
@@ -16,12 +16,44 @@
     <module>hefquin-service</module>
   </modules>
 
+  <!-- Project Meta Data -->
+  <name>HeFQUIN</name>
+  <description>HeFQUIN is a query federation engine for heterogeneous federations of graph data sources, including federations of knowledge graphs.</description>
+  <url>http://github.com/LiUSemWeb/hefquin</url>
```
hartig, line 22:
```suggestion
  <url>https://liusemweb.github.io/HeFQUIN/</url>
```
Or not?

hartig (commented):
Thanks! I have just a few questions.

pom.xml
keski: No reason other than that one of the reference examples I looked at used it. But as far as I can tell, using the shorter version should be fine. Unless we stumble across some motivation for it I think we should go for the shroter one, looks neater. I've pushed the update.

Indeed it should be `https` instead of `http` (part of the same commit as above).

hartig (approved):
Looks all good now.

PR #372, opened 2024-10-24
Add VALUESorFILTER bind join

AdrianaConcha:


hartig (changes requested):
Here is my initial set of changes that I would like you to do in this PR.

Notice that several of my comments include an edit suggestion that you can apply directly from within the GitHub page for the PR. Ideally, instead of applying them all separately, you can apply them as a batch in a single commit (notice the "Add suggestion to batch" button).

After you applied them, do a `git pull` in your branch on your computer before you start addressing the other comments in Eclipse. 

In the end, push your locally-changed branch for this PR to GitHub and request my review again on the GitHub page.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUESorFILTER.java
```diff
@@ -0,0 +1,76 @@
...
+				currentInstance.process(input, sink, execCxt);
+				if (!currentInstance.getExceptionsCaughtDuringExecution().isEmpty()) {
+					// Use FILTER-based bind join instead
+					currentInstance = new ExecOpBindJoinSPARQLwithFILTER(query, fm, useOuterJoinSemantics, collectExceptions);
+					currentInstance.process(input, sink, execCxt);
+				}	
+			} catch ( final ExecOpExecutionException e ) {
+				// Use FILTER-based bind join instead
+				currentInstance = new ExecOpBindJoinSPARQLwithFILTER(query, fm, useOuterJoinSemantics, collectExceptions);
+				currentInstance.process(input, sink, execCxt);
```
hartig, line 55: The code in both of these blocks is identical. Avoid such code duplication, because this equivalence may not be obvious to a reader at a first glance (i.e., without looking and comparing in detail) and may lead to unwanted inconsistencies if someone later changes only one of the two parts.

To fix this issue, I suggest you introduce a boolean-typed variable, say `valuesBasedRequestFailed`, that you set to `false` before the try-catch block and then flip to `true` in the two cases of exceptions, and then add a new if-part after the try-catch block to use the FILTER-based version if `valuesBasedRequestFailed` is `true`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUESorFILTER.java
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+
```
hartig, line 58:
```suggestion
```

hartig (changes requested):
Here are some more comments.

In addition to addressing these, I would like you to make the following changes:

This whole class does not need the functionality of `BaseForExecOpBindJoinSPARQL` as superclass. Instead, it is sufficient to have this class be a subclass of `BaseForExecOpBindJoin<SPARQLGraphPattern, SPARQLEndpoint>`. So, change the `extend` part in line 23. After this change, you can delete the `createExecutableRequestOperator` method below (it is not needed), But instead, you need add a `_concludeExecution` function that overrides the `_concludeExecution` function of the superclass; inside this overridden version of this function, you call the `concludeExecution` function of the `currentInstance` (but only if `currentInstance` is not `null`, of course).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUESorFILTER.java
```diff
@@ -0,0 +1,84 @@
...
+ * continues using the FILTER-based approach for the rest of the requests. If the
+ * first VALUES-based request succeeds, however, then the implementation continues
+ * using the VALUES-based approach for the rest of the requests. 
+ */
+public class ExecOpBindJoinSPARQLwithVALUESorFILTER extends BaseForExecOpBindJoin<SPARQLGraphPattern, SPARQLEndpoint> {
+
+	// will be initialized when processing the first input block of solution mappings
+	protected BaseForExecOpBindJoinSPARQL currentInstance = null;
+	boolean valuesBasedRequestFailed;
+
```
hartig, line 27:
```suggestion
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUESorFILTER.java
```diff
@@ -0,0 +1,84 @@
...
+
+	@Override
+	protected void _process( final IntermediateResultBlock input, final IntermediateResultElementSink sink, final ExecutionContext execCxt )
+			throws ExecOpExecutionException {
+		
+		//If this is the first request
+		if (currentInstance == null) {
+			currentInstance = new ExecOpBindJoinSPARQLwithVALUES(query, fm, collectExceptions);
+			valuesBasedRequestFailed = false;
```
hartig, line 44:
```suggestion
			boolean valuesBasedRequestFailed = false;
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUESorFILTER.java
```diff
@@ -0,0 +1,84 @@
...
+				currentInstance = new ExecOpBindJoinSPARQLwithFILTER(query, fm, currentInstance.useOuterJoinSemantics, collectExceptions);
+				currentInstance.process(input, sink, execCxt);
+			}
+		}
+		else {
+			currentInstance.process(input, sink, execCxt);
+		}
+	}
+
+
```
hartig, line 65:
```suggestion
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithVALUESorFILTER.java
```diff
@@ -0,0 +1,84 @@
...
+            final IntermediateResultElementSink sink,
+            final ExecutionContext execCxt ) {
+    	if (currentInstance!= null) {
+    		 try {
+				currentInstance.concludeExecution(sink, execCxt);
+			} catch (ExecOpExecutionException e) {
+				e.printStackTrace();
+			}
+    	}
+    }
```
hartig, line 83: Printing the stack trace of a potential exception and not actually doing anything about the exception is not really robust code ;-)   In this case, the potential exceptions don't need to be caught but can simply be passed on by this function. Here is how this would look like (plus some code layout changes):
```suggestion
	@Override
	protected void _concludeExecution( final IntermediateResultElementSink sink,
	                                   final ExecutionContext execCxt ) throws ExecOpExecutionException {
		if ( currentInstance != null ) {
			currentInstance.concludeExecution(sink, execCxt);
		}
	}
```

hartig (changes requested):
Just a few minor things remaining; all can be applied directly on the GitHub page.

hartig (approved):
Thanks!!

PR #373, opened 2024-11-06
Integrate VALUES or FILTER bind join operator

AdrianaConcha:


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithVALUESorFILTER.java
```diff
@@ -0,0 +1,91 @@
...
+		super(lop);
+		assert lop.getFederationMember() instanceof SPARQLEndpoint;
+	}
+
+	public PhysicalOpBindJoinWithVALUESorFILTER( final LogicalOpTPOptAdd lop ) {
+		super(lop);
+
+		assert lop.getFederationMember() instanceof SPARQLEndpoint;
+	}
+
```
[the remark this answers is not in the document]
hartig: In fact, can you please add the following change to `ExecOpBindJoinSPARQLwithVALUESorFILTER` into this PR: Remove the `useOuterJoinSemantics` argument from the constructor of that class. The constructor is not doing anything with the argument. So, it shouldn't have been there in the first place.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithVALUESorFILTER.java
```diff
@@ -0,0 +1,91 @@
...
+		else if ( lop instanceof LogicalOpTPOptAdd ) {
+			pt = ( (LogicalOpTPOptAdd) lop ).getTP();
+			fm = ( (LogicalOpTPOptAdd) lop ).getFederationMember();
+			useOuterJoinSemantics = true;
+		}
+		else {
+			throw new IllegalArgumentException("Unsupported type of operator: " + lop.getClass().getName() );
+		}
+
+		return createExecOp(pt, fm, useOuterJoinSemantics, collectExceptions);
```
hartig, line 67: Related to my other comment, this part needs to be changed as well:
```suggestion

		if ( lop instanceof LogicalOpTPAdd ) {
			pt = ( (LogicalOpTPAdd) lop ).getTP();
			fm = ( (LogicalOpTPAdd) lop ).getFederationMember();
		}
		else {
			throw new IllegalArgumentException("Unsupported type of operator: " + lop.getClass().getName() );
		}

		return createExecOp(pt, fm, collectExceptions);
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithVALUESorFILTER.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+	
+	
+	protected UnaryExecutableOp createExecOp( final TriplePattern pattern,
+            final FederationMember fm,
+            final boolean useOuterJoinSemantics,
+            final boolean collectExceptions ) {
+		if ( fm instanceof SPARQLEndpoint )
+			return new ExecOpBindJoinSPARQLwithVALUESorFILTER(pattern, (SPARQLEndpoint) fm, useOuterJoinSemantics, collectExceptions );
```
hartig, line 76: ... and this part as well:
```suggestion
	protected UnaryExecutableOp createExecOp( final TriplePattern pattern,
	                                          final FederationMember fm,
	                                          final boolean collectExceptions ) {
		if ( fm instanceof SPARQLEndpoint )
			return new ExecOpBindJoinSPARQLwithVALUESorFILTER(pattern, (SPARQLEndpoint) fm, collectExceptions );
```

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithVALUESorFILTER.java
```diff
@@ -0,0 +1,89 @@
...
+	public PhysicalOpBindJoinWithVALUESorFILTER( final LogicalOpBGPAdd lop ) {
+		super(lop);
+
+		assert lop.getFederationMember() instanceof SPARQLEndpoint;
+	}
+
+	public PhysicalOpBindJoinWithVALUESorFILTER( final LogicalOpGPAdd lop ) {
+		super(lop);
+
+		assert lop.getFederationMember() instanceof SPARQLEndpoint;
```
hartig, line 43: Please try to be consistent in terms of number of separation lines that you add. Otherwise, the code looks messy.
```suggestion
		super(lop);
		assert lop.getFederationMember() instanceof SPARQLEndpoint;
	}

	public PhysicalOpBindJoinWithVALUESorFILTER( final LogicalOpBGPAdd lop ) {
		super(lop);
		assert lop.getFederationMember() instanceof SPARQLEndpoint;
	}

	public PhysicalOpBindJoinWithVALUESorFILTER( final LogicalOpGPAdd lop ) {
		super(lop);
		assert lop.getFederationMember() instanceof SPARQLEndpoint;
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithVALUESorFILTER.java
```diff
@@ -0,0 +1,89 @@
...
+	public UnaryExecutableOp createExecOp( final boolean collectExceptions,
+	                                       final ExpectedVariables... inputVars ) {
+		final TriplePattern pt;
+		final FederationMember fm;
+
+		if ( lop instanceof LogicalOpTPAdd ) {
+			pt = ( (LogicalOpTPAdd) lop ).getTP();
+			fm = ( (LogicalOpTPAdd) lop ).getFederationMember();
+		}
+		else {
```
hartig, line 62: You need to add cases for the other two types of logical operators for which you have added the two extra constructors. See how this is done in `PhysicalOpBindJoinWithFILTER`.

In this context, you will also need to add two more constructors to `ExecOpBindJoinSPARQLwithVALUESorFILTER`.

hartig (changes requested):

hartig:
This looks good now. The next step is to adapt the code that converts a given logical plan into a physical one.

This conversion is implemented primarily in the class `PhysicalPlanFactory`. This class contains a function called `createPlanWithBindJoinVALUES` and you need to add another such a function called `createPlanWithBindJoinVALUESorFILTER`.

Moreover, while the aforementioned new function is for cases in which the engine wants to explicitly pick the new operator, there is also a default choice of a physical operator for every logical operator. For all tpAdd, bgpAdd, and gpAdd operators with SPARQL endpoints, the default is currently the `PhysicalOpBindJoinWithFILTER`. That should be changed such that the new default is the new operator, for which you have to go into the class called `LogicalToPhysicalOpConverter`. The methods of this class implement the mapping from the logical operators to their respective default physical ones. You need to change the three methods that are for `LogicalOpTPAdd`, `LogicalOpBGPAdd`, and `LogicalOpGPAdd`, respectively. In each of these three methods, you need to change the case for SPARQL endpoints.

hartig (commented):
Looks good for now.

I will add the change that makes the VALUES-based operator support outer-join semantics, and then change your VALUES/FILTER-based operator accordingly. When done, I will get back to you with instructions for the next step.

hartig:
I have implemented the change (see commit https://github.com/LiUSemWeb/HeFQUIN/pull/373/commits/e6aa54fdc604a618cfa24f29134498f6cc7f6bd1).

I will merge this PR now because I pushed some other changes to the main branch yesterday and, due to these changes, merging this PR will result in a compile error. I want to fix this error before you continue.

hartig (approved):

Issue #374, opened 2024-12-19
hefquin-service  query parameter parsing bug


keski:
The `hefquin-service` incorrectly parses incoming the query parameter, causing newlines to be incorrectly trimmed from the query string.

PR #375, opened 2024-12-19
resolves issue #374

keski:


hartig (approved):
Not sure whether it matters in this case, but Java has the [`System.lineSeparator()`](http://docs.oracle.com/javase/8/docs/api/java/lang/System.html#lineSeparator--) method which returns a platform-dependent line separator string.

Yet, if you don't think it matters here, feel free to merge as is.

PR #399, opened 2025-03-05
add maximum bloat factor

keski:


hartig (approved):
Thanks for finding and fixing this bug! Great work!!

PR #401, opened 2025-03-12
400 prevent query engine from crashing on cardinality query exceptions

keski:
## Summary of Changes

This pull request includes the following updates:
- The unchecked exception that may be thrown by `QueryExecution.execSelect()` in `SPARQLRequestProcessorImpl` is now caught and wrapped in a `FederationException` to ensure consistent error handling.
- In `FederationAccessManagerBase1`, `FederationExceptions` raised during calls to `issueCardinalityRequest(...)` are now caught, and `Integer.MAX_VALUE` is returned as a fallback cardinality value.

## Additional Notes
- Since this fix is independent of the specific `FederationAccessManager` implementations, cardinality results continue to be cacheable as usual.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/DataRetrievalResponse.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/DataRetrievalResponse.java
@@ -1,5 +1,6 @@
 package se.liu.ida.hefquin.engine.federation.access;
 
+import java.time.Duration;
 import java.util.Date;
 
 import se.liu.ida.hefquin.engine.federation.FederationMember;
@@ -26,4 +27,41 @@ public interface DataRetrievalResponse
 	 * Returns the time at which the retrieval of this response was completed.
 	 */
 	Date getRetrievalEndTime();
+
+	/**
+	 * Returns the total duration between request start and request end.
+	 */
+	default Duration getRequestDuration() {
+		return Duration.between( getRequestStartTime().toInstant(), getRetrievalEndTime().toInstant() );
+	}
+
+	/**
+	 * Indicates whether the response is based on a fallback (default) value due to
+	 * an error.
+	 */
+	default boolean isFallbackResponse() {
+		return false;
+	};
+
+	/**
+	 * Indicates whether an error occurred during data retrieval.
+	 */
+	default boolean isError() {
+		return getErrorStatusCode() != null;
+	};
+
+	/**
+	 * Returns the HTTP status code if the response resulted in an error, or empty
+	 * otherwise.
+	 */
+	default Integer getErrorStatusCode() {
+		return null;
+	};
+
+	/**
+	 * Returns a short description of the error if available, or empty otherwise.
+	 */
+	default String getErrorDescription() {
+		return null;
+	};
 }
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/FederationAccessManagerBase1.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/FederationAccessManagerBase1.java
@@ -8,6 +8,8 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
+
+import org.apache.http.client.HttpResponseException;
 import org.apache.jena.datatypes.xsd.XSDDatatype;
 import org.apache.jena.graph.Node;
 import org.apache.jena.graph.NodeFactory;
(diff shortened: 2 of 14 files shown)


keski:
Noticed a possible issue, closing PR until resolved.

hartig:
> Noticed a possible issue, closing PR until resolved.

Okay, then I will ignore this PR for the time being. However, just to confirm quickly: What you did in `DataRetrievalResponse` is indeed what I had in mind.

Only, I wouldn't have `isFallbackResponse()`. Instead, if `isError()` returns `true`, then the actual data contained in the `DataRetrievalResponse` object, and accessed via some getter function of each corresponding sub-interface, should be treated as meaningless.

In fact, to force that each code block that consumes such responses does indeed check `isError()` and does something special if `isError()` is `true, we may introduce a new type of exception and have the getter function of the sub-interfaces throw this exception.

Actually, perhaps we can make this even more explicit: Can we make `DataRetrievalResponse` a generic with a type argument? i.e., `DataRetrievalResponse<T>` where `T` represents the type of data obtained via the response; e.g., `T` would be `Iterable<Triple>` for [`TriplesResponse`](https://github.com/LiUSemWeb/HeFQUIN/blob/main/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/TriplesResponse.java) and `Iterable<SolutionMapping>` for [`SolMapsResponse`](https://github.com/LiUSemWeb/HeFQUIN/blob/main/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/SolMapsResponse.java). If this is possible, then we can add
```java
  T getResponseData() throws UnsupportedOperationDueToRetrievalError;
```
directly to `DataRetrievalResponse<T>`. What do you think?

keski:
> Only, I wouldn't have `isFallbackResponse()`. Instead, if `isError()` returns `true`, then the actual data contained in the `DataRetrievalResponse` object, and accessed via some getter function of each corresponding sub-interface, should be treated as meaningless.

Sounds good, removed it.

> In fact, to force that each code block that consumes such responses does indeed check `isError()` and does something special if `isError()` is `true, we may introduce a new type of exception and have the getter function of the sub-interfaces throw this exception.
> 
> Actually, perhaps we can make this even more explicit: Can we make `DataRetrievalResponse` a generic with a type argument? i.e., `DataRetrievalResponse<T>` where `T` represents the type of data obtained via the response; e.g., `T` would be `Iterable<Triple>` for [`TriplesResponse`](https://github.com/LiUSemWeb/HeFQUIN/blob/main/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/TriplesResponse.java) and `Iterable<SolutionMapping>` for [`SolMapsResponse`](https://github.com/LiUSemWeb/HeFQUIN/blob/main/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/SolMapsResponse.java). If this is possible, then we can add
> 
> ```java
>   T getResponseData() throws UnsupportedOperationDueToRetrievalError;
> ```
> 
> directly to `DataRetrievalResponse<T>`. What do you think?

I this this sounds like a very neat solution. I've added the generic type to `DataRetrievalResponse`, and I've successfully updated the related classes and interfaces. But the old method of accessing is still being used in the code blocks that consumes such responses, so there are still a few things left to do before the PR is ready for review.

Until then, I have a couple of questions. The changes mean that most `DataRetrievalResponse` implementations have more than one way of accessing the response data (e.g., `getTriples()` and `getResponseData()`). Do we want to keep both or would it be better to keep only the generic `getResponseData()`? Are there any scenarios where we would like to be able to access the response data even if `getResponseData()` would throw `UnsupportedOperationDueToRetrievalError` (e.g., in case of a partial response or similar)?

hartig:
> Until then, I have a couple of questions. The changes mean that most `DataRetrievalResponse` implementations have more than one way of accessing the response data (e.g., `getTriples()` and `getResponseData()`). Do we want to keep both or would it be better to keep only the generic `getResponseData()`?

Good question. I guess to avoid potential confusions and question (e.g., "is `getTriples()` really doing the same as `getResponseData()` or are there cases in which they may behave differently?") it is better to get rid of all the special ones and keep only the generic `getResponseData()`.

> Are there any scenarios where we would like to be able to access the response data even if `getResponseData()` would throw `UnsupportedOperationDueToRetrievalError` (e.g., in case of a partial response or similar)?

I don't think there is a concrete such case at the moment, but we might have such cases in the future. Let's worry about it when we get there. A way to handle it may be to introduce a class that explicitly represent a partial response and have `getResponseData()` return object of this class; then the consumer of `getResponseData()` can decide what to do with the partial response. But, again, let's not worry about this now.

As for the PR related to these things, I would prefer to first have a PR that only changes `DataRetrievalResponse` to `DataRetrievalResponse<T>`, i.e., without any of the extra things that this PR here was about initially. A follow-up PR can then bring in the actual new functionality (e.g., prevent query engine from crashing on cardinality query exceptions).

keski:
PR created (https://github.com/LiUSemWeb/HeFQUIN/pull/409)

For the follow up PR, creating a new PR referencing this one will probably be neater than reopening this PR. Right?

hartig:
> I've committed all changes to the branch https://github.com/LiUSemWeb/HeFQUIN/tree/400-prevent-query-engine-from-crashing-on-cardinality-query-exceptions, but I'll split it up it to simplify the review. I'll create a new branch and PR with changes only related to DataRetrievalResponse and the directly affected classes.

Thanks!

> For the follow up PR, creating a new PR referencing this one will probably be neater than reopening this PR. Right?

Yes :-)

Issue #405, opened 2025-03-18
Dedicated Endpoints for Query Processing Introspection


keski:
The HeFQUIN service currently supports `GET` and `POST` requests for executing queries. However, the CLI exposes additional functionality related to introspection/execution that could be made available as part of the REST API.

Instead of overloading the endpoint (`/sparql`) with additional parameters, dedicated paths can be provided to expose these additional features, although the service doesn't need to be quite as expressive as the CLI (e.g., when requesting the logical query plan the actual execution could be skipped by default).

## Proposal of additional paths and features

| Purpose                                             | Path                                            | Method |
|-------------------------------------|------------------------------------|--------|
| Get physical query plan                    | `/query-plan/physical`                      | POST, GET   |
| Get logical query plan                      | `/query-plan/logical`                        | POST, GET   |
| Get source assignment plan             | `/query-plan/source-assignment`   | POST, GET   |
| Get query execution statistics          | `/query-metrics/execution`               | POST, GET   |
| Get federation access statistics      | `/query-metrics/federation-access` | POST, GET   |

### `/query-plan/physical`
```json
{
  "physicalPlan": "<serialized physical query plan>"
}
```

### `/query-plan/logical`
```json
{
  "logicalPlan": "<serialized logical query plan>"
}
```

### `/query-plan/source-assignment`
```json
{
  "sourceAssignment": "<serialized source assignment>"
}
```

### `/query-metrics/execution`
```json
{
  "overallQueryProcessingTime": "<overall processing time in ms>",
  "planningTime": "<planning time in ms>",
  "compilationTime": "<compilation time in ms>",
  "executionTime": "<execution time in ms>",
  "queryPlanningStats": {
    "overallQueryPlanningTime": "<overall query planning time in ms>",
    "sourcePlanningTime": "<source planning time in ms>",
     "..."
  },
  "executionStats": { }
}
```

### `/query-metrics/federation-access`
```json
{
  "numberOfSPARQLRequestsIssued": "<int>",
  "numberOfTPFRequestsIssued": "<int>",
  "numberOfBRTPFRequestsIssued": "<int>",
  "..."
}
```

Above is only a draft of the extended REST API, but the basic functionality should be similar to that of the CLI.

@hartig, what do you think?

hartig:
Good proposal!!

Comments on some of the details:

> **Response Formats**
> **`/query-plan/physical`**
> 
> ```json
> {
[...]

PR #406, opened 2025-03-24
405 dedicated endpoints for query processing introspection

keski:
This pull request adds support for query inspection as described in #405.

**Summary of changes**
- A new endpoint (`/query-inspect`) and servlet (`HeFQUINServlet`) has been added that returns inspection related results (similar to what is returned by the CLI). The servlets use a shared engine instance.
- A new `StatsPrinter` (`StatsPrinterJSON`) has been added to create a valid JSON object as output.
- The server API is now described using [OpenAPI](https://www.openapis.org/) and rendered using [redoc](https://github.com/Redocly/redoc), which simplifies maintenance of the API description greatly.
- The servlet tests (both for `HeFQUINServlet` and `HeFQUINServletInspection`) have been (re-)structured to follow a more informative naming scheme.


hartig (approved):
Excellent job! Thanks!

Issue #416, opened 2025-04-11
Known security vulnerabilities in Jetty 11.0.14


keski:
There are reports of a few non-critical vulnerabilities in Jetty 11.0.14: https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server/11.0.14.

This ticket is about updating this dependency in `hefquin-service/pom.xml` to a more recent version.

PR #420, opened 2025-04-17
resolve issue #416

keski:


hartig:
Is the update to the POM file all that you wanted to do here or are you still planning to add something else to this PR?

keski:
Yes, the pom update is the only change required. The old version was referencing a dependency that introduced a vulnerability.

hartig:
Okay, then I will merge this PR right away.

Issue #419, opened 2025-04-16
Extend CardinalityResponseImplWithoutCardinality with access to relevant exception


keski:
CardinalityResponseImplWithoutCardinality should be extended to provide access to the relevant exception (i.e., UnsupportedOperationDueToRetrievalError or IllegalArgumentException) that caused it.

keski:
`TPFResponse.getCardinalityEstimate()` would return `null` and result in a `CardinalityResponseImplWithoutCardinality` only if there is no metadata in the give `TPFResonse`. Do we want to provide this as the cause when creating the `CardinalityResponseImplWithoutCardinality` (e.g. as an exception)?

PR #421, opened 2025-04-17
419 extend cardinalityresponseimplwithoutcardinality with access to relevant exception

keski:
**Note:** This code does not add any exception to `CardinalityResponseImplWithoutCardinality` for the case where `TPFResponse.getCardinalityEstimate()` returns `null`.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/TPFResponse.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/TPFResponse.java
@@ -78,8 +78,8 @@ public interface TPFResponse extends TriplesResponse
 
 	/**
 	 * Returns the cardinality estimate provided as metadata in the given
-	 * TPF response. Throws CardinalityEstimationUnavailableErrorn if there is
-	 * no metadata with a cardinality estimate.
+	 * TPF response. Returns <code>null</code> if there is no metadata
+	 * with a cardinality estimate.
 	 */
-	Integer getCardinalityEstimate() throws CardinalityEstimationUnavailableError;
+	Integer getCardinalityEstimate();
 }
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/TPFResponse.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/TPFResponse.java
@@ -78,8 +78,8 @@ public interface TPFResponse extends TriplesResponse
 
 	/**
 	 * Returns the cardinality estimate provided as metadata in the given
-	 * TPF response. Returns <code>null</code> if there is no metadata
-	 * with a cardinality estimate.
+	 * TPF response. Throws CardinalityEstimationUnavailableErrorn if there is
+	 * no metadata with a cardinality estimate.
 	 */
-	Integer getCardinalityEstimate();
+	Integer getCardinalityEstimate() throws CardinalityEstimationUnavailableError;
 }
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/FederationAccessManagerBase1.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/FederationAccessManagerBase1.java
@@ -49,22 +49,22 @@ public abstract class FederationAccessManagerBase1 implements FederationAccessMa
 
 	protected static final Var countVar = Var.alloc("__hefquinCountVar");
 
-	protected AtomicLong issuedCardRequestsSPARQL  = new AtomicLong(0L);
-	protected AtomicLong issuedCardRequestsTPF     = new AtomicLong(0L);
-	protected AtomicLong issuedCardRequestsBRTPF   = new AtomicLong(0L);
+	protected AtomicLong issuedCardRequestsSPARQL = new AtomicLong( 0L );
+	protected AtomicLong issuedCardRequestsTPF =    new AtomicLong( 0L );
+	protected AtomicLong issuedCardRequestsBRTPF =  new AtomicLong( 0L );
 
 	@Override
-	public CompletableFuture<CardinalityResponse> issueCardinalityRequest(
-			final SPARQLRequest req,
-			final SPARQLEndpoint fm )
-					throws FederationAccessException
+	public CompletableFuture<CardinalityResponse> issueCardinalityRequest( final SPARQLRequest req,
+	                                                                       final SPARQLEndpoint fm )
+		throws FederationAccessException
 	{
 		// The idea of this implementation is to take the graph pattern of the
 		// given request, wrap it in a COUNT(*) query, and send that query as
 		// a request to the given endpoint.
 		final SPARQLGraphPattern pattern = req.getQueryPattern();
-		if ( pattern == null )
+		if ( pattern == null ) {
 			throw new IllegalArgumentException();
+		}
 
 		// set up the query as a SELECT query
 		final Query countQuery = QueryFactory.create();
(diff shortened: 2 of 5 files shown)


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/FederationAccessManagerBase1.java
```diff
@@ -49,66 +49,63 @@ public abstract class FederationAccessManagerBase1 implements FederationAccessMa
...
+	protected AtomicLong issuedCardRequestsBRTPF =  new AtomicLong( 0L );
 
 	@Override
-	public CompletableFuture<CardinalityResponse> issueCardinalityRequest(
-			final SPARQLRequest req,
-			final SPARQLEndpoint fm )
-					throws FederationAccessException
+	public CompletableFuture<CardinalityResponse> issueCardinalityRequest( final SPARQLRequest req,
+	                                                                       final SPARQLEndpoint fm )
+		throws FederationAccessException
```
hartig, line 59: Please revert this change. I have been trying to keep the code laid out in a way that the lines are not longer than 80 characters, which makes it easier for people with small screens (like me on my laptop, where otherwise I would always have to scroll back and forth horizontally to see the whole code).
```suggestion
	public CompletableFuture<CardinalityResponse> issueCardinalityRequest(
			final SPARQLRequest req,
			final SPARQLEndpoint fm )
					throws FederationAccessException
```

hartig: Same in the similar cases below.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/response/CardinalityResponseImplWithoutCardinality.java
```diff
@@ -3,10 +3,48 @@
...
+/**
+ * A subclass of {@link CardinalityResponseImpl} representing a response where the cardinality is not available due to a
+ * possible error.
+ *
+ * This implementation wraps a {@link DataRetrievalResponse} and a {@link DataRetrievalRequest} and captures the
+ * exception that occurred, while assigning {@code Integer.MAX_VALUE} as a placeholder for the unknown cardinality.
+ */
 public class CardinalityResponseImplWithoutCardinality extends CardinalityResponseImpl
 {
+	protected Exception exception;
```
hartig, line 15:
```suggestion
	protected final Exception exception;
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/response/CardinalityResponseImplWithoutCardinality.java
```diff
@@ -3,10 +3,48 @@
...
+	/**
+	 * Constructs a {@code CardinalityResponseImplWithoutCardinality} with the given exception.
+	 *
+	 * @param exception       the exception that caused the absence of cardinality/cardinality estimate
+	 * @param wrappedResponse the original data retrieval response
+	 * @param request         the original data retrieval request
+	 */
+	public CardinalityResponseImplWithoutCardinality( final Exception exception,
+		                                              final DataRetrievalResponse<?> wrappedResponse,
+		                                              final DataRetrievalRequest request ) {
```
hartig, line 37:
```suggestion
	public CardinalityResponseImplWithoutCardinality( final Exception exception,
	                                                  final DataRetrievalResponse<?> wrappedResponse,
	                                                  final DataRetrievalRequest request ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/response/CardinalityResponseImplWithoutCardinality.java
```diff
@@ -3,10 +3,48 @@
...
+ *
+ * This implementation wraps a {@link DataRetrievalResponse} and a {@link DataRetrievalRequest} and captures the
+ * exception that occurred, while assigning {@code Integer.MAX_VALUE} as a placeholder for the unknown cardinality.
+ */
 public class CardinalityResponseImplWithoutCardinality extends CardinalityResponseImpl
 {
+	protected Exception exception;
+
+	/**
+	 * Constructs a {@code CardinalityResponseImplWithoutCardinality} with no specific exception.
```
hartig, line 18: Remove this constructor after implementing [your proposal](https://github.com/LiUSemWeb/HeFQUIN/issues/419#issuecomment-2812048730) for `TPFResponse.getCardinalityEstimate() == null`

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/FederationAccessManagerBase1.java
keski: Ah, I see. That's 80 characters for a single line ignoring indentation? I've had my (soft) limit set at 100 (including indentation) based on existing code. I'll revert the code and use 80 chars as the reference for the future.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/response/CardinalityResponseImplWithoutCardinality.java
keski: Now implemented as suggested in [#419](https://github.com/LiUSemWeb/HeFQUIN/issues/419#issuecomment-2821542793)

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/TPFResponse.java
```diff
@@ -78,8 +78,8 @@ default int getMetadataSize(){
 
 	/**
 	 * Returns the cardinality estimate provided as metadata in the given
-	 * TPF response. Returns <code>null</code> if there is no metadata
-	 * with a cardinality estimate.
+	 * TPF response. Throws CardinalityEstimationUnavailableErrorn if there is
+	 * no metadata with a cardinality estimate.
```
hartig, line 82: I am sorry for the confusion. [My comment](https://github.com/LiUSemWeb/HeFQUIN/issues/419#issuecomment-2812505718) related to this point was not sufficiently detailed. It wasn't my intention to change the behavior of this method. In contrast, the method should indeed return `null` if the TPF response does not contain metadata with a cardinality estimate. The place where the new type of exception (`CardinalityEstimationUnavailableError`) should be created is in `FunctionToObtainCardinalityResponseFromTPFResponse` within `FederationAccessManagerBase1`, at the point when the `getCardinalityEstimate()` function is called and returns `null`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/access/impl/FederationAccessManagerBase1.java
```diff
@@ -194,16 +193,19 @@ else if ( countValueObj instanceof Long value ) {
...
-			if ( cardinality != null ) {
-				final int c = cardinality;
-				return new CardinalityResponseImpl(tpfResp, tpfResp.getRequest(), c);
+			final Integer cardinality;
+			try {
+				cardinality = tpfResp.getCardinalityEstimate();
 			}
-			else {
-				return new CardinalityResponseImplWithoutCardinality( tpfResp, tpfResp.getRequest() );
+			catch ( final CardinalityEstimationUnavailableError e ) {
```
hartig, line 204: Related to my other review comment, this is the point where you should create `CardinalityEstimationUnavailableError` instead of catching it. In other words, revert this part here back to the if-else version that we have currently (before this PR) and, then, create `CardinalityEstimationUnavailableError` within the else-branch.

hartig (changes requested):
Just one mistunderstanding :-/

hartig (approved):
Great, thanks!

Issue #350, opened 2024-08-21
Add online query editor and visualizer


keski:
We should consider adding a query editor and result visualizer to the web service in the style of Fuseki. I'd suggest basing this on [YASGUI](https://github.com/TriplyDB/Yasgui) since it provides it all in a nice package.

keski:
We could also consider whether its possible to load the service URLs from the config files and use these for auto-completion.

hartig:
I totally agree that this would be something very nice to have!!

(But please don't start working on it right away. We have several other things that need to be done and we may have to prioritize.)

keski:
I've added  queryy editor page that uses a YASGUI to leverage the two endpoints of the HeFQUIN service (`/sparql` and `/query-inspect`). I've also added a temporary placeholder logo. (see [#350 ](https://github.com/LiUSemWeb/HeFQUIN/compare/main...350-add-online-query-editor-and-visualizer))

Two screenshots included below:

![Image](https://github.com/user-attachments/assets/df6847cb-fbbe-475a-a160-de16c4667ced)

![Image](https://github.com/user-attachments/assets/f33b93e5-2d16-4fc1-b09d-8669dd919fb6)

hartig:
Great! Do you have a PR for it or not yet?

keski:
I've now created the PR

PR #427, opened 2025-04-28
350 add online query editor and visualizer

keski:
This PR addresses issue #350

It provides a query editor available at `/editor.html` based on YASGUI. The editor can communicate with either the SPARQL query endpoint `/sparql`, or the query inspect endpoint `/query-inspect` by selecting the corresponding endpoint.

This PR also contains some refactoring that let's the servlets access not only the engine but also the context and the model describing the engine configuration description. These extra resources are not used at present but could potentially be leveraged later on to, e.g., make engine configuration information available in the service, provide auto-completion of federation member URIs etc.


hartig (approved):

hartig:
Great! 
The code and the changes look good.
I trust it works as advertised (as I am currently not having the chance to actually try it).

PR #429, opened 2025-05-05
357 remove dbpedia directory from hefquin engine

keski:


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/catalog/FederationDescriptionReader.java
```diff
@@ -208,27 +207,33 @@ else if ( addr.isURIResource() ) {
...
+			return null;
 		}
 
-		return null;
+		final RDFNode pathToMappingFile = fd.getRequiredProperty( fm, FD.vocabularyMappingsFile ).getObject();
+
+		final String path = pathToMappingFile.toString();
+		VocabularyMapping vm = vocabMappingByPath.get( path );
+		if ( vm == null ) {
+			vm = new VocabularyMappingWrappingImpl( path );
```
hartig, line 233: This constructor may throw an exception of type `RiotNotFoundException` (perhaps other ones as well?), but that is not clear from the code here. In other parts of `FederationDescriptionReader` we are throwing `IllegalArgumentException` right? In this sense, for the sake of consistency and transparency, I would prefer this line to be wrapped into a try-catch block where the catch part throws an `IllegalArgumentException` that wraps the exception that was caught.

An even better way would be to remove the constructor of `VocabularyMappingWrappingImpl` that is used here and, instead, do the `RDFDataMgr.loadGraph(path)` directly here, wrapped into a try-catch block as mentioned above.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/federation/catalog/FederationDescriptionReaderTest.java
```diff
@@ -162,4 +165,97 @@ public void tpfFMWithoutRequiredProperty() {
...
+		b.parse( fd );
+
+		final FederationCatalog cat = FederationDescriptionReader.readFromModel( fd );
+		final FederationMember fm = cat.getFederationMemberByURI( "http://dbpedia.org/sparql" );
+		assertNotNull( fm.getVocabularyMapping() );
+	}
+
+	@Ignore("Disabled since it uses live web test")
+	@Test
+	public void vocabularyMappingFileIsLoadedFromRemoteURL() {
```
hartig, line 192: I don't think that we need this test. From the perspective of the `FederationDescriptionReader` code that is tested here, it tests the same code path as the test with the local file. The only difference is in how the internally-used Jena code (`RDFDataMgr.loadGraph(path)`) behaves, but perhaps we can assume that it behaves as advertised.

hartig (changes requested):
Minor points.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/federation/catalog/FederationDescriptionReader.java
keski: Okay, sounds good. Iäll update the code to load the graph here and use the `VocabularyMappingWrappingImpl` constructor that takes in a `Graph` argument instead, and wrap `RDFDataMgr.loadGraph(path)` with the try-except as suggested.

hartig (approved):

PR #431, opened 2025-05-05
WIP: Refactor executable operators to consume single solution mappings instead of blocks thereof

hartig:


hartig:
This PR provides a non-trivial redesign of how the push-based (i.e., multi-threaded) execution model of the engine is implemented. The overall aim is to remove delays caused by batching up output solution mappings into IntermediateResultBlock objects and also to reduce the amount of times the threads communicate with one another. The former is achieved by removing the whole idea of IntermediateResultBlock objects; as soon as some output solution mappings are available now, they can be obtained immediately by the consuming thread. The latter is achieved by adding the option to send and to process lists of solution mappings, rather than individual (only) solution mappings, and facilitating this option in the executable operators.

Issue #423, opened 2025-04-21
Don't push BIND clause into SPARQL request


hartig:
Currently, BIND clauses (LogicalOpBind) that have a SPARQL endpoint request as their subplan are pushed into this request by the logical optimizer. While such a behavior is useful for FILTER clauses, it is not so useful for BIND clauses because it unnecessarily increases the size of the solutions mappings that the SPARQL endpoint send back, in response to the request. Hence, the TODO here is to change the logical optimizer such that BIND clauses are not pushed into SPARQL endpoint requests (unless the order is the opposite, with the BIND clause being the subplan under something like a SPARQL request).

hartig:
In fact, we may even want to turn things around: If there is a SPARQL request that contains a BIND clause and the variable introduced by that BIND clause is not used anywhere else in the request pattern, it may be better to pull the BIND out of the request and place it as a separate operator (LogicalBindOp) on top of the request.

PR #433, opened 2025-05-15
423 dont push bind clause into sparql request

keski:


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
```diff
@@ -179,12 +182,25 @@ else if ( rootOp instanceof LogicalOpFilter filterOp )
 		else if ( rootOp instanceof LogicalOpBind bindOp )
 		{
 			// The BIND clause can be merged into a request operator if that request is for a SPARQL endpoint.
-			final LogicalOperator childOp = rewrittenSubPlans.get(0).getRootOperator();
+			final LogicalOperator childOp = rewrittenSubPlans.get( 0 ).getRootOperator();
```
hartig, line 185: In the future, please try to avoid such types of changes that are not actually related to the purpose of the PR, because that makes it harder for me to see where the actual changes are.

By the way, for code like this, I would not actually add spaces. My rule of thumb is to add spaces only for cases in which the function arguments are something composite, i.e., not a single constant or a single program variable but other function calls or operations. For instance:
```java
get( list.size() );  // with spaces
get( i - 1 );  // with spaces
get(i); // no spaces
```

hartig (changes requested):
It seems you have a misunderstanding here. The variables-related check that you do in this PR is not needed at all. Remember that the flow of solution mappings in the query plans is from the child operators to to their parent operators (i.e., from the leaf nodes towards the root node). In this sense, the case that you are checking for would be an incorrect plan to start with (we cannot assign a variable in a BIND that sits on top of a request that already provides bindings for that variable).

The fix to address issue #423 is to simply remove the code within the else-if-block for `LogicalOpBind`. However, I would not remove that whole else-if-block altogether but, instead, replace its content by a comment explaining why we don't push the BIND into a potential SPARQL request. Such a comment will serve as a reminder in the future, where we might come across this case again and might not remember anymore that it was a deliberate decision to avoid pushing BINDs.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
keski: > In the future, please try to avoid such types of changes that are not actually related to the purpose of the PR, because that makes it harder for me to see where the actual changes are.

Noted

keski: > By the way, for code like this, I would not actually add spaces. My rule of thumb is to add spaces only for cases in which the function arguments are something composite, i.e., not a single constant or a single program variable but other function calls or operations

Okay, I'll keep that in mind.

Issue #413, opened 2025-04-08
Create README.txt for release packages


hartig:
We need to create a README.txt file that we put in the top-level directory of the release packages and that contains similar content as [README.md](https://github.com/LiUSemWeb/HeFQUIN/blob/main/README.md), including pointers to the respective documentation pages.

keski:
@hartig , I've added a first version of a `README.txt` and created a PR as draft. I'm not sure how drafts appear for reviewers though so convert to regular PR if needed to provide feedback.

PR #438, opened 2025-06-02
first version of README.txt for release packages (issue #413)

keski:
This is a first version a of a `README.txt` that can be provided as part of release packages. It uses plain text, provides links to relevant resources, and includes quick start guidelines.

*Note:* This file should probably not be merged into the repo but instead only be included in the releases packages themselves, right? I'm only adding it here so that we discuss and add comments in-place.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -1,0 +1,65 @@
+## What is HeFQUIN?
+
+HeFQUIN is a query‑federation engine for heterogeneous graph‑data sources (e.g., federated knowledge graphs).
+
+Project home and documentation: https://liusemweb.github.io/HeFQUIN/
+
+
+## Key resources
+
+* Feature overview: https://liusemweb.github.io/HeFQUIN/doc/features.html
+* User docs: https://liusemweb.github.io/HeFQUIN/doc/
+* Run using Docker: https://liusemweb.github.io/HeFQUIN/doc/docker.html
+* Run using CLI: https://liusemweb.github.io/HeFQUIN/doc/programs.html
+* Developer docs: https://liusemweb.github.io/HeFQUIN/devdoc/
+* Research publications: https://liusemweb.github.io/HeFQUIN/research/
+* GitHub releases page: https://github.com/LiUSemWeb/HeFQUIN/releases/
+
+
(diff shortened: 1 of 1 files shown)


README.txt
```diff
@@ -0,0 +1,65 @@
+## What is HeFQUIN?
+
+HeFQUIN is a query‑federation engine for heterogeneous graph‑data sources (e.g., federated knowledge graphs).
```
hartig, line 3:
```suggestion
HeFQUIN is a query federation engine for heterogeneous graph data sources (e.g., federated knowledge graphs).
```

hartig:
> This file should probably not be merged into the repo but instead only be included in the releases packages themselves, right?

It would still be good to put it under version control and, thus, maintain it in the repo. We may add a directory to the repo in which we maintain artifacts related to creating releases, including this README.txt and perhaps a shell script with the commands of step 9 of the release logistics (https://github.com/LiUSemWeb/HeFQUIN/wiki/Release-Logistics). In fact, we may also move the release logistics file from the wiki into a `ReleaseLogistics.md` file in that directory (so that we can eventually phase out the use of the wiki completely).

I just don't know what a good name for this new directory would be. Certainly not `releases` or `hefquin-release` or anything like that, because it may make people think that the directory contains the actual releases. Any idea?

keski:
> > This file should probably not be merged into the repo but instead only be included in the releases packages themselves, right?
> 
> It would still be good to put it under version control and, thus, maintain it in the repo. We may add a directory to the repo in which we maintain artifacts related to creating releases, including this README.txt and perhaps a shell script with the commands of step 9 of the release logistics (https://github.com/LiUSemWeb/HeFQUIN/wiki/Release-Logistics). In fact, we may also move the release logistics file from the wiki into a `ReleaseLogistics.md` file in that directory (so that we can eventually phase out the use of the wiki completely).
> 
> I just don't know what a good name for this new directory would be. Certainly not `releases` or `hefquin-release` or anything like that, because it may make people think that the directory contains the actual releases. Any idea?

Maybe `release-assets` could work?

Issue #439, opened 2025-06-02
Provide default implementation of HeFQUINEngineConfigReader.Context


keski:
When instantiating the HeFQUIN engine, the user must first define a `Context`, which involves loading the `FederationCatalog` object, creating `ExecutorService`s for both federation access and plan tasks, and then finally implementing the `Context` interface.

This ticket is about providing a more convenient way of defining the context object that relies primarily on the default implementations that are already already provided as part of `HeFQUINEngineDefaultComponents`.

For example, `HeFQUINEngineConfigReader` could define a `DefaultContext` class defines a constructor with a reference to the federation description file.

@hartig, what do you think?

PR #440, opened 2025-06-04
add helper class to obtain a HeFQUIN engine instance

keski:
This PR provides a helper class (`HeFQUINEngineFactory`) that can be used to create `HeFQUINEngine` instances.

In order to shutdown gracefully it is necessary to not only provide access to the engine instance, but also to the `Context` used to create it, since this is what allows us to access the threaded execution services that need to be closed. Therefore, `HeFQUINEngineFactory#create(...)` returns a wrapper that includes both the engine and the context. For added convenience, this wrapper also implements `AutoCloseable` to easily allow clean-up of the thread pools and other resources.

The following snippet demonstrates its intended usage, while leveraging `AutoCloseable`:

```java
String fedCatFile = "ExampleFederation.ttl";
try ( HeFQUINEngineAndContext e = HeFQUINEngineFactory.create(fedCatFile) ) {
	String queryString = Files.readString( Path.of("ExampleQuery.rq") );
	Query q = QueryFactory.create(queryString);
	e.engine().executeQuery(q, ResultsFormat.FMT_TEXT, System.out);
}
```


hartig:
> Rather than creating a huge number of permutations for create(...) I'd suggest using a builder pattern, which will be much more maintainable.

I agree. Please change the PR accordingly.

> In order to shutdown gracefully it is necessary to not only provide access to the engine instance, but also to the Context used to create it, since this is what allows us to access the threaded execution services that need to be closed. [...] a wrapper [...]

This doesn't feel particularly elegant. A better option is to add a shutdown function to the `HeFQUINEngine` interface and, then, handle the shutdown of the thread pools (executor services) there. What do you think?

keski:
Usage example using the new builder class (all components are optional except the federation catalog):
```java
final HeFQUINEngineBuilder builder = new HeFQUINEngineBuilder()
	.withFederationCatalog("ExampleFederation.ttl")
	.withEngineConfiguration("ExampleEngineConf.ttl")
	.withLogicalPlanPrinter(new TextBasedLogicalPlanPrinterImpl())
	.withPhysicalPlanPrinter(new TextBasedPhysicalPlanPrinterImpl())
	.withSourceAssignmentPrinter(new TextBasedLogicalPlanPrinterImpl())
	.withExperimentRun(false)
	.withSkipExecution(false);

try ( HeFQUINEngineAndContext e = builder.build() ) {
	String queryString = Files.readString( Path.of("ExampleQuery.rq") );
	Query q = QueryFactory.create(queryString);
	e.engine().executeQuery(q, ResultsFormat.FMT_TEXT, System.out);
}
```

keski:
In order to access the execution services that need to be shutdown, I modified the constructor of the `HeFQUINEngineImpl`. I was looking for some other way of accessing the execution services, using one of the context objects that are already being passed around, but I couldn't find any way of accessing both execution services.

The `HeFQUINEngineBuilder` now returns an instance of `HeFQUINEngine`, which can be shutdown using the `shutdown` method. New usage example below:

```java
final HeFQUINEngineBuilder builder = new HeFQUINEngineBuilder()
	.withFederationCatalog("ExampleFederation.ttl")
	.withEngineConfiguration("ExampleEngineConf.ttl")
	.withLogicalPlanPrinter(new TextBasedLogicalPlanPrinterImpl())
	.withPhysicalPlanPrinter(new TextBasedPhysicalPlanPrinterImpl())
	.withSourceAssignmentPrinter(new TextBasedLogicalPlanPrinterImpl())
	.withExperimentRun(false)
	.withSkipExecution(false);

final HeFQUINEngine engine = builder.build();
final String queryString = Files.readString( Path.of("ExampleQuery.rq") );
final Query q = QueryFactory.create(queryString);
engine.executeQuery(q, ResultsFormat.FMT_TEXT, System.out);
engine.shutdown();
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngineBuilder.java
```diff
@@ -0,0 +1,185 @@
...
+	 * parameters configured via this builder.
+	 *
+	 * The federation catalog must be provided via {@code withFederationCatalog},
+	 * default values will be used for all other components unless specified explicitly.
+	 *
+	 * @return an initialized {@link HeFQUINEngine} instance
+	 * @throws IllegalStateException if the federation catalog has not been set
+	 */
+	public HeFQUINEngine build() {
+		assert fedCat != null;
```
hartig, line 153: I believe the `throws`- comment is wrong. I don't think that `assert` throws an `IllegalStateException`.

Please replace the `assert` and do throw an `IllegalStateException`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngineImpl.java
```diff
@@ -30,14 +32,22 @@ public class HeFQUINEngineImpl implements HeFQUINEngine
...
 	protected final FederationAccessManager fedAccessMgr;
 	protected final QueryProcessor qProc;
+	protected final ExecutorService execFed;
+	protected final ExecutorService execPlan;
 
 	public HeFQUINEngineImpl( final FederationAccessManager fedAccessMgr,
-	                          final QueryProcessor qProc ) {
+	                          final QueryProcessor qProc,
+	                          final ExecutorService execFed,
+	                          final ExecutorService execPlan ) {
```
hartig, line 41: It somehow doesn't feel nice that these two exec.services need to be passed explicitly here. I understand why you do it, but I feel that there should be a nicer solution.

For `execFed`, an better option may be to add a `shutdown()` function to the `FederationAccessManager` interface and call that one in the `shutdown()` function below here.

Similarly for `execPlan`: add a `shutdown()` function to `QueryProcessor` and, in the implementation within `QueryProcessorImpl`, get the exec.service from the `QueryProcContext ctxt` to shut it down.

keski: Alright, I'll give that a try!

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngineBuilder.java
hartig: It seems you forgot to address this comment.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngineBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+	 * @return this builder instance for method chaining
+	 */
+	public HeFQUINEngineBuilder setExperimentRun(final boolean isExperimentRun) {
+		this.isExperimentRun = isExperimentRun;
+		return this;
+	}
+
+	/**
+	 * Builds and initializes a {@link HeFQUINEngineAndContext} instance using the
+	 * parameters configured via this builder.
```
hartig, line 145:
```suggestion
	 * Returns a {@link HeFQUINEngine} instance that is created using the
	 * parameters configured via this builder.
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngine.java
```diff
@@ -99,4 +99,9 @@ default Pair<QueryProcStats, List<Exception>> executeQuery( Query query )
 	}
 
 	FederationAccessStats getFederationAccessStats();
+
+	/**
+	 * Shutdown all thread pools associated with this engine.
```
hartig, line 104:
```suggestion
	 * Shutdown the relevant components used by this engine, such as the
	 * federation access manager and the query processor component.
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngineImpl.java
```diff
@@ -123,4 +123,12 @@ protected void executeNonSelectQuery( final QueryExecution qe,
 		}
 	}
 
+	/**
+	 * Shutdown all thread pools associated with this engine.
+	 */
```
hartig, line 128:
```suggestion
```

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngineBuilder.java
keski: Oops, must've misclicked. Will address it.

hartig (approved):

After that, can you please take care of the clean-up part of #439, as described here: https://github.com/LiUSemWeb/HeFQUIN/issues/439#issuecomment-2931351819

PR #441, opened 2025-06-11
Clean up code using HeFQUINEngineBuilder

keski:
This PR introduces the following improvements:

- Refactors code leverage the `HeFQUINEngineBuilder`.
- Require an explicit engine configuration file when creating an engine instance. The builder now loads a default configuration file, instead of relying on hard coded defaults.
- Removes default component instances (e.g., `ec:DefaultPhysicalOptimizer`) from the vocabulary. Configuration files must now instead be fully self-contained and explicitly define all required components.
- Removes `HeFQUINEngineDefaultComponents` since it no longer felt necessary to keep it.
- Introduces tests for the CLI.


hartig (approved):
Perfect!

Issue #344, opened 2024-07-05
Change algorithm of UNION-based bind join


hartig:
Our current UNION-based bind join algorithm (see `PhysicalOpBindJoinWithUNION` and `ExecOpBindJoinSPARQLwithUNION`) creates a UNION pattern with FILTERs added into each UNION part. That's unnecessarily complex for the accessed SPARQL endpoint because the given graph pattern before the FILTER is repeated within each of the UNION parts.

A better idea is to create a UNION pattern where each part of the UNION is a version of the given graph pattern in which the join variables have been substituted by applying one of the solutions of the current input batch. To be able to figure out which of the UNION parts a retrieved solution mapping comes from and, thus, to be able to figure out which of the input solutions the retrieved solution has to be joined with, each UNION part needs to be extended with a BIND clause of the form `BIND( x AS ?cnt )` where `?cnt` is a new variable (needs to be the same in all UNION parts) and `x` is an integer that is different for each UNION part.

hartig:
Actually, instead of adding BIND clauses, one of the variables that remains after substituting (which, thus, is not a join variable) should all be renamed in each of the UNION parts such that it is a different variable in each UNION part (e.g., `?x` is renamed to `?x1`, `?x2`, etc. -- but careful! there shouldn't be another variables with any of these new variable names). Then, depending on which of these new variables is bound in a given solution mapping retrieved from the SPARQL endpoint, it is possible to figure out which of the input solution mappings is the corresponding join partner. This idea is called bound join in the FedX paper and, in comparison to both unions with filters and unions with BIND clauses, it reduces both the size of the request queries and the size of the responses to these queries (i.e., the amount of data that is shipped both ways between HeFQUIN and the SPARQL endpoint).

hartig:
With PR #443 merged, you can now go ahead.

Your new class can be subclass of `BaseForExecOpBindJoinSPARQL`. That base class, in turn, is a subclass of `BaseForExecOpBindJoinWithRequestOps`, and that is where all the generic functionality of the various bind-join implementations is. Then, there are two things that you need to implement in your class:
[...]

PR #446, opened 2025-06-27
344 change algorithm of union based bind join

keski:
Implementation of the bound-join algorithm, which uses variable renaming to minimize the number of requests (see issue #344)

I'm a bit uncertain about how to handle the exception in `createExecutableReqOp`, since it doesn't throw any exception in `BaseForExecOpBindJoinWithRequestOps`. Should it throw a runtime exception?

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
@@ -2,12 +2,10 @@ package se.liu.ida.hefquin.engine.queryplan.executable.impl.ops;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
 
 import org.apache.jena.graph.Node;
@@ -204,7 +202,6 @@ public class ExecOpBindJoinSPARQLwithBoundJoin extends BaseForExecOpBindJoinSPAR
 				                                                   solMapsList.get(i) );
 				final SolutionMapping updatedRequest = new SolutionMappingImpl(renamedAndMerged);
 
-				System.err.println(v.getName().substring( prefix.length() ));
 				// Merge with inputSolutionMappings
 				for ( final SolutionMapping smFromInput : inputSolutionMappings ) {
 					if ( SolutionMappingUtils.compatible(smFromInput, updatedRequest ) ) {


keski:
> Should it throw a runtime exception?

I extended `BaseForExecOpBindJoinWithRequestOps#createExecutableReqOp` so that it too now throws `ExecOpExecutionException`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,308 @@
...
+		final SPARQLRequest request = new SPARQLRequestImpl(pattern);
+		return new ExecOpRequestSPARQL(request, fm, false);
+	}
+
+	protected Element createUnion( final Iterable<Binding> solMaps )
+			throws ExecOpExecutionException
+	{
+		// Populate the ordered list of solution mappings (used for restoring renamed
+		// vars and restoring the join partner vars)
+		solMaps.forEach(solMapsList::add);
```
hartig, line 110: I suggest you clear `solMapsList` first. Just in case.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,308 @@
...
+		// and replacing the renamed variable
+		int i = 0;
+		for ( final Binding solMap : solMaps ) {
+			// Create a new pattern in which we replace the variables
+			// with values based on the incoming solution mapping
+			SPARQLGraphPattern patternWithBindings;
+			try {
+				patternWithBindings = query.applySolMapToGraphPattern( new SolutionMappingImpl(solMap) );
+			} catch ( VariableByBlankNodeSubstitutionException e ) {
+				throw new ExecOpExecutionException(e, this);
```
hartig, line 147: Throw an `IllegalArgumentException` instead. The base class should take care that none of the solution mappings passed to `createExecutableReqOp` has a bnode for any of the join variables. Therefore, at this point here, the `VariableByBlankNodeSubstitutionException` should never occur. If it does, there is a bug somewhere.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,308 @@
...
+			
+			// Create new variable 
+			final Var v = Var.alloc( renamedVar.getVarName() + "_" + i );
+			
+			// Rename the variable in the pattern
+			final Element elt2 = renameVar(patternWithBindings, renamedVar, v);
+			union.addElement(elt2);
+			
+			// Add the solution mapping to the list of solution mappings
+			solMapsList.add(solMap);
```
hartig, line 158: Remove. You are doing that already at the beginning of this function.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,308 @@
...
+			// 2. Rename that variable and merge with the corresponding solMapList entry.
+			// 3. For each compatible input mapping, merge and collect the results.
+			for( final Entry<Var, Integer> entry : renamedVars.entrySet() ){
+				final Var v = entry.getKey();
+				final int i = entry.getValue();
+				final Binding b = smFromRequest.asJenaBinding();
+
+				if ( ! b.contains(v) ) {
+					continue;
+				}
```
hartig, line 197: After dropping `renamedVars`, you need to do this differently.

One option would be to enumerate the possible renamed variables, recreating each of them again using:
```java
final Var v = Var.alloc( renamedVar.getVarName() + "_" + i );
```

But there is a better option which avoids creating new Java objects: You simply iterate over the variables in `b` until you come to one that has a name beginning with `renamedVar.getVarName() + "_"`. The rest of the name of that variable should be an integer which you can use as the `i` here.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,308 @@
...
+				final Binding renamedAndMerged = BindingLib.merge( renameVar(b, v, renamedVar),
+				                                                             solMapsList.get(i) );
+				final SolutionMapping updatedRequest = new SolutionMappingImpl(renamedAndMerged);
+
+				// Merge with inputSolutionMappings
+				for ( final SolutionMapping smFromInput : inputSolutionMappings ) {
+					if ( SolutionMappingUtils.compatible(smFromInput, updatedRequest ) ) {
+						solMapsForOutput.add( SolutionMappingUtils.merge(smFromInput, updatedRequest) );
+					}
+				}
```
hartig, line 209: I don't think that this is necessary in your case. You have done the join already. Therefore, all that remains at this point is the following.
```suggestion
				solMapsForOutput.add(updatedRequest);
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,308 @@
...
+		} else {
+			return new MyIntermediateResultElementSink2(currentBatch);
+		}
+	}
+
+	// ------- helper classes ------
+
+	protected class MyIntermediateResultElementSink2 extends MyIntermediateResultElementSink
+	{
+		protected final Set<SolutionMapping> inputSolutionMappingsWithoutJoinPartners = new HashSet<>();
```
hartig, line 179: What's that for? You don't use it.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,308 @@
...
+					}
+				}
+			}
+		}
+	} // end of helper class MyIntermediateResultElementSink2
+
+
+	protected class MyIntermediateResultElementSinkOuterJoin2 extends MyIntermediateResultElementSink2
+	{
+		protected final Set<SolutionMapping> inputSolutionMappingsWithJoinPartners = new HashSet<>();
```
hartig, line 217: A better option is to have `Set<Integer>` here. Then, within the `_send` method, after extracting the integer from the renamed variable in the given `smFromRequest`, you add that integer to this set. Thereafter, in `flush()`, you iterate over all integers from 0 to `solMapsList.size()-1`; each of these integers that is not in this set is the index of a solution mapping in `solMapsList` that did not have a join partner and, thus, needs to be added to `solMapsForOutput`.

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
```diff
@@ -0,0 +1,302 @@
...
+	public final static int DEFAULT_BATCH_SIZE = BaseForExecOpBindJoinWithRequestOps.DEFAULT_BATCH_SIZE;
+
+	protected final Element pattern;
+
+	// Represents a list of input solution mappings (ordered)
+	protected final List<Binding> solMapsList = new ArrayList<>();
+	// Var used for renaming
+	protected Var renamedVar;
+	// Mapping from a renamed var to an index in solMapsList
+	protected final Map<Var, Integer> renamedVars = new HashMap<>();
```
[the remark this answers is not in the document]
keski: Yes, this is indeed possible. This will require us to match the string representation of each renamed variable and parse the integer part, as you mention further down. I'll update the code accordingly and get rid of the map.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
keski: `BaseForExecOpBindJoinWithRequestOps#_processJoinableInput` only adds the restricted solution mappings to `currentSolMapsForRequest` (i.e., the `renamedAndMerged` only contains the join variables), which is why I had to do the join "again". However, this is probably not what we should be doing. I'll override `_processJoinableInput` to add the full input mappings instead, although this means that we no longer filer based on the restricted bindings.

An alternative design would be to restrict the solution mappings used in the request, as in the original implementation, but maintain a mapping from each request solution to the set of input solution mappings it corresponds to. This approach would introduce an additional data structure to track these associations, but would have the advantage of reducing the size of both the generated query and the response size.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
keski: Removed

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpBindJoinSPARQLwithBoundJoin.java
keski: As in the case of `_send` above, `solMapsList` contains only the restricted solution mappings so we still need to loop over `inputSolutionMappings`.

hartig:
> I think I merged with main instead of rebasing... 

You mean you merged the branch of this PR into main? Or the other way around (which I would have guessed is not even possible)?

> I'll clean up the branch to fix the commit history but I'm not sure how git will handle the comments though. Do you know?

I don't know.

Perhaps it is better to create a new branch, copy over your changes, and create a new PR from that one.

keski:
> > I think I merged with main instead of rebasing...
> 
> You mean you merged the branch of this PR into main? Or the other way around (which I would have guessed is not even possible)?
> 

I mean that when I synchronized this branch with `main`, I accidentally placed the commits from `main` on top of my commits, rather than the other way around. Then `main` changed again and the commit history became a mess.

> > I'll clean up the branch to fix the commit history but I'm not sure how git will handle the comments though. Do you know?
> 
> I don't know.
> 
> Perhaps it is better to create a new branch, copy over your changes, and create a new PR from that one.

Will do.

PR #459, opened 2025-07-25
449 multiple engine configuration files

keski:
This PR updates the code to use a single default engine configuration file across all modules. The shared configuration files now reside in `config/`, which is added to the modules' classpaths.


keski:
A default federation configuration file (`DefaultFedConf.ttl`) has also been added to `config/`. This is currently only being used by the `hefquin-service`.

hartig:
Looks good. I will merge this PR now.

PR #464, opened 2025-08-13
Union based bind join physical op

keski:


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithBoundJoin.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * A physical operator that implements (a batching version of) the bound-join
+ * algorithm that uses UNION clauses with variable renaming (as proposed in
+ * the FedX paper by Schwarte et al. 2011). The variable that is renamed can
+ * be any non-join variable.
+ *
+ * <p>
+ * <b>Semantics:</b> TODO
+ * </p>
```
hartig, line 21: ?

You can simply copy the corresponding part from the other `PhysicalOpBindJoinWith*` classes.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
```diff
@@ -575,7 +584,8 @@ public static List<PhysicalPlan> enumeratePlansWithUnaryOpFromReq( final Physica
 				plans.add( createPlanWithBindJoinFILTER(gpAdd, subplan) );
 				plans.add( createPlanWithBindJoinUNION(gpAdd, subplan) );
 				//plans.add( createPlanWithBindJoinVALUES(gpAdd, subplan) );
-				plans.add( createPlanWithBindJoinVALUESorFILTER(gpAdd, subplan) );
+				//plans.add( createPlanWithBindJoinVALUESorFILTER(gpAdd, subplan) );
```
hartig, line 587: This one should not be commented. (The reason for the previous line being commented is that the bind join with VALUES or FILTER is an actual replacement for the bind join with VALUES only.)
```suggestion
				plans.add( createPlanWithBindJoinVALUESorFILTER(gpAdd, subplan) );
```

hartig (changes requested):
Just two very minor things..

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
keski: Ah, that makes sense!

hartig (approved):

hartig:
Thanks!!

PR #466, opened 2025-08-15
skip cardinality join ordering when there are no joins outside requests

keski:
Resolves https://github.com/LiUSemWeb/HeFQUIN/issues/465


hartig (changes requested):
Something is wrong with the indentation of the code in the new `containsJoinOp` function. Please fix.

Other than that, everything looks good.

PR #472, opened 2025-08-25
resolves #471

### LINKED ISSUE #471: Provide User-Agent Policy with SPARQL request
Wikimedia recently announced that they will [Enforcement of the Wikimedia User-Agent policy](https://lists.wikimedia.org/hyperkitty/list/wikitech-l%40lists.wikimedia.org/thread/APW6FQBGIVLCEN7WZ65D4NVZ6XQIWGCW/?sort=date) (2025-09-30). This means that a [User-Agent](https://en.wikipedia.org/wiki/User-Agent) header is required for all requests. Furthermore, the user-agent must comply with their [policy](https://foundation.wikimedia.org/wiki/Policy%3AWikimedia_Foundation_User-Agent_Policy). The user-agent should uniquely identify the client and include project name/version, and provide a contact page/URL (or email).

The enforcement has been rolled out gradually and now prevents all HeFQUIN requests to, e.g., https://query.wikidata.org/sparql. Any request will return a `403 Forbidden`.

In principle, it should be possible to read version/snapshot from the `pom.xml` (although not trivial), but I'm not sure it will work if HeFQUIN is imported as a library. Instead, I'd suggest to use a fixed string, like `HeFQUIN/0.1 (https://github.com/LiUSemWeb/HeFQUIN)`. The method that needs to be modified is `SPARQLRequestProcessorImpl#performRequestWithQueryExecutionHTTP`.

@hartig

keski:


keski:
> > Three comments:
> > 
> > 1. I think it would be better to include the URL of the HeFQUIN Website instead of the GitHub page.
> 
> Sounds good, I'll update it.
>

Done. New structure looks like: `HeFQUIN/0.0.7-SNAPSHOT (https://liusemweb.github.io/HeFQUIN/)`

> > 2. Ideally, we do this not only for the requests to SPARQL endpoints but for all requests.
> 
> Okay, I'll extend for the other request types as well.
> 

I've added the `User-Agent` to TFP and brTPF requests. Does it also need to be added to GraphQL requests?

> > 3. It would indeed be nice to include the actual version number. Can you check how Jena does it? Their CLI tools provide a `--version` argument.
> 
> I'll look into it!

Jena 4.10.0 seems to provide the version as a variables in helper files, while later versions use a different method. However, the easiest way seems to be to use a built-in feature in maven. By introducing a `version.properties` file under `hefquin-access/main/resources` and apply a trick called *filtering* in the module's `pom.xml` file, we can fill in the version number in `version.properties` at compile time. This works both when running in an IDE and when running as jar.

keski:
Since we need to access the user-agent header from modules other than `hefquin-access`, I created a new class (`BuildInfo.java`) in `hefquin-base`, which allows all connections to access the user-agent string in the same way.

hartig:
> Since we need to access the user-agent header from modules other than `hefquin-access`, I created a new class (`BuildInfo.java`) in `hefquin-base`, which allows all connections to access the user-agent string in the same way.

Makes sense!

hartig (approved):
Looks good now. Thanks again for the initiative.

Issue #210, opened 2022-07-09
add toString() to all executable operators


hartig:
... to make debugging output and mentioning such operators in exceptions look more informative.

diptechexpo:
I am interested to contribute in this project. However I am not seeing any setup guideline of the project in the documentation. If there are setup instructions available, can that be shared?
I am new to open source and looking for a good-first-issue to begin.

hartig:
@diptechexpo thanks a lot for your offer to contribute. Much appreciated!! Yet, the timing is a bit unfortunate, as we are all on vacation at the moment (we have relatively long summer breaks in Sweden). Additionally, we have just started putting an effort in documenting the project and making it easier to use the engine (and, eventually, also to contribute to its development).

Having said that, at least there is already some information for potential contributors in the wiki of the repo, including the following page with some "getting started" instructions.

https://github.com/LiUSemWeb/HeFQUIN/wiki/Getting-started-as-a-contributor

If there is anything unclear in these instructions or you think something important is missing, please create a separate issue about it.

Other than that, feel free to create a PR that addresses this issue (or ask if you have any more questions about it).

PR #473, opened 2025-08-25
Add executable plan printer

keski:
This PR provides a first implementation of a text based printer for executable plans (see https://github.com/LiUSemWeb/HeFQUIN/issues/27).

The printer currently supports only `PushBasedExecutablePlanImpl`. Also note that the `toString()` methods have not yet been implemented for the executable operators (see https://github.com/LiUSemWeb/HeFQUIN/issues/210), which means the printer defaults to printing class reference + hash code.

@hartig, I was also experimenting with implementing the visitor pattern for executable plans. This would provide more control over how the operators are actually printed, but the executable operators aren't available from the tasks in the current code. Thoughts?


hartig:
Can you please paste an example of a plan printed using the provided functionality.

Generally, I don't think we don't need to make an effort for the printing of executable plans. These plans are very much like the physical ones. Also, they are actually more interesting to see after (or during) execution, for which we have the StatsPrinter.

hartig (approved):

PR #475, opened 2025-09-09
propagate pre-exec exceptions when stats is null

### LINKED ISSUE #474: Propagate pre-execution exceptions when stats have not been instantiated
When an exception occurs after query parsing but _before_ execution begins, it isn’t propagated correctly to the user. This happens because the stats object (`QueryProcessingStatsAndExceptions`) may not have been created yet, and the error path assumes it exists. As a result, the original exception is swallowed or replaced with a more generic error.

For example, the query below executes a request to a missing federation member:
```sparql
PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX owl:    <http://www.w3.org/2002/07/owl#>
PREFIX dbo:    <http://dbpedia.org/ontology/>

SELECT *
WHERE {
	SERVICE <http://missing_federation_member> {
		<http://dbpedia.org/resource/Berlin> dbo:country ?c .
		?c owl:sameAs ?cc
	}
}
```

When executed, this currently reports:
```bash
Cannot invoke "se.liu.ida.hefquin.engine.QueryProcessingStatsAndExceptions.getOverallQueryProcessingTime()" because "other" is null
java.lang.NullPointerException: Cannot invoke "se.liu.ida.hefquin.engine.QueryProcessingStatsAndExceptions.getOverallQueryProcessingTime()" because "other" is null
        at se.liu.ida.hefquin.engine.queryproc.impl.QueryProcessingStatsAndExceptionsImpl.<init>(QueryProcessingStatsAndExceptionsImpl.java:63)
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execAndPrint(HeFQUINEngine.java:269)
        at se.liu.ida.hefquin.engine.HeFQUINEngine.executeQueryAndPrintResult(HeFQUINEngine.java:171)
        at se.liu.ida.hefquin.cli.RunQueryWithoutSrcSel.exec(RunQueryWithoutSrcSel.java:139)
        at org.apache.jena.cmd.CmdMain.mainMethod(CmdMain.java:87)
        at org.apache.jena.cmd.CmdMain.mainRun(CmdMain.java:56)
        at se.liu.ida.hefquin.cli.TestOp.main(TestOp.java:40)
```

While the underlying issue:
```bash
java.util.NoSuchElementException: no federation member with URI <http://missing_federation_member>
```
is lost.

keski:
Resolves #474

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngine.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/HeFQUINEngine.java
@@ -1,6 +1,7 @@
 package se.liu.ida.hefquin.engine;
 
 import java.io.PrintStream;
+import java.util.Arrays;
 
 import org.apache.jena.query.Query;
 import org.apache.jena.query.QueryExecution;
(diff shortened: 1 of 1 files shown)


hartig (approved):
Good catch!!!

Did you want to add anything more into this PR or can I merge it?

keski:
> Good catch!!!
> 
> Did you want to add anything more into this PR or can I merge it?

I have nothing more to add so this can be merged.

PR #477, opened 2025-09-12
476 improve robustness when creating physical plans from logical plans

keski:
This PR provides a new way of converting `LogicalOperator`s to `PhysicalOperator`s. The idea is that every `PhysicalOperator` has a `Provider`, which exposes a method `Provider#supported(...)` that is used to verify that a conversion is supported, and a method `Provider#create(...)` that actually produces the instantiates the physical operator.

`Provider`s are registered in a `PhysicalOpFactory`. To convert a `LogicalOperator` we use the factory (`PhsyicalOpFactory#create`), which goes through the registered providers in order and returns the first physical op that supports the input.

This basically means that (once complete) the `PhysicalOpFcatory` replaces most (if not all) of the functionality in `LogicalToPhysicalOpConverter`. One change is that from the `PhysicalOpFactory` side we need to make sure to pass the expected variables to the factory, since this is required to check if e.g. we can find a non-joining var for `PhysicalOpBindJoinWithBoundJoin`.

Right now I've only added providers to `PhysicalOpBindJoinWithBoundJoin` and `PhysicalOpBindJoinWithUnion` and hooked in the factory call inside the `LogicalToPhysicalOpConverter`. This means demonstrates that we use a fallback when there is no variable for renaming. For example:
```sparql
PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#>
PREFIX owl:    <http://www.w3.org/2002/07/owl#>
PREFIX dbo:    <http://dbpedia.org/ontology/>

SELECT * WHERE {
	SERVICE <http://dbpedia.org/sparql> {
		<http://dbpedia.org/resource/Berlin> dbo:country ?c .
		?c owl:sameAs ?cc
	}
	SERVICE <https://query.wikidata.org/sparql> {
		?cc rdfs:label ?o
	}
}
```
 still gives the physical plan:
```
--------- Physical Plan ---------
bound join-based bind join (7) 
│  - lop: gpAdd (6)
│  - fm (3) SPARQL endpoint at https://query.wikidata.org/sparql
│  - pattern (-1989281089): { ?cc rdfs:label ?o }
│  - certain variables: ?cc, ?o, ?c
│  - possible variables: none
│  - query planning info: none
│
└── req (4)
      - lop: req (1)
      - fm (0) SPARQL endpoint at http://dbpedia.org/sparql
      - pattern (-136179866): { ?c owl:sameAs ?cc . <http://dbpedia.or[...]ttp://dbpedia.org/ontology/country> ?c }
      - certain variables: ?cc, ?c
      - possible variables: none
      - query planning info:  cardinality = 338 (accurate)
                              max cardinality = 338 (accurate)
                              min cardinality = 338 (accurate)
    
--- pattern (-1989281089) se.liu.ida.hefquin.base.query.impl.BGPImpl
{ ?cc  rdfs:label  ?o }

--- pattern (-136179866) se.liu.ida.hefquin.base.query.impl.BGPImpl
{ ?c        owl:sameAs            ?cc .
<http://dbpedia.org/resource/Berlin>
          <http://dbpedia.org/ontology/country>  ?c }
```
while
```sparql
PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs:   <http://www.w3.org/2000/01/rdf-schema#>
PREFIX owl:    <http://www.w3.org/2002/07/owl#>
PREFIX dbo:    <http://dbpedia.org/ontology/>

SELECT * WHERE {
	SERVICE <http://dbpedia.org/sparql> {
		<http://dbpedia.org/resource/Berlin> dbo:country ?c .
		?o owl:sameAs ?cc
	}
	SERVICE <https://query.wikidata.org/sparql> {
		?cc rdfs:label ?o
	}
}
```
no longer crashes but instead produces:
```
--------- Physical Plan ---------
UNION-based bind join (7) 
│  - lop: gpAdd (6)
│  - fm (3) SPARQL endpoint at https://query.wikidata.org/sparql
│  - pattern (-1989281089): { ?cc rdfs:label ?o }
│  - certain variables: ?cc, ?o, ?c
│  - possible variables: none
│  - query planning info: none
│
└── req (4)
      - lop: req (1)
      - fm (0) SPARQL endpoint at http://dbpedia.org/sparql
      - pattern (-136180704): { <http://dbpedia.org/resource/Berlin> <[...]tology/country> ?c . ?o owl:sameAs ?cc }
      - certain variables: ?o, ?cc, ?c
      - possible variables: none
      - query planning info:  cardinality = 136614431 (accurate)
                              max cardinality = 136614431 (accurate)
                              min cardinality = 136614431 (accurate)
    
--- pattern (-1989281089) se.liu.ida.hefquin.base.query.impl.BGPImpl
{ ?cc  rdfs:label  ?o }

--- pattern (-136180704) se.liu.ida.hefquin.base.query.impl.BGPImpl
{ <http://dbpedia.org/resource/Berlin>
          <http://dbpedia.org/ontology/country>  ?c .
?o        owl:sameAs            ?cc }
```

@hartig, please let me know what you think of this approach. Anything that I've missed perhaps or maybe we can use something simpler?

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpSymmetricHashJoin.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpSymmetricHashJoin.java
@@ -51,10 +51,10 @@ public class PhysicalOpSymmetricHashJoin extends BaseForPhysicalOpBinaryJoin
 		visitor.visit(this);
 	}
 
-    @Override
-    public String toString(){
-       return "> symmetricHashJoin ";
-    }
+	@Override
+	public String toString() {
+		return "> symmetricHashJoin ";
+	}
 
 	public static Factory getFactory() {
 		return factory;


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/PhysicalOpProvider.java
```diff
@@ -0,0 +1,30 @@
...
+	 * Creates a physical operator for the given logical operator.
+	 *
+	 * Precondition: call only if {@link #supports(...)} has returned true for the
+	 * same logical operator and input variables.
+	 *
+	 * @param lop the logical operator
+	 * @return a physical operator
+	 * @throws IllegalArgumentException if the precondition was violated
+	 */
+	PhysicalOperator create( final LogicalOperator lop );
```
hartig, line 29: In interfaces, using `final` for method arguments is meaningless  ;-)
```suggestion
	boolean supports( LogicalOperator lop, ExpectedVariables inputVars );

	/**
	 * Creates a physical operator for the given logical operator.
	 *
	 * Precondition: call only if {@link #supports(...)} has returned true for the
	 * same logical operator and input variables.
	 *
	 * @param lop the logical operator
	 * @return a physical operator
	 * @throws IllegalArgumentException if the precondition was violated
	 */
	PhysicalOperator create( LogicalOperator lop );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/PhysicalOpFactory.java
```diff
@@ -0,0 +1,57 @@
...
+     * @return this factory (for chaining)
+     */
+	public PhysicalOpFactory register( final PhysicalOpProvider provider ) {
+		providers.add(provider);
+		return this;
+	}
+
+	/**
+     * Creates a physical operator for the given logical operator by
+     * consulting registered providers in order, or throws {@link UnsupportedOperationException}
```
hartig, line 41: I think, a better exception to throw in this case is `NoSuchElementException`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithBoundJoin.java
```diff
@@ -76,4 +82,46 @@ public boolean equals( final Object o ) {
...
+			else if ( lop instanceof LogicalOpGPOptAdd op ) {
+				return new PhysicalOpBindJoinWithBoundJoin(op);
+			}
+
+			throw new UnsupportedOperationException( "Unsupported type of logical operator: " + lop.getClass().getName() + "." );
+		}
+
+		private boolean isSupported( final FederationMember fm,
+		                                    final SPARQLGraphPattern pattern,
+		                                    final ExpectedVariables vars )
```
hartig, line 113: Something is wrong here with the indentation.

hartig (changes requested):
Good start!! I have a few smaller comments directly within the code, and a more general one is: The naming of the main new classes is not what I would expect from having seen the factory pattern in, e.g., Jena. What is now called `PhysicalOpProvider` should be called `PhysicalOpFactory` (yes, every type of physical operator has its own factory object), and what is now called `PhysicalOpFactory` should be called `PhysicalOpRegistry`.

hartig:
Is this ready to be reviewed again?

There is a failure in the CI run.

keski:
Ready for review. I've added factories and tests for all physical operators except `PhysicalOpParallelMultiLeftJoin`.

keski:
I've now substituted the logic in `LogicalToPhysicalOpConverter` to use the `PhysicalOpRegistry` with default factories for conversion into physical operators. This leaves the other files unchanged, except that we now need to cast the physical operators returned to `PhysicalPlanFactory` and `ExecOpParallelMultiwayLeftJoin`, since `LogicalToPhysicalOpConverter#convert` only returns `PhysicalOperator`.

I've commented out the registration of the factories that are not used, but we may wish to clean that up before merge.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpBindJoinWithBoundJoin.java
```diff
@@ -76,4 +84,49 @@ public boolean equals( final Object o ) {
...
+		private boolean isSupported( final FederationMember fm,
+		                             final SPARQLGraphPattern pattern,
+		                             final ExpectedVariables vars )
+		{
+			return (fm instanceof SPARQLEndpoint) && hasNonJoiningVar(pattern, vars);
+		}
+
+		private boolean hasNonJoiningVar( final SPARQLGraphPattern pattern, final ExpectedVariables vars ) {
+			final Set<Var> certainVars = vars != null ? vars.getCertainVariables() : Set.of();
+			final Set<Var> possibleVars = vars != null ? vars.getPossibleVariables() : Set.of();
```
hartig, line 122: To avoid creating Java objects for the empty set in the `vars == null` case, it would be better to handle this case first:
```suggestion
			if ( vars == null )
				return true;

			final Set<Var> certainVars = vars.getCertainVariables();
			final Set<Var> possibleVars = vars.getPossibleVariables();
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpFilter.java
```diff
@@ -41,4 +44,23 @@ public LogicalOpFilter getLogicalOperator() {
...
 	}
 
+	public static class Factory implements PhysicalOpFactory
+	{
+		@Override
+		public boolean supports( final LogicalOperator lop, final ExpectedVariables inputVars ) {
+			if( lop instanceof LogicalOpFilter ){
+				return true;
+			}
+			return false;
```
hartig, line 54: Maybe simplify as follows:
```suggestion
			return ( lop instanceof LogicalOpFilter );
```

hartig: This one is applicable also in several other cases here.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalToPhysicalOpConverter.java
```diff
@@ -24,125 +13,32 @@
...
+		.register( new PhysicalOpMultiwayUnion.Factory() )         // Multiway union
+		.register( new PhysicalOpBind.Factory() )                  // Bind clauses
+		.register( new PhysicalOpFilter.Factory() )                // Filter clauses
+		.register( new PhysicalOpRequest.Factory() )               // Request at a federation member
+		.register( new PhysicalOpGlobalToLocal.Factory() )         // Apply vocab mappings global to local
+		.register( new PhysicalOpLocalToGlobal.Factory() )         // Apply vocab mappings local to global
+		.register( new PhysicalOpBindJoin.Factory() )              // Bind-join for brTPF interface
+		.register( new PhysicalOpBindJoinWithBoundJoin.Factory() ) // Bind-join for SPARQL interface
+		.register( new PhysicalOpBindJoinWithUNION.Factory() )     // (fallback) if no non-joining var available
+		// .register( new PhysicalOpBindJoinWithVALUESorFILTER.Factory() )
```
hartig, line 27: PhysicalOpBindJoinWithVALUESorFILTER is a better fallback (at least for now -- let's see what Adriana's experiments will show us).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalToPhysicalOpConverter.java
```diff
@@ -24,125 +13,32 @@
...
-	public static NaryPhysicalOp convert( final LogicalOpMultiwayLeftJoin lop ) {
-		throw new UnsupportedOperationException();
+	public static PhysicalOperator convert( final LogicalOperator lop ) {
+		return convert(lop, null);
 	}
 
-	public static NaryPhysicalOp convert( final LogicalOpMultiwayUnion lop ) {
-		return new PhysicalOpMultiwayUnion();
+	public static PhysicalOperator convert( final LogicalOperator lop, final ExpectedVariables inputVars ) {
+		return registry.create(lop, inputVars);
```
hartig, line 42: With this change, we will have to do a cast in almost every place where we call this `convert` function, like you have done now, for instance, in your changes to `PhysicalPlanFactory`. I would prefer to keep these casts local within `LogicalToPhysicalOpConverter`. To this end, you can replace each of the two `convert` functions here by four different functions:
* NullaryPhysicalOp convert( final NullaryLogicalOp lop )
* UnaryPhysicalOp convert( final UnaryLogicalOp lop )
* BinaryPhysicalOp convert( final BinaryLogicalOp lop )
* NaryPhysicalOp convert( final NaryLogicalOp lop )

... and likewise for the version with `ExpectedVariables` as second argument.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
```diff
@@ -396,7 +396,7 @@ public static PhysicalPlan createPlan( final BinaryLogicalOp rootOp,
 	                                       final QueryPlanningInfo qpInfo,
 	                                       final PhysicalPlan subplan1,
 	                                       final PhysicalPlan subplan2 ) {
-		final BinaryPhysicalOp pop = LogicalToPhysicalOpConverter.convert(rootOp);
+		final BinaryPhysicalOp pop = (BinaryPhysicalOp) LogicalToPhysicalOpConverter.convert(rootOp);
```
hartig, line 399: This doesn't seem right. For binary operators we need to be able to check whether `subplan1.getExpectedVariables()` and `subplan2.getExpectedVariables()` can be used. To this end, it is actually necessary to change `PhysicalOpFactory#supports` to have the following signature:
```java
boolean supports( LogicalOperator lop, ExpectedVariables ... inputVars );
```
And likewise for `PhysicalOpRegistry#create`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
```diff
@@ -466,7 +466,7 @@ public static PhysicalPlan createPlan( final NaryLogicalOp rootOp,
 	public static PhysicalPlan createPlan( final NaryLogicalOp rootOp,
 	                                       final QueryPlanningInfo qpInfo,
 	                                       final List<PhysicalPlan> subplans ) {
-		final NaryPhysicalOp pop = LogicalToPhysicalOpConverter.convert(rootOp);
+		final NaryPhysicalOp pop = (NaryPhysicalOp) LogicalToPhysicalOpConverter.convert(rootOp);
```
hartig, line 469: Also for this one, we need to push the `ExpectedVariables` object of every subplan into the `convert` function and, there, further into `PhysicalOpRegistry#create` and then into `PhysicalOpFactory#supports`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/LogicalToPhysicalOpConverter.java
keski: Resolved in commit 5d9ca9d2fc6ac9939dd0e699252ed0817014c60a

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
keski: There are a few cases when we can explicitly create plans with a particular physical operator (e.g. `PhysicalPlanFactory#createPlanWithUnion` and `PhysicalPlanFactory#createPlanWithHashJoin`). Do we also want to check that the physical operator is supported in those cases or would this be the responsibility of the calling function?

Also, is it correct to assume that we would like to add the requirement that at least one variables is shared for `PhysicalOpHashJoin, `PhysicalOpNaiveNestedLoopsJoin`, and `PhysicalOpSymmetricHashJoin`? Which one should fallback/degrade to Cartesian product?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/PhysicalPlanFactory.java
```diff
@@ -180,7 +182,7 @@ public static PhysicalPlan createPlanWithBindJoin( final LogicalOpGPAdd lop,
 	 */
 	public static PhysicalPlan createPlanWithIndexNLJ( final LogicalOpGPAdd lop,
 	                                                   final PhysicalPlan subplan ) {
-		final UnaryPhysicalOp pop = new PhysicalOpIndexNestedLoopsJoin(lop);
+		final UnaryPhysicalOp pop = PhysicalOpIndexNestedLoopsJoin.Factory.getInstance().create(lop);
```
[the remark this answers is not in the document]
keski: I've made the physical operator constructors protected, but the factory classes have to be public for them to be accessible in classes outside its package (e.g., in `PhysicalPlanFactory`). Another alternative would be to instead let `getFactory()` return the public `PhysicalOpFactory` interface, but then we would instead have to cast the physical ops returned when calling `Factory#create` method.

hartig (approved):

Issue #492, opened 2026-01-14
More flexibility for parameterized REST interfaces


hartig:
Currently, the support of REST endpoints as wrapped federation members is limited in terms of the way in which the URL of the endpoint can be specified in the [federation description](https://liusemweb.github.io/HeFQUIN/doc/federation_description.html). In particular, currently it is only possible to specify query parameters for such URLs and all these query parameters are considered as mandatory. We should make this more flexible to support parameterizing parts of the URL itself, and to support parameters that are optional.

Within the federation description, this may look something like the following.
```turtle
ex:openMeteo
      fd:serviceURI <https://api.open-meteo.com/v1/forecast> ;
      fd:interface _:openMeteoRestEndpoint ;
      fd:wrapper [ ... ] .

_:openMeteoRestEndpoint a fd:RESTInterface ;
      fd:urlTemplate "https://api.open-meteo.com/v1/forecast?latitude={lat}&longitude={long}&current={curr}" ;
      fd:templateParameters (
            [ fd:paramName "lat" ; fd:paramType xsd:float ; fd:mandatoryParam true  ]
            [ fd:paramName "long" ; fd:paramType xsd:float ; fd:mandatoryParam true ]
            [ fd:paramName "curr" ; fd:paramType xsd:string ; fd:mandatoryParam false ; fd:defaultValue "temperature_2m,wind_speed_10m" ]
      ) .
```

The idea is that the `fd:urlTemplate` is a template of a URL and contains placeholders for the parameters, where these placeholder are denoted by a pair of curly braces with a name of one of the parameters in between these curly braces (e.g,, `{lat}`). While the example template contains such placeholders only for query parameters, it is possible to have such placeholders also somewhere within the path part of the URL template.

To form an actual URL from such a template, each placeholder is replaced by a value that is provided via the `PARAMS` feature of the corresponding `SERVICE` clause in the SPARQL query to be executed over the federation.

For each of the parameters, the `fd:mandatoryParam` property specifies whether the parameter is mandatory, in which case the value for the parameter must be given via the `PARAMS` feature. For non-mandatory parameters (such as the one named `"curr"` in the example), if there is no value provided via `PARAMS`, the given default value is used (hence, for every parameter for which `fd:mandatoryParam` is `false`, the federation description must contain a `fd:defaultValue` property).

Hence, for queries that are meant to access the federation member of the example above, it is necessary to write a `SERVICE` clause such as
```sparql
[...]

PR #493, opened 2026-01-20
First version solution for #492

DrJonasWestman:
See #492 for discussions.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl.java
@@ -137,7 +137,7 @@ public class TextBasedLogicalPlanPrinterImpl extends BaseForTextBasedPlanPrinter
 			final StringBuilder b = new StringBuilder("parameter variables:");
 			if ( op.hasParameterVariables() ) {
 				for ( final Map.Entry<String,Var> e : op.getParameterVariables().entrySet() )
-					b.append( " " + e.getValue().toString() + " AS " + e.getKey() + " " );
+					b.append( " " + e.getValue().toString() + " AS \"" + e.getKey() + "\" " );
 			}
 			else {
 				b.append( " none" );
--- a/ExampleQuery.rq
+++ b/ExampleQuery.rq
@@ -30,7 +30,7 @@ PREFIX ex:     <http://example.org/>
 #	BIND ( "52.52"^^xsd:float AS ?lat )
 #	BIND ( "13.41"^^xsd:float AS ?long )
 #	BIND ( "temperature_2m,wind_speed_10m" AS ?req)
-#	SERVICE <https://api.open-meteo.com/v1/forecast> PARAMS(?lat, ?long, ?req) {
+#	SERVICE <https://api.open-meteo.com/v1/forecast> PARAMS(?lat AS "latitude", ?long AS "longitude", ?req AS "current") {
 #		?x ex:temperature ?temp .
 #		?x ex:windSpeed ?wind .
 #	}
@@ -41,7 +41,7 @@ SELECT * WHERE {
 		<http://dbpedia.org/resource/Berlin> geo:lat ?lat ; geo:long ?long .
 	}
 	BIND ( "temperature_2m,wind_speed_10m" AS ?req)
-	SERVICE <https://api.open-meteo.com/v1/forecast> PARAMS(?lat, ?long, ?req) {
+	SERVICE <https://api.open-meteo.com/v1/forecast> PARAMS(?lat AS "latitude", ?long AS "longitude", ?req AS "current") {
 		?x ex:temperature ?temp .
 		?x ex:windSpeed ?wind .
 	}


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpLookupJoinViaWrapperWithParamVars.java
```diff
@@ -193,33 +192,35 @@ protected void _concludeExecution( final List<SolutionMapping> input,
 	 * @return the parameter values (in the form of a Jena {@link Node}
 	 *         objects) or {@code null}
 	 */
-	protected Node[] extractParamValues( final SolutionMapping sm ) {
+	protected Map<String, Node> extractParamValues( final SolutionMapping sm ) {
 		final Binding solmap = sm.asJenaBinding();
 
-		if ( solmap.size() < paramVars.size() ) return null;
```
hartig: Why did you remove this one?

Consider a case in which the federation description defines an IRI template with three parameters---say, "m1", "m2", and "o"---where "o" is optional and the other two are mandatory. Hence, a corresponding PARAMS part of a query may have either two or three entries. In the case that it has three entries---for instance, `PARAMS(?x AS "m1", ?y AS "m2", ?z AS "o")`---don't we expect that the solution mappings that are produced before the corresponding `SERVICE` clause (and, thus, are passed into this operator here) have bindings for all three variables (including ?z for the optional parameter)?

My understanding of the notion of "optional" for parameters was that such parameters are not required to be mentioned in a PARAMS part, but if they are mentioned, then it is an error if the earlier parts of the query do not produce values for them.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpLookupJoinViaWrapperWithParamVars.java
```diff
@@ -193,33 +192,35 @@ protected void _concludeExecution( final List<SolutionMapping> input,
...
-			if ( ! paramDecl.getType().equals(typeOfNode) ) return null;
+			if (!paramDecl.getType().equals(typeOfNode)) return null;
 
-			result[i++] = paramValueAsNode;
+			result.put(paramVarName, paramValueAsNode);
 		}
 
+		// TODO Also check that all required parameters are present... but this should
+		// already be done previously?
+
```
hartig: Yes, this check should have happened long before HeFQUIN executes a plan with this operator here. The best place do check would be in the constructor of the corresponding logical operator, which is `LogicalOpGPAdd`. I have added a review comment there.

And then you can delete this code comment here:

```suggestion

```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpLookupJoinViaWrapperWithParamVars.java
```diff
@@ -229,19 +230,11 @@ protected Node[] extractParamValues( final SolutionMapping sm ) {
...
-			final String paramValue = paramValueAsNode.getLiteralValue().toString();
-			params.add( paramDecl.getName(), paramValue );
-		}
-
-		return new RESTRequestImpl( fm.getURL(), params );
+	protected RESTRequest createRequest( final Map<String,Node> paramValues ) {
+		return new RESTRequestImpl(fm.getURLTemplate(), paramValues.entrySet().stream()
+				.collect(HashMap::new,
+						(m, e) -> m.put(e.getKey(), e.getValue().getLiteralValue().toString()),
+						HashMap::putAll));
```
hartig: I can imagine that you like this functional way of writing code but, at least for me, it makes it a bit difficult to quickly grasp what the code is doing, especially if things are nested several levels deep. Are you okay with at least disentangling this part by moving the creation of the second argument of the `RESTRequestImpl` constructor before the constructor call? Something like:
```suggestion
		final Map<String,String> params = paramValues.entrySet().stream().collect(
				HashMap::new,
				(m, e) -> m.put( e.getKey(), e.getValue().getLiteralValue().toString() ),
				HashMap::putAll );
		return new RESTRequestImpl( fm.getURLTemplate(), params );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/utils/TextBasedLogicalPlanPrinterImpl.java
```diff
@@ -136,7 +136,7 @@ public void visit( final LogicalOpGPAdd op ) {
 
 			final StringBuilder b = new StringBuilder("parameter variables:");
 			if ( op.hasParameterVariables() ) {
-				for ( final Var v : op.getParameterVariables() )
+				for ( final Var v : op.getParameterVariables().values() )
 					b.append( " " + v.toString() );
```
hartig: It would be useful to have the plan printer print not only the variables but also the parameters for which they are used. Something like:

```suggestion
				for ( final Map.Entry<String,Var> e : op.getParameterVariables().entrySet() )
					b.append( " " + e.getValue().toString() + " AS " + e.getKey() + " " );
```

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpLookupJoinViaWrapperWithParamVarsTest.java
```diff
@@ -203,8 +204,8 @@ public void twoSolMapsWithSameParamValueInSameBatch() throws ExecutionException
 		final Iterator<SolutionMapping> it = sink.getCollectedSolutionMappings().iterator();
 		assertTrue( it.hasNext() );
 		assertEquals( 4, it.next().asJenaBinding().size() );
-		assertTrue( it.hasNext() );
-		assertEquals( 4, it.next().asJenaBinding().size() );
+		// assertTrue( it.hasNext() );
+		// assertEquals( 4, it.next().asJenaBinding().size() );
```
hartig: I don't see the reason for this change. The changes in this PR should not have an effect on the solution mappings produced by the operator in this test.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpLookupJoinViaWrapperWithParamVars.java
DrJonasWestman: You are correct. I was confusing the paramVars with the declared/possible params. This check makes sense.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpLookupJoinViaWrapperWithParamVarsTest.java
hartig: Yes, the test worked before.

I have just investigated the issue and it turns out that the test worked because of a bug in my original version of the code! :-(

I have pushed a commit (https://github.com/LiUSemWeb/HeFQUIN/pull/493/commits/0d06cdd9776fb73374d794390224b078b7478f42) to this PR to fix the bug. The commit also restores the test to its original form and it passes now.

The fix was to change from a `Set` to a `List` in the snippet of code that you are quoting in your reply here. This way, the two identical solution mappings will still **_both_** be considered for the join that happens with the data obtained via the REST request. The reason why the problem didn't surface in my original code is that `paramValues` was of type `Node[]` and Java treats different _but identical_ objects of this type as _not_ equal, while different but identical objects of the type `Map<String,Node>` are considered as equal. As a consequence of this difference, `paramsForRequests.get(paramValues)` behaves different now in the snippet of code that you are quoting above.

hartig:
Excellent! Thanks a lot for the contribution! Let me know when you have something else; we can also set up a call if you want to discuss something (but notice that I will be on vacation next week -- vi ska åka till fjällen :-)

PR #538, opened 2026-03-19
Add support for multiple fd arguments for bin/hefquin-server

kentthang010:
Example usage:
Running the server with `"args": "--fd examples/ExampleFederation.ttl --fd examples/ExampleFederation2.ttl"`

Due to generic information being removed at runtime, the methods `withFederationCatalog( List<String> )` and `withFederationCatalog( List<Model> )` had to be renamed to avoid the `Method has the same erasure as another method` compilation error. Let me know if another solution is preferred. 

Also, I was not sure where to put some constants that I defined to avoid magical numbers/strings. These were put in `hefquin-base/.../HeFQUINConstants.java`. Please let me know if this was incorrect / proper place to put them or if they are necessary.

### DIFF
--- a/hefquin-service/src/main/java/se/liu/ida/hefquin/service/SharedResourceInitializer.java
+++ b/hefquin-service/src/main/java/se/liu/ida/hefquin/service/SharedResourceInitializer.java
@@ -25,7 +25,6 @@ public class SharedResourceInitializer implements ServletContextListener
 		
 		logger.info( "--- Initialize engine ---" );
 		logger.info( "hefquin.configuration:    {}", confDescr );
-		logger.info( "hefquin.federation.count: {}", fedCount );
 		
 		if ( fedCount == null ) {
 			final String fedCat = System.getProperty("hefquin.federation", "config/DefaultFedConf.ttl");
@@ -36,6 +35,7 @@ public class SharedResourceInitializer implements ServletContextListener
 			for ( int i = 0; i < Integer.parseInt(fedCount); i++ ) {
 				fedCatList.add( System.getProperty("hefquin.federation." + (i + 1)) );
 			}
+			logger.info( "hefquin.federation.count: {}", fedCount );
 			logger.info( "hefquin.federation.list:  {}", fedCatList );
 		}
 


hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHeFQUINServer.java
```diff
@@ -61,7 +62,16 @@ protected String getCommandName() {
...
+		System.setProperty( "hefquin.federation.count", modServer.getFederationCount() );
+		
+		if ( modServer.getFederationDescription().size() == HeFQUINConstants.DEFAULT_FED_DESCR_COUNT_INT ) {
+			System.setProperty( "hefquin.federation", modServer.getFederationDescription().get(0) );
+		}
+		else {
+			for ( int i = 0; i < modServer.getFederationDescription().size(); i++ ) {
+				System.setProperty( "hefquin.federation." + (i+1), modServer.getFederationDescription().get(i) );
+			}
+		}
```
hartig: The `"hefquin.federation.count"` property should be set only in the else case of the if-else statement. In other words, the properties `"hefquin.federation"` and `"hefquin.federation.count"` are mutually exclusive (either one is set or the other, never both).

Moreover, the constant 1 in the if-condition can be hardcoded in the code here:
```java
if ( modServer.getFederationDescription().size() == 1 ) {
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/modules/ModServer.java
```diff
@@ -42,10 +48,32 @@ public void processArgs( final CmdArgModule cmdLine ) {
...
+			fedDescr = new ArrayList<>();
+			
+			for ( final String filename : filenames ) {
+				if ( isURIOrExistingFile( filename ) ) {
+					fedDescr.add( filename );
+				}
+				else
+					cmdLine.cmdError( "Invalid federation description file: " + filename, false );
+			}
 		} else {
```
hartig:
```suggestion
		}
		else {
```

PR #546, opened 2026-03-24
Add ChronicleMap-backed persistent cache for requests

keski:
This PR introduces a persistent caching based on `ChronicleMap` for requests. The persistent cache stores resolved requests, while the in-memory cache is used to store complable futures, which avoids duplicate requests.


hartig (commented):
I have several questions about the things in this PR.

hartig (changes requested):
I here some review comments. But I have stopped half way when I saw that `FederationAccessManagerWithChronicleMapCache` creates new `TPFResponseImpl` and new `SolMapsResponseImpl` objects based on the `cachedObject`, and that `ChronicleMapCacheObject` has member variables for all these things (even if they remain `null` for the cases in which they are not needed). This all feels somehow not elegant, not only because it means to re-create Java objects but also because it means that `ChronicleMapCacheObject` will be overloaded further if we support additional types of responses in the future.

An idea to make this nicer (but I am not 100% sure whether this is going to work):
1. `ChronicleMapCacheObject` is turned into an interface.
2. The classes/interfaces that represent responses to be put into the cache (e.g., `TPFResponse`) extend this interface.
3. After doing `chronicleMapCache.get(key)`, `FederationAccessManagerWithChronicleMapCache` can then simply cast the resulting `cachedObject` to the correct type and use the resulting object as is (without re-creating another Java object from it).
4. `ChronicleMapCacheObjectMarshaller` needs to distinguish based on the concrete type of the given `ChronicleMapCacheObject`. That is, when writing, `ChronicleMapCacheObjectMarshaller` has and if-else-if-block with `instanceof` checks for the different possible types; within each case, the marshaller first writes some type-specific marker and, then, writes the object. When reading, the marshaller first reads the type marker, based on which it can then distinguish which type of object it continues reading.

There is one complication that I foresee with this idea: The interfaces that represent responses to be put into the cache assume that the Java objects for these responses contain the corresponding federation member and the corresponding request (see the `getFederationMember()` and `getRequest()` methods of `DataRetrievalResponse`). These may be difficult to marshal (or not?). But perhaps we can also get rid of them?

I took a quick look and we can indeed get rid of these two methods; they are not really used anywhere in the end. So, what do you think if I create a separate PR first (to remove these methods) and, thereafter, you rebase this PR and implement the idea above? (But please also think about the idea first, maybe it has some other flaws that I don't see)

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCacheKey.java
```diff
@@ -0,0 +1,114 @@
...
+				throw new IllegalArgumentException( "Unexpected type of server: " + fm.getClass().getName() );
+		}
+		else if (    req instanceof BRTPFRequest brtpfRequest
+		          && fm instanceof BRTPFServer brtpfServer ) {
+			requestUrl = brtpfServer.createRequestURL(brtpfRequest);
+		}
+		else {
+			throw new IllegalArgumentException( "Unexpected request type: " + req.getClass().getName()
+					+ "(server type: " + fm.getClass().getName() + ")" );
+		}
```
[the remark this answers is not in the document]
keski: Issue created (#590)

keski:
The average startup time increases when activating the `ChronicleMapCache`, from approximately 0.7 seconds to 1.6 seconds (tested on Macbook Pro 2021 with M1 Max). The overhead seems relatively constant regardless of the cache size. When caching 0-200 `SolMapResponse`s with an average of >600 solution mappings the difference in startup time varied in the interval 1.6 to 2.0 seconds.

I'm not sure if this overhead is really motivated as the default access manager for the CLI, perhaps we should keep the non-persistent cache there for now. For the `hefquin-service`, however, it is a great fit since the overhead only matters during JVM startup.

A better fit for the CLI could be [MapDB](https://github.com/jankotek/mapdb/), where the maximum throughput is lower but the startup time is an order magnitude faster. MapDB is slightly less feature-rich than Chronicle Map, but actual the implementations would be very similar.

hartig:
Just to double check: are you still planning to add more commits to this PR?

hartig:
Also, is it okay for you if I create a new release before this PR gets merged? (It seems that the changes to CHANGELOG.md in this PR are already in conflict with `main`.)

hartig:
@keski 
1. Are you still planning to add more commits to this PR or do you consider it ready now?
2. Is it okay for you if I create a new release before this PR gets merged? (It seems that the changes to CHANGELOG.md in this PR are already in conflict with `main`.)

keski:
> @keski
> 
>     1. Are you still planning to add more commits to this PR or do you consider it ready now?
> 
>     2. Is it okay for you if I create a new release before this PR gets merged? (It seems that the changes to CHANGELOG.md in this PR are already in conflict with `main`.)

Sure, create the release. No more commits are planned for this one now (I only restored `config/` to use the `FederationAccessManagerWithCache` instead of the persistent one). If you create the release I can rebase.

hartig:
Thanks. Then, I will look at this PR after having created a release, and the next release thereafter can then contain all the new cache functionality.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithChronicleMapCache.java
```diff
@@ -1,226 +1,399 @@
...
+	protected static final long DEFAULT_TIME_TO_LIVE = 300_000; // 5 minutes
+	protected final ChronicleMapCache chronicleMapCache;
+
+	/**
+	 * Creates a federation access manager with a ChronicleMap-backed persistent
+	 * cache using the given capacity and cache policies.
+	 *
+	 * @param fedAccMan           the wrapped federation access manager
+	 * @param cacheCapacity       the maximum cache capacity
+	 * @param maxParallelRequests the maximum default number of parallel request to a given endpoint address
```
hartig: That's the one that will be removed in a follow-up PR, right?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithChronicleMapCache.java
```diff
@@ -1,226 +1,399 @@
...
+	 * @throws IOException if the persistent cache cannot be created or opened
+	 */
+	public FederationAccessManagerWithChronicleMapCache( final FederationAccessManager fedAccMan,
+	                                                     final int cacheCapacity,
+	                                                     final int maxParallelRequests,
+	                                                     final CachePolicies<ChronicleMapCacheKey, CompletableFuture<? extends DataRetrievalResponse<?>>, ChronicleMapCacheEntry> chronicleMapCachePolicies )
```
hartig, line 75: Please add a Javadoc comment for this argument (for the sake of consistency with the fact that there is such a comment for each of the other arguments).

Also for all the other constructors here.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithChronicleMapCache.java
```diff
@@ -1,226 +1,399 @@
...
+	 * @param req the data retrieval request (TPF, BRTPF, or SPARQL)
+	 * @param fm  the federation member handling the request
+	 * @return a {@link CompletableFuture} with the cardinality response
+	 *
+	 * @throws FederationAccessException if the request fails
+	 * @throws IllegalStateException     if the request/member combination is
+	 *                                   unsupported
+	 */
+	public < ReqType extends DataRetrievalRequest,
+	         RespType extends DataRetrievalResponse<?>,
```
hartig: `RespType` is not needed here, is it?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+		invalidPolicy = policies.getInvalidationPolicy();
+
+		this.filename = filename;
+		map = initializeMap(filename, capacity);
+		inMemoryCache = new HashMap<>(capacity);
+
+		// Initialize replacement policy
+		for( final Entry<ChronicleMapCacheKey, ChronicleMapCacheEntry> entry : map.entrySet() ) {
+			replacementPolicy.entryWasAdded( entry.getKey(), entry.getValue() );
+		}
```
hartig, line 159: Is that what makes the initialization process time consuming?

Doesn't ChronicleMap has its own replacement policies? Or, maybe not. It is really only a disk-backed `Map` not a full-blown cache, right?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+
+		// Enforce max capacity
+		while ( map.size() > capacity ) {
+			final Iterable<ChronicleMapCacheKey> evictionCandidates = replacementPolicy.getEvictionCandidates(1);
```
hartig: Why do you evict one entry at a time here? You should be able to call `getEvictionCandidates` with `map.size() - capacity`.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+			final ChronicleMapCacheKey evictionCandidate = evictionCandidates.iterator().next();
+			evict(evictionCandidate);
+		}
+
+		logger.info( toString() );
```
hartig: Please include a message, saying something "ChronicalMap-based cache created"

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+	 * Stores the given key-value pair in the cache.
+	 *
+	 * <p>
+	 * The value is written to the ChronicleMap and also inserted into the in-memory
+	 * cache to avoid future deserialization overhead.
+	 * </p>
+	 *
+	 * <p>
+	 * If the key already exists, the corresponding entry is replaced and the
+	 * replacement policy is notified accordingly. Otherwise, entry is added. If the
```
hartig:
```suggestion
	 * replacement policy is notified accordingly. Otherwise, the entry is added. If the
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+			replacementPolicy.clear();
+		}
+	}
+
+	/**
+	 * Stores the given key-value pair in the cache.
+	 *
+	 * <p>
+	 * The value is written to the ChronicleMap and also inserted into the in-memory
+	 * cache to avoid future deserialization overhead.
```
hartig, line 246: Does this mean that the in-memory cache will be a complete copy of the ChronicleMap, at least with respect to newly inserted entries?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+				if ( oldEntry == null )
+					replacementPolicy.entryWasAdded(k, entry);
+				else
+					replacementPolicy.entryWasRewritten(k, entry);
+				return entry;
+			} );
+
+			// Check capacity
+			if ( map.size() > capacity ) {
+				final Iterable<ChronicleMapCacheKey> evictionCandidates = replacementPolicy.getEvictionCandidates(1);
```
hartig: For this one, it is more logical to try to evict only one entry. Yet, it wouldn't hurt to also evict `map.size() - capacity` many, just to get back on track if there is some inconsistency somewhere.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+	public CompletableFuture<? extends DataRetrievalResponse<?>> get( final ChronicleMapCacheKey key ) {
+		synchronized (map) {
+			final ChronicleMapCacheEntry e = inMemoryCache.get(key);
+			if ( e != null ) {
+				if( invalidPolicy.isStillValid(e) ) {
+					replacementPolicy.entryWasRequested(key, e);
+					return e.getObject();
+				}
+				else {
+					inMemoryCache.remove(key);
```
hartig: Remove the entry from the ChronicleMap immediately here and then return `null`. This way, you avoid having to do another call of `isStillValid` below, for which it is already clear that it will return `false`.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
```diff
@@ -0,0 +1,414 @@
...
+				}
+			}
+
+			final ChronicleMapCacheEntry entry = map.get(key);
+			if ( entry == null )
+				return null;
+
+			if ( ! invalidPolicy.isStillValid(entry) ) {
```
hartig, line 328: This is the `isStillValid` that you can avoid having to do again for entries that had already been in the in-memory cache.

hartig (changes requested):
Nothing serious anymore. Just a few suggestions and some questions :-)

hartig:
@keski did you see my review?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithChronicleMapCache.java
keski: Yes, exactly. Not having to pass `maxParallelRequests` will make the constructors much cleaner.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithChronicleMapCache.java
keski: Ah, that's correct, `RespType` is no longer used here! I'll remove it.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
keski: > Is that what makes the initialization process time consuming?

This process is actually fast. The time consuming part is where the map is actually created:
```java
ChronicleMap.of( ChronicleMapCacheKey.class, ChronicleMapCacheEntry.class )
			.name("cache-map")
			.entries(10 * capacity)
			.averageKeySize(248)
			.averageValueSize(4096)
			.maxBloatFactor(10.0)
			.valueMarshaller(ChronicleMapCacheEntryMarshaller.INSTANCE)
			.createPersistedTo(file);
```
This process takes more or less the same amount of time, even if the map is empty.


> Doesn't ChronicleMap has its own replacement policies? Or, maybe not. It is really only a disk-backed `Map` not a full-blown cache, right?

Correct, it has no caching policies of its own.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
keski: I think I had some concurrency related thing in mind here but now I don't see any need for it. I'll update it to use get all eviction candidates directly instead. Both cleaner and faster.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
keski: Yes, this is correct. It keeps an in-memory map of all newly inserted and retrieved entries. It also makes sure that the same `CompletableFuture` object is returned for all request for the same key within a query execution. When retrieved from the `ChronicleMap` it could (in principle) have to read from disk and reconstruct a new completed future.  

It also improves performance significantly for `get` requests. For example, when running a simple performance benchmark I found that the in-memory map only negligibly increases insertion time, but retrieval time is an order of magnitude faster (or more) for the in-memory entries. The downside is of course the added complexity, and that it could (potentially) require more memory.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
keski: I agree, I'll use the same pattern as above.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
keski: The reason I did not return immediately was that I was taking into consideration that some other process (e.g., another instance of the query engine) could have updated the cache (on disk). That mapping could then still be "fresh" but only if retrieved from map (not the in-memory one). But this is not really a likely scenario.

I'll follow your suggestion instead.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
hartig: > Yes, this is correct. It keeps an in-memory map of all newly inserted and retrieved entries. [...]

Okay. Thanks for the confirmation.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/chroniclemap/ChronicleMapCache.java
hartig: Now that you mention it, is there actually a deeper issue here? I mean, in cases in which multiple instances of the engine have a copy of an entry from the ChronicleMap in their respective in-memory caches; if one of the instances decides to remove the entry (from both the ChronicleMap and its in-memory cache), then the other one would still have the entry in its in-memory cache. Can that lead to a problem?

hartig:
I see you pushed a great number of commits today. Let me know when the PR is ready for another review.

keski:
> I see you pushed a great number of commits today. Let me know when the PR is ready for another review.

All comments addressed.

hartig (approved):
Looks good now. You can merge it after you have updated the CHANGELOG in this PR, but please make sure that you only add new lines into the CHANGELOG (the diff in this PR currently shows that multiple of the existing lines are modified).

hartig:
@keski As mentioned in my review last week, you can merge this PR now (after updating the CHANGELOG in this PR).

Issue #544, opened 2026-03-23
Implement ExecOpDuplicateRemoval


hartig:
Implement an executable operator for DISTINCT. Hence, this operator should remove duplicates from the input stream of solution mappings and pass only distinct solution mappings to its output.

To this end, add a new Java class called `ExecOpDuplicateRemoval` in https://github.com/LiUSemWeb/HeFQUIN/tree/main/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops
Hence, class should be in the Java package `se.liu.ida.hefquin.engine.queryplan.executable.impl.ops`

This new class should extend the superclass `UnaryExecutableOpBase`. The most similar class to this new one is `ExecOpFilter` and the new one should override and implement the same methods, except for the version of `_process` that takes an iterator of solution mappings as input (this one is not needed for the new class).

As a first version, you can implement the functionality of this operator in the most naive way: the `_process` method simply puts all input solution mappings into a set (`java.util.Set` implemented via `java.util.HashSet`), which is a trivial way to remove duplicates. The `_concludeExecution` can then send all solution mappings from the set to the given sink.

Like `ExecOpFilter`, the new class should add `numberOfOutputMappingsProduced` to the statistics object created by `createStats`.

In contrast to `ExecOpFilter`, the constructor of the new class only needs the two arguments that are passed to the constructor of the superclass (i.e., `collectExceptions` and `qpInfo`).

To test the functionality of the new class, add a corresponding `ExecOpDuplicateRemovalTest` in https://github.com/LiUSemWeb/HeFQUIN/tree/main/hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops
To begin with, put one unit test in this class, namely one in which two *different* solution mappings are passed to the operator and, thus, both of them should become part of the output. Take a look at `ExecOpUnfoldTest` to get an idea how such a unit test may be implemented.

Further unit tests will need to be added once I have given you feedback on the first one.

PR #547, opened 2026-03-24
Add executable operator DISTINCT and tests

kentthang010:
Solves #544.

I opted to add two tests, the one @hartig mentioned in the original thread and another one which tests that two value-equivalent solution mappings are actually removed by the operator. Let me know if it is unnecessary.

### DIFF
--- a/hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpDuplicateRemovalTest.java
+++ b/hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpDuplicateRemovalTest.java
@@ -18,123 +18,123 @@ import se.liu.ida.hefquin.engine.queryplan.executable.impl.CollectingIntermediat
 
 public class ExecOpDuplicateRemovalTest 
 {
-    @Test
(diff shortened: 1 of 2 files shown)


hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpDuplicateRemovalTest.java
```diff
@@ -0,0 +1,78 @@
...
+        final CollectingIntermediateResultElementSink sink = new CollectingIntermediateResultElementSink();
+		
+        op.process(sm1, sink, null);
+        op.process(sm2, sink, null);
+        op.concludeExecution(sink, null);
+        
+        final Iterator<SolutionMapping> it = sink.getCollectedSolutionMappings().iterator();
+        
+        final SolutionMapping expectedSM1 = SolutionMappingUtils.createSolutionMapping(var1, URI1);
+        final SolutionMapping expectedSM2 = SolutionMappingUtils.createSolutionMapping(var2, URI2);
```
hartig: In this case you don't actually need to create these again. Instead, simply use `sm1` and `sm2` (as created above) for the following assert-tests.

Issue #545, opened 2026-03-23
Implement LogicalOpDedup


hartig:
Implement an executable operator for DISTINCT. We will later connect it to the executable operator that you implement for #544 

The Java class for this new operator should be called `LogicalOpDedup` and it needs to be placed in: 
https://github.com/LiUSemWeb/HeFQUIN/tree/main/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl

In this case, the most similar class that you can take inspiration from is `LogicalOpJoin` and you will need to override the same methods that `LogicalOpJoin` overrides. However, the Java interface the new class should implement is `UnaryLogicalOp`, and the implementation of the `getExpectedVariables` method should rather be as in `LogicalOpFilter`.

No need for unit tests for this new class.

However, you will have to also extend the `LogicalPlanVisitor` interface, and this will lead to the need to extend every class and subclass that implements this interface. Try to figure out from the existing code how to do that for the new `LogicalOpDedup`.

PR #550, opened 2026-03-25
Implement LogicalOpDedup

kentthang010:
Solves #545.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/cardinality/CardinalityEstimationWorkerImpl.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/cardinality/CardinalityEstimationWorkerImpl.java
@@ -468,9 +468,17 @@ public class CardinalityEstimationWorkerImpl implements CardinalityEstimationWor
 		final QueryPlanProperty maxIn = qpInfoSubPlan.getProperty(MAX_CARDINALITY);
 		final QueryPlanProperty minIn = qpInfoSubPlan.getProperty(MIN_CARDINALITY);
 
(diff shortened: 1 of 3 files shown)


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/cardinality/CardinalityEstimationWorkerImpl.java
```diff
@@ -457,7 +458,38 @@ public void visit( final LogicalOpUnion op ) {
...
+		final QueryPlanningInfo qpInfo = currentSubPlan.getQueryPlanningInfo();
+		final QueryPlanningInfo qpInfoSubPlan = currentSubPlan.getSubPlan(0).getQueryPlanningInfo();
+
+		final QueryPlanProperty crdIn = qpInfoSubPlan.getProperty(CARDINALITY);
+		final QueryPlanProperty maxIn = qpInfoSubPlan.getProperty(MAX_CARDINALITY);
+		final QueryPlanProperty minIn = qpInfoSubPlan.getProperty(MIN_CARDINALITY);
+
+		// Heuristic: assume 50% duplicates
+		final int crdValue = crdIn.getValue() / 2;
+		final Quality crdQuality = QueryPlanProperty.getReducedQuality( crdIn.getQuality() );
```
hartig: There is a special case here that I forgot to mention in my email: If `crdIn` is 0, then it needs to be copied over as is (i.e., with the quality score that it has).

Then, I notice that `crdValue` may become 0, namely if `crdIn.getValue()` is 1. This is another case in which `crdIn` needs to be copied over as is (i.e., with the quality score that it has).

PR #552, opened 2026-03-25
Add SHACL constraints

KentThang:
Builds upon @DrJonasWestman's initial SHACL file by adding constraints that were present in `FederationDescriptionReader.java` and `feddesc.ttl` but not in `feddesc.shacl.ttl`. Also changes the shape definitions to only accept xsd:anyURI literals.

May not cover all constraints present in `FederationDescriptionReader.java` or `feddesc.ttl`.

Ignores `BoltInterface`, `GraphQLEndpointInterface` and `MappingConfiguration` constraints for now as they may get reworked. It does not implement modular / incremental validation


DrJonasWestman:
I will happily do that but I think it makes more sense if I do it after our meeting on Friday. 
Even with the current vocabulary there are some uncertainties that makes it hard to evaluate whether or not the shapes are complete enough. 
Most importantly, we need to decide whether or not we want to assume that the verification will run on a graph with inference from the vocabulary or not.

hartig:
@kentthang010 With PR #597 merged, you can resume the work on the SHACL shapes for the feddesc vocabulary.

KentThang:
Latest commit rewrites `feddesc.shacl.ttl` to fit the new vocabulary! I wasn't quite sure how to handle the new wrapper conditions
> wrappers associated with a federation member whose interface has fd:GenericWebAPIProtocol as the fd:supportedProtocol, must have an fd:rmlTriplesMaps property; and for wrappers associated with a federation member whose interface has fd:BoltProtocol or fd:GraphQLProtocol as the fd:supportedProtocol, the wrapper must have an fd:mappingConfiguration property.

I solved it by adding two new shapes `RMLWrapperShape` and `MappingConfigWrapperShape` as well as an `sh:or` in `WrappedFederationMemberShape` that makes sure the wrapper conditions above are enforced.

Not sure if this is the cleanest solution, let me know if there is a better one!

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+# Main shape for FederationMember
+fd:FederationMemberShape
+    a sh:NodeShape ;
+    sh:or (
+        [ sh:class fd:RDFBasedFederationMember ;
+          sh:node  fd:RDFBasedFederationMemberShape ]
+        [ sh:class fd:WrappedFederationMember ;
+          sh:node  fd:WrappedFederationMemberShape ]
+    ) ;
+    sh:message "Interface must be one of the supported types" ;
```
hartig: Mentioning `Interface` in the message is wrong.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+        sh:datatype xsd:anyURI ;
+        sh:message "Each federation member must have exactly one fd:serviceURI." ;
+    ] ;
+    sh:property [
+        sh:path fd:interface ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+        sh:node fd:InterfaceShape ;
+        sh:message "Each federation member must have exactly one fd:interface." ;
+    ] .
```
hartig, line 32: Given that the `sh:targetClass` of this shape is `fd:FederationMember`, these constraints are checked only for cases such as:
```turtle
ex:fm a fd:FederationMember ;
      a fd:RDFBasedFederationMember ;
      # ...
```
and
```turtle
ex:fm a fd:FederationMember ;
      a fd:WrappedFederationMember ;
      # ...
```
but not in cases such as
```turtle
ex:fm a fd:RDFBasedFederationMember ;
      # ...
```
and
```turtle
ex:fm a fd:WrappedFederationMember ;
      # ...
```
Can you change it such that these constraints are checked even if something is only stated to be of `rdf:type fd:RDFBasedFederationMember` or of `rdf:type fd:WrappedFederationMember` but not explicitly stated to be `rdf:type fd:FederationMember` ? (for which I guess you need to make two separate copies of this part of the shape?)
Or is the idea that this shape (and all the shapes in this file) is meant to be used for validation only after adding all entailments to the graph to be validated?

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+
+fd:RDFBasedFederationMemberShape
```
hartig, line 34: Another constraint for this shape is that an `fd:RDFBasedFederationMember` is not also an `fd:WrappedFederationMember`.

(Same comment for the `fd:WrappedFederationMemberShape` below.)

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+        sh:message "A federation member can have at most one fd:vocabularyMappingsFile." ;
+    ] .
+
+fd:WrappedFederationMemberShape
+    a sh:NodeShape ;
+    sh:property [
+        sh:path fd:wrapper ;
+        sh:maxCount 1 ;
+        sh:node fd:WrapperShape ;
+        sh:message "A federation member can have at most one fd:wrapper." ;
```
hartig: Actually, it should be exactly one (i.e., minCount = 1)

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+          sh:node  fd:FixedEndpointInterfaceShape ]
+        [ sh:class fd:TemplateBasedInterface ;
+          sh:node  fd:TemplateBasedInterfaceShape ]
+        [ sh:class fd:FragmentInterface ;
+          sh:node  fd:FragmentInterfaceShape ]
+    ) ;
+    sh:message "Interface must be one of the supported types" .
+
+# Shape for Fixed Endpoint interfaces
+fd:FixedEndpointInterfaceShape
```
hartig, line 120: Same kind of comment as for `fd:RDFBasedFederationMemberShape` above.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+# Shape for Template-based interfaces
+fd:TemplateBasedInterfaceShape
+    a sh:NodeShape ;
+    sh:targetClass fd:TemplateBasedInterface ;
+    sh:property [
+        sh:path fd:vocabularyMappingsFile ;
+        sh:maxCount 0 ;
+        sh:datatype xsd:string ;
+        sh:message "Template-based interface cannot have a vocabulary mapping." ;
+    ] ;
```
hartig: It is not clear to me why you have this here? Especially, if you have this here, why don't you have it also in all the other interface-related shapes?

The only place where I may see something like this to fit would be in `fd:WrappedFederationMemberShape`.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+        sh:or (
+            [ sh:hasValue fd:SPARQLProtocol ]
+            [ sh:hasValue fd:BoltProtocol ]
+            [ sh:hasValue fd:GraphQLProtocol ]
+            [ sh:hasValue fd:GenericWebAPIProtocol ]
+        ) ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+        sh:message "Fixed endpoint interface must have exactly one fd:supportedProtocol." ;
+    ] .
```
hartig, line 146: Will this message be printed only for cases in which a `FixedEndpointInterface` has != 1 `supportedProtocol` property or also in cases in which a `FixedEndpointInterface` has exactly one `supportedProtocol` property but with a wrong value? (in which the message is misleading)

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+        sh:message "URI template must have exactly one hydra:template." ;
+    ] ;
+    sh:property [
+        sh:path hydra:mapping ;
+        sh:node fd:MappingShape ;
+        sh:message "URI template mappings must conform to mapping shape." ;
+    ] .
+
+# Shape for Hydra Mapping
+fd:MappingShape
```
hartig: Better call this `HydraMappingShape`.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+        sh:path hydra:required ;
+        sh:maxCount 1 ;
+        sh:datatype xsd:boolean ;
+    ] ;
+    sh:property [
+        sh:path fd:paramType ;
+        sh:maxCount 1 ;
+        sh:nodeKind sh:IRI ;
+        sh:in ( xsd:string xsd:integer xsd:float xsd:double ) ;
+        sh:message "Parameter type must be either xsd:string, xsd:integer, xsd:float or xsd:double." ;
```
hartig:
```suggestion
        sh:in ( xsd:string xsd:integer xsd:decimal xsd:float xsd:double ) ;
        sh:message "Parameter type must be one of xsd:string, xsd:integer, xsd:decimal, xsd:float, or xsd:double." ;
```

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,263 @@
...
+
+# Shape for RML TriplesMap
+fd:RmlTriplesMapShape
+    a sh:NodeShape ;
+    sh:property [
+        sh:path rml:logicalSource ;
+        sh:minCount 1 ;
+        sh:nodeKind sh:BlankNodeOrIRI ;
+        sh:message "TriplesMap must have a logicalSource." ;
+    ] ;
```
hartig, line 285: It must be exactly one rml:logicalSource property (i.e., maxCount = 1 as well).

hartig (changes requested):

hefquin-vocabs/feddesc.shacl.ttl
KentThang: True, this is an error. It originates from `FederationDescriptionReader` where various endpoints (Rest APIs, GraphQl and Bolt) cannot have a vocabulary mapping.

For example:
```
if ( vocabMap != null )
	throw new IllegalArgumentException("REST APIs cannot have a vocabulary mapping.");
```

I added this to the old `feddesc.shacl.ttl` under the respective interfaces but it was moved here after the vocabulary change for some reason. I will remove this constraint from here but is it worthwhile to add it somewhere else? (Seeing as how it is a constraint in `FederationDescriptionReader`)

hefquin-vocabs/feddesc.shacl.ttl
KentThang: To clarify, you mean this comment?
> Another constraint for this shape is that an fd:RDFBasedFederationMember is not also an fd:WrappedFederationMember.

hefquin-vocabs/feddesc.shacl.ttl
KentThang: > Or is the idea that this shape (and all the shapes in this file) is meant to be used for validation only after adding all entailments to the graph to be validated?

That's probably not ideal for our case right?

Another possible solution could be to add a helper shape so we'd have:

- `fd:BaseFederationMemberShape`
  - with the common constraints (exactly one `serviceURI` & `interface`)
- `fd:FederationMemberShape`
  - with `sh:targetClass fd:FederationMember ;`
  - and `sh:or` ... `RDFBasedFederationMember` ... `WrappedFederationMember`
- `fd:RDFBasedFederationMemberShape`
  - with `sh:node fd:BaseFederationMemberShape ;`
- `fd:WrappedFederationMemberShape`
  - with `sh:node fd:BaseFederationMemberShape ;`

I believe this would solve it? Although, I am not sure if this is a cleaner solution than simply duplicating the constraints for both shapes.

DrJonasWestman: Another way could be to enforce explicit typing of the lowest level types... so instead of checking whether it conforms to one of the lower level shapes, we check that it has rdf:type with one them. Then the other shapes will be applicable and then they can also refer to the abstract one with sh:node fd:FederationMemberShape.

That is replacing: 
sh:or (
        [ sh:class fd:RDFBasedFederationMember ;
          sh:node  fd:RDFBasedFederationMemberShape ]
        [ sh:class fd:WrappedFederationMember ;
          sh:node  fd:WrappedFederationMemberShape ]
    ) ;
 with the type based check.
 I think perhaps that this is what we want if we do not want to rely on inference. Then we want to enforce that they are explictly typed.

DrJonasWestman: Looking more carefully... we are already doing that with the sh:class in the sh:or... 

so, I think just removing the sh:node reference here would be okay... and then instead refer to the abstract shape from the more concrete ones

That is... in fd:FederationMemberShape:
      sh:or (
          [ sh:class fd:RDFBasedFederationMember ]
          [ sh:class fd:WrappedFederationMember ]
      ) ;
      
and then in e.g. fd:RDFBasedFederationMemberShape
       sh:node fd:FederationMemberShape;
       sh:targetClass fd:RDFBasedFederationMember;

KentThang: That sounds good. Should I go with that, @hartig?

hefquin-vocabs/feddesc.shacl.ttl
hartig: The background of my comment was that `fd:vocabularyMappingsFile` is not a property of `fd:Interface` (or any of its sub-classes).

This constraint should be moved to `fd:WrappedFederationMemberShape`

hefquin-vocabs/feddesc.shacl.ttl
hartig: > That sounds good. Should I go with that, @hartig?

Sure, as long as it works for cases such as
```turtle
ex:fm a fd:RDFBasedFederationMember ;
      # ...
```
and
```turtle
ex:fm a fd:WrappedFederationMember ;
      # ...
```
(i.e., cases in `ex:fm` is not explicit stated to be of type `fd:FederationMember`.

hefquin-vocabs/feddesc.shacl.ttl
hartig: Yes, that's the comment that I meant. In other words, the different types of interfaces are mutually exclusive and there should be a constraint to check that every interface is of only one of these types.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+PREFIX xsd:    <http://www.w3.org/2001/XMLSchema#>
+PREFIX sh:     <http://www.w3.org/ns/shacl#>
+PREFIX fd:     <http://w3id.org/hefquin/feddesc#>
+PREFIX hydra:  <http://www.w3.org/ns/hydra/core#>
+PREFIX rml:    <http://w3id.org/rml/>
+
+# Main shape for FederationMember
+fd:FederationMemberShape
+    a sh:NodeShape ;
+    sh:or (
```
DrJonasWestman: Can't we just put xone here and then remove sh:not cases in the more refined shapes?

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+        sh:maxCount 1 ;
+        sh:node fd:InterfaceShape ;
+        sh:message "Each federation member must have exactly one fd:interface." ;
+    ] .
+
+fd:RDFBasedFederationMemberShape
+    a sh:NodeShape ;
+    sh:node fd:FederationMemberShape ;
+    sh:targetClass fd:RDFBasedFederationMember;
+    sh:not [
```
DrJonasWestman: Can possibly be removed?

KentThang:
Latest commit resolves all PR comments. 

Although, while testing, I noticed that `ExampleFederation.ttl` fails validation because `fd:rmlTriplesMaps` points to an `rdf:List` while the shape applies `fd:RmlTriplesMapShape` directly to `fd:rmlTriplesMaps`. 

So in other words, I think the validator currently validates the RDF list node itself against `fd:RmlTriplesMapShape`, instead of validating the individual `rml:TriplesMap` elements inside the list.

At this point, do we want to:
1)  just simplify the constraint to check that fd:rmlTriplesMaps exists and points to an RDF node/list or
2) add proper RDF list traversal?

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+        sh:maxCount 1 ;
+        sh:datatype xsd:string ;
+        sh:message "RDF-based federation member can have at most one fd:vocabularyMappingsFile." ;
+    ] .
+
+fd:WrappedFederationMemberShape
+    a sh:NodeShape ;
+    sh:node fd:FederationMemberShape ;
+    sh:targetClass fd:WrappedFederationMember;
+    sh:not [
```
DrJonasWestman: Can possibly be removed?

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+                sh:node fd:MappingConfigWrapperShape
+            ] ;
+            sh:message "Wrapped federation member using fd:BoltProtocol or fd:GraphQLProtocol must have a wrapper with fd:mappingConfiguration." ;
+        ]
+    ) .
+
+# Shape for interfaces
+fd:InterfaceShape
+    a sh:NodeShape ;
+    sh:or (
```
DrJonasWestman: switch to xone and remove the sh:node? Should we not follow the same pattern here as in the FederationMemberShape case? 
That is, if something is declared with a higher level interface type... we force it to also be typed with a more specific type. Then we now the more specific shapes for these will apply, and then we let them include the constraints from the more abstract shape with sh:node.

That is, in general, we force all instances to have concrete types where their shapes also include the constraints for shapes related to their more abstract types. I think this is a safe pattern to use given that we don't want to rely on inference. It can lead to redudant and not optimal checks but it is okay i would say. 

In practice that means that it is okay to declare something being a higher level type e.g. fd:Interface... but then the more concrete type most also be declared... and that will be detected.

DrJonasWestman: I think we should try to follow some general pattern/strategy here (such as the above). We need one given that we do not want to use inference. Perhaps we need a short meeting on this or are you following what i mean? :)

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+    ] ;
+    sh:message "Wrapped federation member cannot also be an fd:RDFBasedFederationMember." ;
+    sh:property [
+        sh:path fd:wrapper ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+        sh:node fd:WrapperShape ;
+        sh:message "Wrapped federation member must have exactly one fd:wrapper." ;
+    ] ;
+        sh:property [
```
hartig: The indentation of this one line is wrong.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+    ] ;
+    sh:property [
+        sh:path fd:interface ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+        sh:node fd:InterfaceShape ;
+        sh:message "Each federation member must have exactly one fd:interface." ;
+    ] .
+
+fd:RDFBasedFederationMemberShape
```
hartig, line 34: For the `WrappedFederationMemberShape` below you have the additional constraint to check that there is no `fd:vocabularyMappingsFile` property. For symmetry, it would make sense to add a constraint here, checking that an `RDFBasedFederationMember` does not have an `fd:wrapper` property.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+fd:InterfaceShape
+    a sh:NodeShape ;
+    sh:or (
+        [ sh:class fd:FixedEndpointInterface ;
+          sh:node  fd:FixedEndpointInterfaceShape ]
+        [ sh:class fd:TemplateBasedInterface ;
+          sh:node  fd:TemplateBasedInterfaceShape ]
+        [ sh:class fd:FragmentInterface ;
+          sh:node  fd:FragmentInterfaceShape ]
+    ) ;
```
hartig: Is it still necessary to have these `sh:node` entries?
In the case of `fd:FederationMemberShape` above, you have removed them now.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+        sh:path fd:endpointAddress ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+        sh:datatype xsd:anyURI ;
+        sh:message "Fixed endpoint interface must have exactly one fd:endpointAddress." ;
+    ] ;
+    sh:property [
+        sh:path fd:supportedProtocol ;
+
+        sh:minCount 1 ;
```
hartig, line 133:
```suggestion
        sh:path fd:supportedProtocol ;
        sh:minCount 1 ;
```

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+    sh:property [
+        sh:path fd:supportedProtocol ;
+        sh:or (
+            [ sh:hasValue fd:TPFProtocol ]
+            [ sh:hasValue fd:brTPFProtocol ]
+        ) ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+        sh:message "Fragment interface must have exactly one fd:supportedProtocol." ;
+    ] .
```
hartig, line 173: Please add a message also for the case that the value of this property is not one of the two expected values.

I notice that, for `FixedEndpointInterfaceShape` above, your have these as two separate `sh:property` statements. Not sure whether that is needed. In any case, it would be good to consistently use the same pattern.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+        sh:node fd:UriTemplateShape ;
+        sh:message "Template-based interface must have exactly one fd:uriTemplate." ;
+    ] ;
+    sh:property [
+        sh:path fd:supportedProtocol ;
+        sh:hasValue fd:GenericWebAPIProtocol ;
+        sh:minCount 1 ;
+        sh:maxCount 1 ;
+        sh:message "Template-based interface must have exactly one fd:supportedProtocol." ;
+    ] .
```
hartig, line 198: Also here, add a message for the case in which another value is used for this property.

hartig:
> Latest commit resolves all PR comments.

Review coming up..

> Although, while testing, I noticed that `ExampleFederation.ttl` fails validation because `fd:rmlTriplesMaps` points to an `rdf:List` while the shape applies `fd:RmlTriplesMapShape` directly to `fd:rmlTriplesMaps`.
> 
> So in other words, I think the validator currently validates the RDF list node itself against `fd:RmlTriplesMapShape`, instead of validating the individual `rml:TriplesMap` elements inside the list.

Indeed. That's not how it should be.

> At this point, do we want to:
>     1. just simplify the constraint to check that fd:rmlTriplesMaps exists and points to an RDF node/list or
>     2. add proper RDF list traversal?

The latter would be better. I think you should be able to use the SHACL Property Paths feature to navigate from the root list element (the object of the `fd:rmlTriplesMaps` triple in the federation description) to every `rdf:first` object to check that each of these objects either is `rdf:nil` or satisfies `fd:RmlTriplesMapShape`. The corresponding SPARQL property path expression would be `rdf:rest*/rdf:first`

hartig (changes requested):

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,300 @@
...
+PREFIX rml:    <http://w3id.org/rml/>
+
+# Main shape for FederationMember
+fd:FederationMemberShape
+    a sh:NodeShape ;
+    sh:or (
+        [ sh:class fd:RDFBasedFederationMember ]
+        [ sh:class fd:WrappedFederationMember ]
+    ) ;
+    sh:message "Federation member must be one of the supported types" ;
```
DrJonasWestman: Perhaps we need to do: 

sh:node [
        sh:xone (
            [ sh:class fd:RDFBasedFederationMember ]
            [ sh:class fd:WrappedFederationMember ]
        ) ;
        sh:message "Federation member must be typed as exactly one supported concrete member type: fd:RDFBasedFederationMember or fd:WrappedFederationMember." ;
    ] ;

hefquin-vocabs/feddesc.shacl.ttl
hartig: > Can possibly be removed?

`fd:RDFBasedFederationMember` and `fd:WrappedFederationMember` are mutually exclusive. I believe the `sh:xone` constraint that you propose in your other comment above wouldn't catch this _for cases_ in which only these two subclasses are used (because the `fd:FederationMemberShape` for which you propose the `sh:xone` constraint has `sh:targetClass fd:FederationMember`).

KentThang:
Latest commit resolves the PR comments above. Both the interface and federation member shapes now use the same pattern (`sh:xone` in the abstract shape and then `sh:node` in the more refined shapes) and I have also added RDF list traversal for `fd:RMLWrapperShape`.

hefquin-vocabs/feddesc.shacl.ttl
```diff
@@ -0,0 +1,296 @@
...
+            sh:property [
+                sh:path fd:wrapper ;
+                sh:node fd:MappingConfigWrapperShape
+            ] ;
+            sh:message "Wrapped federation member using fd:BoltProtocol or fd:GraphQLProtocol must have a wrapper with fd:mappingConfiguration." ;
+        ]
+    ) .
+
+# Shape for interfaces
+fd:InterfaceShape
```
hartig, line 107: For `fd:FederationMemberShape` you have `sh:targetClass fd:FederationMember` but here you do not have `sh:targetClass fd:Interface`. Is that an oversight or on purpose?

KentThang: Oversight!

hartig (approved):
From my perspective, this PR looks good now.

@DrJonasWestman what do you think?

PR #556, opened 2026-03-25
535 limited number of parallel requests to same server

keski:
This PR extends the `HttpClientProvider` to limit the number of parallel request against a specific endpoint/server. The default global value can be set via `HttpClientProvider#setDefaultMaxParallelRequests`, and specific limits can be configured for individual endpoints using `HttpClientProvider#registerEndpointLimiter`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [unreleased]
 
 ### Added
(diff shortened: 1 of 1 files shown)


hefquin-base/src/test/java/se/liu/ida/hefquin/base/shared/http/HttpClientProviderTest.java
```diff
@@ -0,0 +1,205 @@
+package se.liu.ida.hefquin.base.shared.http;
```
hartig: The package name should be changed accordingly.

PR #558, opened 2026-03-26
Make HeFQUIN aware of the new dedup operator

### LINKED ISSUE #551: Make HeFQUIN aware of the new dedup operator
Once the new logical operator and its corresponding executable operator are finished, we also need a physical operator for it and everything needs to be registered within HeFQUIN such that HeFQUIN will actually use it.

The physical operator should be called `PhysicalOpDuplicateRemoval` and goes into ./hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/ You can use `PhysicalOpBinaryUnion` as an example, but the superclass of the new class should be `UnaryPhysicalOpForLogicalOp` (instead of `BinaryPhysicalOpForLogicalOp`).

After creating the class for the physical operator, you need to add it to `PhysicalPlanVisitor`, which means you will also have to extend the class that implement this interface.

Finally, to hook everything together, three more steps are needed:

First, `ServiceClauseBasedSourcePlannerImpl` implements the functionality to convert an Apache Jena-based representation of the queries into a first logical plan of HeFQUIN. Extend `ServiceClauseBasedSourcePlannerImpl` by:
1. adding a function `createPlanForDistinct(OpDistinct, QueryProcContext)`, for which you can take `createPlanForFilter` as an example, and
2. adding an elseif-branch for it into `createPlan`.

Second, `OpExecutorHeFQUIN` overrides the functionality of the Jena query planner to use HeFQUIN for the queries that HeFQUIN can handle. Extend `OpExecutorHeFQUIN` by:
1. adding an `execute` function for `OpDistinct` (use the one of `OpBGP` as an example), and
2. changing the `visit` function for `OpDistinct` at the end of the file.

Third, `LogicalToPhysicalOpConverterImpl` implements the conversion from logical operators into physical operators by using the `PhysicalOpFactory` classes that the physical operators have. To register the factory of the new `PhysicalOpDuplicateRemoval`, edit the file ./config/DefaultConfDescr.ttl and add a corresponding entry around line 73. Do the same in the file ./config/DefaultConfDescrExtended.ttl  To pick up these updates for running the HeFQUIN CLI within your IDE, you may have to update or reimport the Maven project in your IDE (in Eclipse, I always have to do "Maven" -> "Update project .." for this to work).

That's it. Now you should be able to try queries with DISTINCT and, for instance, see the new operators in the printed query plans.

kentthang010:
Closes #551.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpDedup.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/impl/LogicalOpDedup.java
@@ -10,7 +10,7 @@ public class LogicalOpDedup implements UnaryLogicalOp
 
 	public static LogicalOpDedup getInstance() { return singleton; }
 
-	public LogicalOpDedup() {}
+	protected LogicalOpDedup() {}
 
 	@Override
 	public ExpectedVariables getExpectedVariables( final ExpectedVariables... inputVars ) {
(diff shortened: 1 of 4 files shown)


hartig (changes requested):
Just a few small things.

Issue #557, opened 2026-03-26
Implement logical-plan rewriting rules using the visitor pattern


hartig:
The classes that implement rewriting rules for logical plans (as used during the logical optimization process) should be changed such that they use the visitor pattern. At the moment, each of them contains some if-elseif-else block with cases for the different logical operators (probably not even all of these operators) and an exception in the final else-block. The problem with this is that, when adding a new logical operator, we may forget to extend all of these classes with a new case for that new operator. This problem can be avoided if these classes are implemented based on the `LogicalPlanVisitor` instead. A good example for how this may be done can be found in `CardinalityEstimationWorkerImpl` as well as in `MyPropertiesExtractor` of `TextBasedLogicalPlanPrinterImpl`.

The classes that need to be changed are:
* [x] MergeRequests
* [x] FilterPushDown
* [x] RemoveUnnecessaryL2gAndG2l
* [ ] ~~PullUpLtgOverUnion~~ (edit: this one shouldn't have been here)
* [x] ApplyVocabularyMappings

I suggest you take one at a time (in separate PRs). For most of them, we have unit tests, which you can use to check that your changes do not break the behavior. You can also add more unit tests where it seems reasonable.

hartig:
PR #575 for RemoveUnnecessaryL2gAndG2l was the last one. With this one merged, all tasks here are completed :-)

kentthang010:
Yippie!

PR #560, opened 2026-03-27
Refactor MergeRequests to use visitor pattern

kentthang010:
Addresses the `MergeRequests.java` section of #557.

The solution is not ideal since it makes the visitor stateful but not sure what better solution there is.
The `rewriteRoot` method is there to isolate the statefulness to one location.

`Visit()` for `LogicalOpFixedSolMap` is currently unimplemented.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
@@ -407,7 +407,7 @@ public class MergeRequests implements HeuristicForLogicalOptimization
 	 * merged pattern.
 	 */
 	protected LogicalPlan mergeSPARQLRequestsViaJoin( final FederationMember fm,
-														final List<LogicalPlan> reqPlans ) {
+	                                                  final List<LogicalPlan> reqPlans ) {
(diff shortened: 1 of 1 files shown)


hartig (changes requested):
It is indeed cleaner to introduce a helper class inline within `MergeRequests`.
```java
public class MergeRequests implements HeuristicForLogicalOptimization
{
	@Override
		...

	protected class Worker implements LogicalPlanVisitor {
		protected final List<LogicalPlan> rewrittenSubPlans;
		protected LogicalPlan returnPlan;

		public Worker( final List<LogicalPlan> rewrittenSubPlans ) {
			this.rewrittenSubPlans = rewrittenSubPlans;
		}

		public LogicalPlan getRewrittenPlan() { return returnPlan; }

		@Override
		public void visit( final LogicalOpRequest<?, ?> op ) {
			...

	} // end of Worker
}
```
Notice that the constructor of `Worker` does not have a `numberOfSubPlans` argument. It is not necessary to pass this number around explicitly because the value of it is just the size of the `rewrittenSubPlans` list. So, whichever `visit` method needs it, can just get it from `rewrittenSubPlans.size()`.

After adding this helper class, you can remove your `rewrite` method and, at the place where you call it, you create an instance of `Worker` instead.

kentthang010:
comments resolved!

hartig (approved):
Looks good now. I will merge it.

PR #575, opened 2026-04-02
Refactor RemoveUnnecessaryL2gAndG2l to use visitor pattern

kentthang010:
Addresses the `RemoveUnnecessaryL2gAndG2l` section of #557.

This hasn't been tested however and I am not sure if the recursion is working as intended with this refactor. Should I make a test?


hartig (changes requested):
A better way to get the recursion is to use `LogicalPlanWalker`. That is, in `extractTPs`, call
```java
LogicalPlanWalker.walk(plan, tpCollector, null);
```
where `tpCollector` is an instance of the visitor that you have added as a helper class called `Worker` (where a better name now would be `TriplePatternsCollector`). Within this visitor, you populate an initially empty `Set<TriplePattern>` by having the `visit` methods for `LogicalOpRequest`, `LogicalOpGPAdd`, and `LogicalOpGPOptAdd` add into this set; all other `visit` methods can remain empty with a comment such as:
```java
// nothing to do here; this operator does not contain any triple pattern
```

Additionally, yes, please add a few unit tests for this class.

kentthang010:
It should use LogicalPlanWalker the way you described now.

I also added some tests! Let me know if you want me to add more.

hartig (changes requested):

hartig (approved):

Issue #562, opened 2026-03-28
Add native support for projection


hartig:
The project operator is essentially what a SELECT clause with variables does. That is, a project operator has a **list** of variables and produces an output table with these variables as columns and the rows are the input solution mappings with their values for these variables.

However, the version of projection that we need first (and which this issue here is about) is only about restricting solution mappings in terms of the variables for which they have bindings. That is, given a **set** _V_ of variables of the project operator, for every input solution mapping _m_, the operator produces an output solution mapping that has only the variables from _m_ that are also in _V_.

The work on this issue can be separated into the following steps, with one PR per step.
* [x] Add a logical operator called LogicalOpProject, which includes extending LogicalPlanVisitor and all classes that implement it. The new logical operator is most similar to LogicalOpFilter. Hence, you can use the cases for that as a starting point for implementing the cases for LogicalOpProject.
* [x] Add an executable operator called `ExecOpProject`, including a class `ExecOpProjectTest` with unit tests for it.
* [x] Add a physical operator called `PhysicalOpProject` and wire everything together, like you have done for dedup as described in #551. The relevant Jena operator in this case is `OpProject`.
* [x] Add a rewriting rule called `ProjectPushDown`, including a class `ProjectPushDownTest` with unit tests for it. The similar type of rewriting rule, which you can use as a starting point, is `FilterPushDown`. Once it is done, the new `ProjectPushDown` needs to be registered in HeFQUIN by adding it after `FilterPushDown` in both `DefaultConfDescr.ttl` and `DefaultConfDescrExtended.ttl`.

kentthang010:
Ah, so I don't have to manually write `heuristics.add( new ProjectPushDown );` into `HeuristicsBasedLogicalOptimizerImpl`? Just adding it to the two ./config/DefaultConfDescr*.ttl files will be enough?

hartig:
Exactly!!

hartig:
With PR #580 merged, the work on this issue is complete :-)

PR #580, opened 2026-04-10
Add ProjectPushDown and tests

kentthang010:
Addresses the final part of #562 which closes it.


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithUnaryRootImpl;
+import se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicForLogicalOptimization;
+
+public class ProjectPushDown implements HeuristicForLogicalOptimization
+{
+
+/**
+ * Pushes project variables as much as possible towards the leaf nodes
+ *  of a given logical plan.
+ */
```
hartig, line 73: I assume that this was meant as a Javadoc comment for the class itself, rather than for the `apply` method. In this sense, it must be moved right _before_ the `public class ..` line.

Additionally, I would like you to extend the description such that it elaborates more on the cases in which this implementation pushes the project operator. I know that some of the other heuristics do not have a very detailed description either, but my goal is to have more detailed descriptions for them as well.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+	@Override
+	public LogicalPlan apply( LogicalPlan inputPlan ) {
```
hartig:
```suggestion
	public LogicalPlan apply( final LogicalPlan inputPlan ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+		public void visit( final LogicalOpMultiwayUnion op ) {
+			createdPlan = createPlanForUnionUnderProject( projectOp, subPlanUnderProject );
+		}
+
+		@Override
+		public void visit( final LogicalOpFilter op ) {
+			// Do not push the project operator below the filter operator.
+			// FilterPushDown has priority over ProjectPushDown because pushing
+			// filters reduces the number of solution mappings earlier, which
+			// is typically more beneficial than reducing their width.
```
hartig, line 210: Great that you included this comment here!!!

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+	                                                     final LogicalPlan subPlanUnderBind,
+	                                                     final LogicalPlan inputPlan ) {
+		// Check whether the project can be pushed under the bind operator.
+		// This is only safe if all variables required to evaluate the bind
+		// expressions are preserved by the projection. Otherwise, pushing
+		// the project would remove variables needed by the bind.
+		final Set<Var> varsInProject = projectOp.getVariables();
+		final Set<Var> varsUsedInBind = new HashSet<>();
+
+		for ( Expr expr : bindOp.getBindExpressions().getExprs().values() ) {
```
hartig:
```suggestion
		for ( final Expr expr : bindOp.getBindExpressions().getExprs().values() ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+
+	protected LogicalPlan createPlanForBindUnderProject( final LogicalOpProject projectOp,
+	                                                     final LogicalOpBind bindOp,
+	                                                     final LogicalPlan subPlanUnderBind,
+	                                                     final LogicalPlan inputPlan ) {
+		// Check whether the project can be pushed under the bind operator.
+		// This is only safe if all variables required to evaluate the bind
+		// expressions are preserved by the projection. Otherwise, pushing
+		// the project would remove variables needed by the bind.
+		final Set<Var> varsInProject = projectOp.getVariables();
```
hartig: To simplify the code, remove this and use `projectOp.getVariables()` directly in the if-condition below.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+
+	protected LogicalPlan createPlanForUnfoldUnderProject(
+			final LogicalOpProject projectOp,
+			final LogicalOpUnfold unfoldOp,
+			final LogicalPlan subPlanUnderUnfold,
+			final LogicalPlan inputPlan ) {
+		// Check whether the project can be pushed under the given
+		// unfold operator, which is possible only if none of the
+		// variables assigned by the unfold operator are included in
+		// the project variables.
```
hartig: Multiple observations here:
1. The unfold operator also has an expression (similar to the bind operator). The project operator cannot be pushed if the variables mentioned in that unfold expression are not a subset of the project variables. In other words, you need the same kind of check here as you have it for the case of bind operators.
2. The condition that you mention here is, however, still relevant. In fact, the same kind of condition is also relevant for the bind operator! If the variables assigned by the bind operator are not all in the set of projection variables, then we need to keep the projection operator on top of the bind operator. Likewise for unfold.
3. Yet, even then, it may be possible to duplicate the project operator and push one of them under the bind (resp. unfold). Of course, this can be done only if the variables mentioned in the bind expressions are a subset of the set of projection variables. In this case, for the pushed duplicate of the project operator, it is necessary to remove the bind variables from the set of projection variables in the pushed project operator. The advantage of duplicating a project operator in this way is that the pushed part of the operator may then be pushed further down into the subplan.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+	 * Assumes that the given child operator is either a {@link LogicalOpLocalToGlobal}
+	 * or a {@link LogicalOpGlobalToLocal}.
+	 */
+	protected LogicalPlan createPlanForL2GOrG2LUnderProject( final LogicalOpProject parentProjectOp,
+	                                                         final UnaryLogicalOp childOp,
+	                                                         final LogicalPlan subPlanUnderChildOp,
+	                                                         final LogicalPlan inputPlan ) {
+		// The current implementation does not try to push project operators
+		// conditions under the vocabulary rewriting operators,
+		// LogicalOpLocalToGlobal and LogicalOpGlobalToLocal.
```
hartig: Please change that. In contrast to filter operators, it is always possible to push project operators under and l2g or a g2l operator. It is not even necessary to consider any conditions here; instead, the project operator and the l2g/g2l operator can simply be swapped.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+	}
+
+	/**
+	 * Assumes that the given child operator is either
+	 * a {@link LogicalOpTPAdd},
+	 * a {@link LogicalOpTPOptAdd},
+	 * a {@link LogicalOpBGPAdd},
+	 * a {@link LogicalOpBGPOptAdd},
+	 * a {@link LogicalOpGPAdd}, or
+	 * a {@link LogicalOpGPOptAdd}.
```
hartig:
```suggestion
	 * a {@link LogicalOpGPAdd} or
	 * a {@link LogicalOpGPOptAdd}.
```
We don't have these other types anymore.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+		// currently, as such operators do not appear in the logical plans we optimize.
+		// This optimization can be revisited in the future, taking into account that
+		// for optional patterns (xxOptAdd), pushing projections may not always be safe.
+
+		final ExpectedVariables expVarsInSubPlan = subPlanUnderChildOp.getExpectedVariables();
+		final Set<Var> certainVarsInSubPlan = expVarsInSubPlan.getCertainVariables();
+		final Set<Var> varsInProject = parentProjectOp.getVariables();
+
+		// Can only push if all projected variables are guaranteed by subplan
+		if ( ! certainVarsInSubPlan.containsAll(varsInProject) ) {
```
hartig: I don't think that's right. If the expected variables from the subplan include a variable that is possible (not certain) and this variable occurs also in the graph pattern of the gpAdd operator, then this variable will be a join variable and we cannot simply remove it from the solution mappings that are produced by the subplan as input to the gpAdd.

Orthogonal to this issue (which needs to be fixed), I have another idea: Similar to the idea mentioned for bind and unfold operators, the may be cases in which a project operator on top of gpAdd or gpOptAdd may be duplicated and the duplicate be pushed.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,595 @@
...
+			return new LogicalPlanWithUnaryRootImpl(projectOp, null, subPlanAfterProjectPushDown);
+		}
+
+		final Set<Var> varsInProject = projectOp.getVariables();
+		final LogicalPlan[] newSubPlans = new LogicalPlan[numberOfSubPlansUnderJoin];
+
+		// For each subplan under the join, compute the intersection between the
+		// projected variables and the variables produced by that subplan.
+		// If the intersection is non-empty, push a project operator with those
+		// variables into that subplan.
```
hartig: That's not sufficient. In addition to the variables from the intersection, the pushed project operator also needs to keep all join variables of the subplan. This is essentially the same issue as for the gpAdd/gpOptAdd case.

hartig (changes requested):
Here is a first set of comments. For some cases, the conditions that you consider for the pushing to be done are not correct. I didn't look through all the cases, but I would assume that some of the other cases have the same kinds of issues.

Perhaps these issues are another topic that is easier to discuss in person.

kentthang010:
I think I covered the cases that you've mentioned so far. However, I left out the special case where the bind operator gets dropped if the bind variable is not in the project operator. Should that be covered in this PR or in the future project-related issue that you briefly mentioned?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithBinaryRootImpl;
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithNaryRootImpl;
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithUnaryRootImpl;
+import se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicForLogicalOptimization;
+
+/**
+ * Pushes project operators as much as possible towards the leaf nodes
+ * of a given logical plan. Whether and how a particular project operator
+ * can be pushed depends on the variables that are required and produced
+ * by the operators in the plan.
```
hartig: Adding a sentence that makes the top-down nature of the approach explicit.

```suggestion
 * Pushes project operators as much as possible towards the leaf nodes
 * of a given logical plan. This implementation works recursively in a
 * top-down manner such that, after pushing a project operator under
 * some other operator, the pushed project is considered again for
 * further pushing. Whether and how a particular project operator
 * can be pushed depends on the variables that are required and
 * produced by the operators in the plan.
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+		}
+	}
+
+	protected LogicalPlan applyToPlanWithProjectAsRootOperator( final LogicalOpProject projectOp,
+	                                                            final LogicalPlan subPlanUnderProject,
+	                                                            final LogicalPlan inputPlan ) {
+		final LogicalOperator childOpUnderProject = subPlanUnderProject.getRootOperator();
+		final Worker worker = new Worker( projectOp, subPlanUnderProject, inputPlan );
+
+		childOpUnderProject.visit(worker);
```
hartig: Some simplification:

```suggestion
		// Visit the operator directly under the project operator using
		// a worked that attempts to create a new plan by pushing the
		// project operator.
		final Worker worker = new Worker(projectOp, subPlanUnderProject, inputPlan);
		subPlanUnderProject.getRootOperator().visit(worker);

```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+		final LogicalPlan createdPlan = worker.getCreatedPlan();
+		if ( createdPlan != null ) return createdPlan;
+		else
+		{
+			throw new IllegalArgumentException( childOpUnderProject.getClass().getName() );
+		}
```
hartig:
```suggestion
		if ( createdPlan != null ) return createdPlan;

		// We may end up here only if there is a bug in the worker.
		throw new IllegalArgumentException( childOpUnderProject.getClass().getName() );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+		public void visit( final LogicalOpRequest<?, ?> op ) {
+			createdPlan = createPlanForRequestUnderProject( projectOp,
+			                                                op,
+			                                                inputPlan );
+		}
+
+		@Override
+		public void visit( final LogicalOpFixedSolMap op ) {
+			// The project cannot be pushed under this operator.
+			createdPlan = inputPlan;
```
hartig: While the statement of the comment is true, it is still possible to apply the projection directly to the solution mapping of the `LogicalOpFixedSolMap`. As a result, the project operator can be dropped altogether. That is, you create a new `LogicalOpFixedSolMap` with a solution mapping that contains only the variables that have been both in the original solution mapping and in the variables of the project operator.

There is one special case: If the set of variables in the solution mapping is the same as the set of the project operator, then you don't need to create a new `LogicalOpFixedSolMap`; instead, you simply drop the project operator (essentially, pulling up the `LogicalOpFixedSolMap` as is).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+			                                             subPlanUnderProject,
+			                                             inputPlan );
+		}
+
+		@Override
+		public void visit( final LogicalOpLeftJoin op ) {
+			createdPlan = createPlanForLeftJoinUnderProject( projectOp,
+			                                                 subPlanUnderProject.getSubPlan(1), // non-optional subplan
+			                                                 subPlanUnderProject.getSubPlan(0), // optional subplan
+			                                                 inputPlan );
```
hartig: That's incorrect. The non-optional part is the first subplan, and the second subplan is the optional part.
```suggestion
			createdPlan = createPlanForLeftJoinUnderProject( projectOp,
			                                                 subPlanUnderProject.getSubPlan(0), // non-optional subplan
			                                                 subPlanUnderProject.getSubPlan(1), // optional subplan
			                                                 inputPlan );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+			                                                op,
+			                                                subPlanUnderProject.getSubPlan(0) );
+		}
+
+		@Override
+		public void visit( final LogicalOpDedup op ) {
+			createdPlan = createPlanForL2GOrG2LUnderProject( projectOp,
+			                                                op,
+			                                                subPlanUnderProject.getSubPlan(0) );
+		}
```
hartig, line 249: Calling `createPlanForL2GOrG2LUnderProject` here is incorrect.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+		// - Variables assigned by the bind operator are removed, since they
+		//   are produced by the bind and not needed as input.
+		// - Variables required to evaluate the bind expressions are added,
+		//   to ensure the bind can still be evaluated after pushdown.
+		//
+		// If this adjustment does not change the projection, the whole project
+		// can be pushed below the bind.
+		//
+		// Otherwise, the project is split: the adjusted projection is pushed below
+		// the bind, while the original projection remains on top.
```
hartig: There is one more condition: If the adjusted projection ends up containing all the variables of the subplan under the bind (including both the certain and the possible variables), then there is no need to push the adjusted projection at all and, thus, the plan can remain as is. (The same applies to the next method about the unfold operator.)

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+		}
+
+		// Compute pushed vars
+		final Set<Var> pushedProjectVars = new HashSet<>(projectOp.getVariables());
+
+		// Remove vars produced by bind
+		pushedProjectVars.removeAll(bindVars);
+
+		// Add vars needed for evaluating bind
+		pushedProjectVars.addAll(varsUsedInBind);
```
hartig: Initializing `bindVars` way before it is actually used makes the code harder to read than necessary. So, you should move it to right before the `removeAll` statement. But then, it is also not necessary to actually create a new `HashSet` for it and copy all the bind variables into that `HashSet`. Instead, you can simply do the `removeAll` directly with the variables from the bind expressions. 

Another simplification is to avoid populating `varsUsedInBind`. Instead, you can simply do the `pushedProjectVars.addAll(..)` directly within the for-loop.

In summary, you can simplify as follows.
```suggestion
		// Compute pushed variables, starting with all the projection variables.
		final Set<Var> pushedProjectVars = new HashSet<>( projectOp.getVariables() );

		// Remove vars produced by bind
		pushedProjectVars.removeAll( bindOp.getBindExpressions().getVars() );

		// Add vars needed for evaluating bind
		for ( final Expr expr : bindOp.getBindExpressions().getExprs().values() ) {
			pushedProjectVars.addAll( ExprVars.getVarsMentioned(expr) );
		}
```

hartig: The same simplification can be applied in the next method about unfold.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,756 @@
...
+
+		// If nothing changes, push whole project
+		if ( pushedProjectVars.equals(projectOp.getVariables()) )
+			return createPlanForUnaryOpUnderProject(projectOp, bindOp, subPlanUnderBind);
+
+		// Otherwise, split
+		final LogicalPlan pushed = new LogicalPlanWithUnaryRootImpl(
+			new LogicalOpProject(pushedProjectVars, false),
```
hartig: Why `false`? It should be the `mayReduce` flag of the original project operator (or of the bind operator, which should be the same).

hartig (changes requested):
I found some time to come back to this PR. There are several new comments and requests for changes.

So far, I have managed to get until (and including) `createPlanForBindUnderProject`. Yet, some comments may also give you an idea for additional changes that you may have to do in the rest of the file.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
KentThang: Ah, should this same change be made in `FilterPushDown`? I copied the code from there and used it in `ProjectPushDown`.
`FilterPushDown`:
```
@Override
		public void visit( final LogicalOpLeftJoin op ) {
			createdPlan = createPlanForLeftJoinUnderFilter( filterOp,
			                                                subPlanUnderFilter.getSubPlan(1), // non-optional subplan
			                                                subPlanUnderFilter.getSubPlan(0), // optional subplan
			                                                inputPlan );
		}
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
KentThang: I changed it to createPlanForUnaryOpUnderProject which I think is correct? Is there a special case here which makes it unsafe?

kentthang010:
Latest commit resolves PR comments above but comments with questions were left unresolved (in GitHub, but addressed in code!).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+	                                                        final LogicalPlan inputPlan ) {
+		// Pushing a project operator into request operator is not possible yet TODO #570
+		return inputPlan;
+	}
+
+	protected LogicalPlan createPlanForFixedSolMapUnderProject( final LogicalOpProject projectOp,
+	                                                            final LogicalOpFixedSolMap fixedSolMapOp,
+	                                                            final LogicalPlan inputPlan ) {
+		// Apply the projection directly to the solution mappings
+		// produced by the fixed solution mapping operator.
```
hartig:
```suggestion
		// Apply the projection directly to the solution mapping
		// of the fixed solution mapping operator.
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		final Set<Var> projectedVars = projectOp.getVariables();
+		final Binding b = fixedSolMapOp.getSolutionMapping().asJenaBinding();
+		final Set<Var> bindingVars = b.varsMentioned();
+		// Special case: If set of variables in solution mapping
+		// equals set of projection variables; then simply drop
+		// the project operator.
+		if ( bindingVars.equals(projectedVars) ) {
```
hartig: I realize now that this special case can be made more general: It is not necessary that the two sets of variables must be equal. Instead, it is sufficient, if `bindingVars` is a subset of `projectedVars`. I believe it should be possible to check this using `projectedVars.containsAll(bindingVars)` or something (but please also adapt the corresponding comment).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
hartig: Yes!!

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
hartig: > I changed it to createPlanForUnaryOpUnderProject which I think is correct? Is there a special case here which makes it unsafe?

Good observation. Yes, that's the correct rewriting for this case. No special cases need to be considered.

I will mark this point as resolved.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		public void visit( final LogicalOpProject op ) {
+			createdPlan = createPlanForProjectUnderProject( projectOp,
+			                                                op,
+			                                                subPlanUnderProject.getSubPlan(0) );
+		}
+
+		@Override
+		public void visit( final LogicalOpMinus op ) {
+			//unimplemented
+		}
```
hartig, line 263: What about this one?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+
+		// If nothing changes, push whole project
+		if ( pushedProjectVars.equals(projectOp.getVariables()) )
+			return createPlanForUnaryOpUnderProject(projectOp, bindOp, subPlanUnderBind);
+
+		// If adjusted projection contains all variables of the subplan under bind,
+		// return the plan as is
+		final Set<Var> subPlanVars = new HashSet<>(subPlanUnderBind.getExpectedVariables().getCertainVariables());
+		subPlanVars.addAll( subPlanUnderBind.getExpectedVariables().getPossibleVariables() );
+		if ( pushedProjectVars.containsAll(subPlanVars) ) {
```
hartig: To reduce the memory footprint of the engine, I usually try to avoid creating short-lived Java objects if not really necessary. The `HashSet` here is a case where it is indeed unnecessary.

```suggestion
		final ExpectedVariables expVars = subPlanUnderBind.getExpectedVariables();
		if (    pushedProjectVars.containsAll(expVars.getCertainVariables())
		     && pushedProjectVars.containsAll(expVars.getPossibleVariables()) ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		// If the adjusted projection contains all the variables of the subplan
+		// under the bind (including both the certain and the possible variables),
+		// then there is no need to push the adjusted projection at all and the
+		// plan can remain as is.
+		//
+		// Otherwise, the project is split: the adjusted projection is pushed below
+		// the bind, while the original projection remains on top.
+
+		// Compute pushed vars
+		final Set<Var> pushedProjectVars = new HashSet<>(projectOp.getVariables());
```
hartig:
```suggestion
		// Determine the projection variables to be pushed, starting with all the projection variables.
		final Set<Var> pushedProjectVars = new HashSet<>( projectOp.getVariables() );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		// If the adjusted projection contains all the variables of the subplan
+		// under the bind (including both the certain and the possible variables),
+		// then there is no need to push the adjusted projection at all and the
+		// plan can remain as is.
+		//
+		// Otherwise, the project is split: the adjusted projection is pushed below
+		// the unfold, while the original projection remains on top.
+
+		// Compute pushed vars
+		final Set<Var> pushedProjectVars = new HashSet<>(projectOp.getVariables());
```
hartig:
```suggestion
		// Determine the projection variables to be pushed, starting with all the projection variables.
		final Set<Var> pushedProjectVars = new HashSet<>( projectOp.getVariables() );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		}
+
+		// Add vars needed for evaluating unfold
+		pushedProjectVars.addAll( ExprVars.getVarsMentioned(unfoldOp.getExpr()) );
+
+		// If nothing changes, push whole project
+		if ( pushedProjectVars.equals(projectOp.getVariables()) )
+			return createPlanForUnaryOpUnderProject(projectOp, unfoldOp, subPlanUnderUnfold);
+
+		// If adjusted projection contains all variables of the subplan under bind,
```
hartig:
```suggestion
		// If adjusted projection contains all variables of the subplan under the unfold operator,
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		// return the plan as is
+		final Set<Var> subPlanVars = new HashSet<>(subPlanUnderUnfold.getExpectedVariables().getCertainVariables());
+		subPlanVars.addAll( subPlanUnderUnfold.getExpectedVariables().getPossibleVariables() );
+		if ( pushedProjectVars.containsAll(subPlanVars) ) {
```
hartig:
```suggestion
		final ExpectedVariables expVars = subPlanUnderUnfold.getExpectedVariables();
		if (    pushedProjectVars.containsAll(expVars.getCertainVariables())
		     && pushedProjectVars.containsAll(expVars.getPossibleVariables()) ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+			return inputPlan;
+		}
+
+		// Otherwise, split
+		final LogicalPlan pushed = new LogicalPlanWithUnaryRootImpl(
+			new LogicalOpProject(pushedProjectVars, false),
```
hartig:
```suggestion
			new LogicalOpProject(pushedProjectVars, projectOp.mayReduce()),
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+	/**
+	 * Assumes that the given child operator is either a {@link LogicalOpLocalToGlobal}
+	 * or a {@link LogicalOpGlobalToLocal}.
+	 */
+	protected LogicalPlan createPlanForL2GOrG2LUnderProject( final LogicalOpProject projectOp,
+	                                                        final UnaryLogicalOp childOp,
+	                                                        final LogicalPlan subPlanUnderChildOp ) {
+		// Swaps a project operator with a LocalToGlobal or GlobalToLocal operator.
+		// There are no variable or semantic constraints that restrict pushing
+		// a project operator past L2G/G2L operators. The result is a reordered
```
hartig: Unfortunately, it's not so easy. At least one of the two operators (I don't remember which one, or maybe it is even both) may translate an input solution mapping into multiple output solution mappings. If the difference between these output solution mappings is only in terms of the values that they have for one of the variables that are projected away, then pushing the project operator may have an impact on the number of duplicate solution mappings that come out of the subplan. In fact, if I remember correctly, the translation performed by l2g or by g2l may even result in dropping some of the input solution mappings, which may be _incorrectly_ prevented if the project operator is pushed. Therefore, for the time being, it is better not to push at all for these operators. Therefore, please change this method here such that, the only thing it does is to recursively call `apply` on the `subPlanUnderChildOp` (because there may be other project operators somewhere inside that subplan). Additionally, place the following comment into this method:

> While there may be cases in which pushing the project operator under an l2g or a g2l operator is safe, there are certainly cases in which it is not safe: At least one of the two operators (perhaps even both?) may translate an input solution mapping into multiple output solution mappings. If the difference between these output solution mappings is only in terms of the values that they have for one of the variables that are projected away, then pushing the project operator may have an impact on the number of duplicate solution mappings that come out of the subplan. Additionally, if I remember correctly, the translation performed by l2g or by g2l may even result in dropping some of the input solution mappings, which may be *incorrectly* prevented if the project operator is pushed. TODO: think more about cases in which pushing a project operator is safe.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		final Set<Var> certainVarsInSubPlan = subPlanUnderChildOp.getExpectedVariables().getCertainVariables();
+
+		// Case 1:
+		// We can push the full projection below the GPAdd/GPOptAdd operator
+		// only if:
+		//	(1) the subplan guarantees all projected variables, and
+		//	(2) the operator does not require any extra variables
+		//	  (e.g., parameter variables would be lost by projection)
+		if ( certainVarsInSubPlan.containsAll(parentProjectOp.getVariables())
+		  && operatorDoesNotRequireExtraVars(childOp, parentProjectOp) ) {
```
hartig: But this is still incorrect in case the (pushed) project operator would project away some join variables. So, if the variables of the project operator do not include _all_ join variables (which is the intersection of the expected variables from the subplan and the expected variables from the pattern of the gpAdd/gpOptAdd operator), then it is not possible to simply push the project operator as is. Instead, in this case, it is possible split: keep the project operator on top but make a copy with all join variables added and push that copy; yet, if that copy ends up containing all of the variables of the subplan, then there is no need to add it.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		// - variables required for join alignment (joinVars)
+		// - variables required by the parent projection
+
+		final int numberOfSubPlansUnderJoin = subPlanUnderProject.numberOfSubPlans();
+
+		// If there is only one subplan under the join, then remove the join altogether.
+		if ( numberOfSubPlansUnderJoin == 1 ) {
+			final LogicalPlan subPlan = subPlanUnderProject.getSubPlan(0);
+			final LogicalPlan subPlanAfterProjectPushDown = apply(subPlan);
+			return new LogicalPlanWithUnaryRootImpl(projectOp, null, subPlanAfterProjectPushDown);
```
hartig: The order of these steps is wrong. If you do `apply(subPlan)` first, there will never be an attempt to push `projectOp` further into `subPlan`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+	protected Set<Var> computeJoinVars( final LogicalPlan joinPlan ) {
+		final Map<Var, Integer> seenOnce = new HashMap<>();
+		final Set<Var> joinVars = new HashSet<>();
+
+		for ( int i = 0; i < joinPlan.numberOfSubPlans(); i++ ) {
+			final LogicalPlan sp = joinPlan.getSubPlan(i);
+
+			final Set<Var> vars = new HashSet<>();
+			vars.addAll(sp.getExpectedVariables().getCertainVariables());
+			vars.addAll(sp.getExpectedVariables().getPossibleVariables());
```
hartig: Instead of creating this extra Java object, can you please do versions of the for-loop below, one for the certain variables and the other one for the possible variables.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+				newPlanUnderChildRoot );
+		final LogicalPlan newPlan = new LogicalPlanWithUnaryRootImpl(
+				parentProjectOp,
+				null,
+				newPlanUnderProject );
+		return newPlan;
+	}
+
+	protected Set<Var> computeJoinVars( final LogicalPlan joinPlan ) {
+		final Map<Var, Integer> seenOnce = new HashMap<>();
```
hartig: Since you are not actually using this as a map, make this a set.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+		// - variables required by the top-level projection
+		// - variables required to preserve join correctness in this branch
+		//
+		// The union of these variables is pushed down as a projection into the branch.
+		boolean noChanges = true; // set to false if something has changed in the plan
+		for (int i = 0; i < numberOfSubPlansUnderJoin; i++) {
+			final LogicalPlan subPlan = subPlanUnderProject.getSubPlan(i);
+
+			final Set<Var> varsInSubPlan = new HashSet<>(subPlan.getExpectedVariables().getCertainVariables());
+			varsInSubPlan.addAll( subPlan.getExpectedVariables().getPossibleVariables() );
```
hartig: As a shortcut, use:
```
			final Set<Var> varsInSubPlan = ExpectedVariablesUtils.unionOfAllVariables( subPlan.getExpectedVariables() );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+				newSubPlan = apply(subPlan);
+			} else {
+				final LogicalOpProject branchProject = new LogicalOpProject(pushDownVars, projectOp.mayReduce());
+				final LogicalPlan withProject = new LogicalPlanWithUnaryRootImpl(branchProject, null, subPlan);
+
+				newSubPlan = apply(withProject);
+			}
+
+			newSubPlans[i] = newSubPlan;
+			if ( !newSubPlan.equals(subPlan) )
```
hartig: Better use `isSamePlan` instead of `equals`

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+			final Set<Var> pushDownVars = new HashSet<>();
+			pushDownVars.addAll( neededByProject );
+			pushDownVars.addAll( neededByJoin );
+
+			final LogicalPlan newSubPlan;
+
+			if ( pushDownVars.isEmpty() ) {
+				// No projection needed for this branch
+				newSubPlan = apply(subPlan);
+			} else {
```
hartig: Another case in which no projection is needed is if `pushDownVars.equals(varsInSubPlan)`

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,817 @@
...
+			return inputPlan;
+
+		final LogicalOperator joinOp = subPlanUnderProject.getRootOperator();
+
+		final LogicalPlan newJoin = LogicalPlanUtils.createPlanWithSubPlans(
+				joinOp,
+				null,
+				newSubPlans );
+
+		// Keep original project on top
```
hartig: There is a case in which the original project operator can even be dropped: If the set of variables of that project operator is equal to the union of all the `pushDownVars` sets from the for-loop.

hartig (changes requested):
Here is the next round of comments. This time I managed to get to `createPlanForJoinUnderProject`.

One more request: most of these `createPlanFor...` methods have some comment in the beginning, describing what the method does. Please turn these comments in Javadoc comments for these methods.

kentthang010:
Latest commit resolves PR comments above and moves all comments within the `createPlanFor` methods into Javadoc comments.

I was not certain whether it is correct for Join, LeftJoin, Minus _**and**_ MultiwayJoin to all use the new helper method `createPlanForNaryOpUnderProject` but I opted to do so for the time being.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		newSubPlansUnderJoin[0] = newNonOptSubPlan;
+
+		// Now set up the new optional subplans of the join by
+		// applying this heuristic recursively to each of them.
+		boolean noChanges = true; // set to false if something has changed in the plan
+		for ( int i = 1; i < numberOfSubPlansUnderJoin; i++ ) {
+			final LogicalPlan oldSubPlan = subPlanUnderProject.getSubPlan(i);
+			final LogicalPlan newSubPlan = apply(oldSubPlan);
+
+			if ( newSubPlan.equals(oldSubPlan) ) {
```
hartig:
```suggestion
			if ( newSubPlan.isSamePlan(oldSubPlan) ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		}
+		else {
+			// For any other type of root operator, simply apply
+			// this heuristic recursively to all of the subplans.
+			final LogicalPlan[] newSubPlans = new LogicalPlan[numberOfSubPlans];
+			boolean noChanges = true; // set to false if the heuristic changes any of the subplans
+			for ( int i = 0; i < numberOfSubPlans; i++ ) {
+				final LogicalPlan oldSubPlan = inputPlan.getSubPlan(i);
+				newSubPlans[i] = apply(oldSubPlan);
+				if ( ! newSubPlans[i].equals(oldSubPlan) ) {
```
hartig:
```suggestion
				if ( ! newSubPlans[i].isSamePlan(oldSubPlan) ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+	 * Pushes a projection below a n-ary operator by splitting required variables per branch.
+	 *
+	 * Each branch must retain:
+	 * <ul>
+	 *	<li>variables required by the parent projection</li>
+	 *	<li>variables required to preserve operator semantics (e.g., join compatibility,
+	 *      or variables required by asymmetric operators such as left join or minus)</li>
+	 * </ul>
+	 */
+	protected LogicalPlan createPlanForNaryOpUnderProject( final LogicalOpProject projectOp,
```
hartig: Please rename this method to `createPlanForJoinLikeOpUnderProject`

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+	 * TODO: think more about cases in which pushing a project operator is safe.
+	 */
+	protected LogicalPlan createPlanForL2GOrG2LUnderProject( final LogicalOpProject projectOp,
+	                                                        final UnaryLogicalOp childOp,
+	                                                        final LogicalPlan subPlanUnderChildOp ) {
+		return apply(subPlanUnderChildOp);
+	}
+
+	/**
+	 * Pushes a project operator below a GPAdd or GPOptAdd operator.
```
hartig:
```suggestion
	 * Pushes a project operator below a gpAdd or gpOptAdd operator.
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+	 *
+	 * The pushdown must preserve:
+	 * <ul>
+	 * 	<li>variables required by the parent projection</li>
+	 * 	<li>variables required by the operator (e.g., parameter variables or optional bindings)</li>
+	 * 	<li>join variables between the subplan and the operator pattern
```
hartig:
```suggestion
	 * 	<li>join variables between the subplan and the operator pattern</li>
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+			  || possibleVarsInSubPlan.contains(v) )
+				joinVars.add(v);
+		}
+
+		// intersect with possible vars
+		for ( final Var v : patternPossible ) {
+			if ( certainVarsInSubPlan.contains(v)
+			  || possibleVarsInSubPlan.contains(v) )
+				joinVars.add(v);
+		}
```
hartig: You can use `ExpectedVariablesUtils.intersectionOfAllVariables(..)` to determine the `joinVars`

```suggestion
		final ExpectedVariables expVarsInSubPlan = subPlanUnderChildOp.getExpectedVariables();

		final ExpectedVariables expVarsInPattern;
		if ( childOp instanceof LogicalOpGPAdd gpAdd ) {
			expVarsInPattern = gpAdd.getPattern().getExpectedVariables();
		}
		else {
			final LogicalOpGPOptAdd op = (LogicalOpGPOptAdd) childOp;
			expVarsInPattern = op.getPattern().getExpectedVariables();
		}

		final Set<Var> joinVars = ExpectedVariablesUtils.intersectionOfAllVariables(
				expVarsInPattern,
				expVarsInSubPlan );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+
+		// Case 1:
+		// We can push the full projection below the GPAdd/GPOptAdd operator
```
hartig:
```suggestion
		// We can push the full projection below the gpAdd/gpOptAdd operator
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		// only if:
+		//	(1) the subplan guarantees all projected variables,
+		//	(2) the operator does not require any extra variables
+		//	  (e.g., parameter variables would be lost by projection), and
+		//	(3) the projected variables include all join variables
+		if ( certainVarsInSubPlan.containsAll(parentProjectOp.getVariables())
```
hartig:
```suggestion
		if ( expVarsInSubPlan.getCertainVariables().containsAll(parentProjectOp.getVariables())
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		// Case 2:
+		// We attempt a partial pushdown of the projection.
+		// We push down only:
+		//	(1) variables from the projection that are needed by the subplan output
+		//	(2) variables required by the operator (e.g., GPAdd parameters, optional bindings)
+		//	(3) join variables
+		//
+		// The remaining projection is kept above the operator.
+		final Set<Var> varsInSubPlan = new HashSet<>(certainVarsInSubPlan);
+		varsInSubPlan.addAll( possibleVarsInSubPlan );
```
hartig:
```suggestion
		final Set<Var> varsInSubPlan = ExpectedVariablesUtils.unionOfAllVariables(expVarsInSubPlan);
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		final Set<Var> neededByProject = new HashSet<>(parentProjectOp.getVariables());
+		neededByProject.retainAll( varsInSubPlan );
+
+		final Set<Var> neededByOperator = new HashSet<>();
+		if ( childOp instanceof LogicalOpGPAdd gpAdd && gpAdd.hasParameterVariables() ) {
+			// GPAdd parameter variables must be preserved by the projection.
+			// If they are not present in the outer projection, we must abort pushdown.
+			final Set<Var> paramVars = new HashSet<>(gpAdd.getParameterVariables().values());
```
hartig: No need to create set and copy the variables over.
```suggestion
			final Collection<Var> paramVars = gpAdd.getParameterVariables().values();
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+
+			if ( ! parentProjectOp.getVariables().containsAll(paramVars) )
+				return inputPlan; // cannot push at all
```
hartig: Why not?? Isn't it the whole point of `neededByOperator` to make sure that all the parameter variables get pushed?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+
+			neededByOperator.addAll( paramVars );
+		}
+		else if ( childOp instanceof LogicalOpGPOptAdd ) {
+			neededByOperator.addAll( possibleVarsInSubPlan );
+		}
```
hartig: Why?
It is a genuine question. I don't see why a gpOpAdd should need (all and only) the possible variables from the subplan. But maybe I'm overlooking something?
I would think that gpOptAdd needs the join variables, but these you cover elsewhere anyways.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		neededByOperator.retainAll( varsInSubPlan );
+
+		final Set<Var> pushDownVars = new HashSet<>();
+		pushDownVars.addAll( neededByProject );
+		pushDownVars.addAll( neededByOperator );
+		pushDownVars.addAll( joinVars );
+
+		if ( pushDownVars.isEmpty() || pushDownVars.equals(varsInSubPlan) ) {
+			// No projection needed for this branch
+			return inputPlan;
```
hartig: Shouldn't we recurse here, applying the heuristic to the subplan? Or is this done elsewhere?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		}
+
+		final LogicalPlan pushed = new LogicalPlanWithUnaryRootImpl(
+			new LogicalOpProject(pushDownVars, false),
```
hartig: Why `false`?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+	}
+
+	/**
+	 * Pushes a projection below a n-ary operator by splitting required variables per branch.
+	 *
+	 * Each branch must retain:
+	 * <ul>
+	 *	<li>variables required by the parent projection</li>
+	 *	<li>variables required to preserve operator semantics (e.g., join compatibility,
+	 *      or variables required by asymmetric operators such as left join or minus)</li>
```
hartig:
```suggestion
	 *	<li>its subset of the variables of the parent projection</li>
	 *	<li>its subset of the join variables</li>
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+			final Set<Var> neededByProject = new HashSet<>(projectOp.getVariables());
+			neededByProject.retainAll( varsInSubPlan );
+
+			final Set<Var> neededByJoin = new HashSet<>(joinVars);
+			neededByJoin.retainAll( varsInSubPlan );
+
+			final Set<Var> pushDownVars = new HashSet<>();
+			pushDownVars.addAll( neededByProject );
+			pushDownVars.addAll( neededByJoin );
+			allPushdownVars.addAll( pushDownVars );
```
hartig:
```suggestion
			final Set<Var> varsInSubPlan = ExpectedVariablesUtils.unionOfAllVariables( subPlan.getExpectedVariables() );

			final Set<Var> pushDownVars = new HashSet<>();
			pushDownVars.addAll( projectOp.getVariables() );
			pushDownVars.addAll( joinVars );
			pushDownVars.retainAll( varsInSubPlan );

			allPushdownVars.addAll( pushDownVars );
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+				newSubPlan = apply(subPlan);
+			} else {
+				final LogicalOpProject branchProject = new LogicalOpProject(pushDownVars, projectOp.mayReduce());
+				final LogicalPlan withProject = new LogicalPlanWithUnaryRootImpl(branchProject, null, subPlan);
+
+				newSubPlan = apply(withProject);
+			}
+
+			newSubPlans[i] = newSubPlan;
+			if ( !newSubPlan.isSamePlan(subPlan) )
```
hartig:
```suggestion
			if ( pushDownVars.isEmpty() || pushDownVars.equals(varsInSubPlan) ) {
				// No projection needed for this branch
				newSubPlans[i] = apply(subPlan);
			}
			else {
				final LogicalOpProject branchProject = new LogicalOpProject(pushDownVars, projectOp.mayReduce());
				final LogicalPlan withProject = new LogicalPlanWithUnaryRootImpl(branchProject, null, subPlan);

				newSubPlans[i] = apply(withProject);
			}

			if ( !newSubPlans[i].isSamePlan(subPlan) )
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+		return new LogicalPlanWithUnaryRootImpl(projectOp, null, newJoin);
+	}
+
+	/**
+	 * Only pushes the project operator into the non-optional subplan.
+	 *
+	 * For the optional subplans, we only apply the heuristic recursively,
+	 * but do not push the project operator into them.
+	 */
+	protected LogicalPlan createPlanForMultiwayLeftJoinUnderProject( final LogicalOpProject projectOp,
```
hartig: The implementation of this method doesn't seem right. For the multiway left join, we can also use `createPlanForJoinLikeOpUnderProject`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+
+		return apply(newPlan);
+	}
+
+
+
+	protected LogicalPlan createPlanForUnaryOpUnderProject( final LogicalOpProject projectOp, final UnaryLogicalOp op, final LogicalPlan subPlanUnderOp ) {
+		// Create a new subplan with the project as root operator on top of the
+		// subplan that was under the given unary operator, and apply this
+		// heuristic recursively to this new subplan.
```
hartig:
```suggestion
	/**
	 * Returns a plan in which the given project operator is pushed under the
	 * given unary operator, with the subplan underneath being a version of
	 * the given subplan in which the project push down heuristic has been
	 * applied recursively.
	 */
	protected LogicalPlan createPlanForUnaryOpUnderProject( final LogicalOpProject projectOp,
	                                                        final UnaryLogicalOp op,
	                                                        final LogicalPlan subPlanUnderOp ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+				projectOp,
+				null,
+				subPlanUnderOp );
+		final LogicalPlan newSubPlan2 = apply(newSubPlan1);
+
+		// Put together the new plan with the given unary operator as root.
+		return LogicalPlanUtils.createPlanWithSubPlans(op, null, newSubPlan2);
+	}
+
+	protected LogicalPlan createPlanAfterPushingInSubPlan( final LogicalOpProject parentProjectOp,
```
hartig: Please add a Javadoc comment that describes what this method returns, similar to the Javadoc comment that I am proposing in my edit suggestion for the previous method.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+	                                                       final LogicalPlan subPlanUnderChildOp,
+	                                                       final LogicalPlan inputPlan ) {
+		// Apply the heuristic recursively within the given subplan.
+		final LogicalPlan newPlanUnderChildRoot = apply(subPlanUnderChildOp);
+
+		// If the heuristic cannot be applied within the subplan (more
+		// precisely, if there are no project operators in the subplan that can
+		// be pushed), then we return the given input plan (instead of
+		// recreating another, identical version of that plan).
+		if ( newPlanUnderChildRoot.equals(subPlanUnderChildOp) ) {
```
hartig:
```suggestion
		if ( newPlanUnderChildRoot.isSamePlan(subPlanUnderChildOp) ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -0,0 +1,856 @@
...
+				null,
+				newPlanUnderChildRoot );
+		final LogicalPlan newPlan = new LogicalPlanWithUnaryRootImpl(
+				parentProjectOp,
+				null,
+				newPlanUnderProject );
+		return newPlan;
+	}
+
+	protected Set<Var> computeJoinVars( final LogicalPlan joinPlan ) {
```
hartig, line 723: Please add a Javadoc comment for this one as well, something like:

```java
/**
 * Determines the set of all join variables between the subplans of the
 * given plan, where a join variable is ...
```

hartig (changes requested):
We are getting there. I have the feeling that this is the last review with change requests ;-)

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
KentThang: While writing the comment for this method, I noticed this method is actually unused. It was copied over from `FilterPushDown` and it was used at some point here in `ProjectPushDown` but not after some commits. So I am removing the method altogether!

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
KentThang: I believe I thought something like, "since gpOptAdd is optional, we might to preserve more variables to avoid losing information"; nothing that makes sense to me now looking at it. I have removed that if else statement.

hartig (approved):
Now it can be merged! :-)

PR #584, opened 2026-04-15
Add rewriting to drop subplans with empty results

### LINKED ISSUE #563: Rewriting to drop subplans with empty results
We should add a rewriting rule for logical plans that recursively removes subplans that are guaranteed to produce an empty result. The implementation of this rule can be based on the cardinality estimator: a subplan can be removed if it has a cardinality estimate of 0 with 'accurate' as its quality score.

It may be the case that the complete query plan can be "removed", in which case HeFQUIN can stop the processing immediately and return the empty result without even executing any plan.

In this context it is important to double check that the cardinality estimator is both correct and exhaustive especially for cases in which it is possible to determine that a subplan produces an empty result.

The rule can be applied as the first one because it may reduce the work of other rules (or make them even completely obsolete).

kentthang010:
Closes #563.


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,120 @@
...
+import se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicForLogicalOptimization;
+
+/**
+ * Recursively removes subplans that are guaranteed to produce empty results.
+ * A subplan is considered empty only if it has a cardinality estimate of 0
+ * with ACCURATE quality.
+ *
+ * The heuristic first ensures that all subplans are annotated with
+ * cardinality estimates. It then traverses the plan bottom-up and removes
+ * subplans that are guaranteed to be empty.
```
hartig: I think that top-down is more reasonable. If there is a sub-plan that is guaranteed to produce the empty result and, thus, will be removed by this class, why first making an effort to try to remove something somewhere within this sub-plan? Likewise, if the overall plan is guaranteed to produce the empty result and, thus, will be replaced by the `LogicalPlanWithoutResult`, why first making an effort for its sub-plans?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * For union and multiway union operators, empty subplans are removed. If all
+ * subplans are removed, the entire plan is replaced with a
+ * {@link LogicalPlanWithoutResult}. If exactly one subplan remains, the union
```
hartig: When going top down, you should never have the case mentioned in the second sentence here.

It was not the idea that `LogicalPlanWithoutResult` is placed somewhere into a bigger plan. The only use case for `LogicalPlanWithoutResult` is if the overall plan (i.e., the one given to the main `apply` function here) is known to produce the empty result, in which case `apply` should return `LogicalPlanWithoutResult` such that the query processor can pick it up and stop any further processing of the query immediately (which is the functionality that I implemented in PR #583 yesterday).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,120 @@
...
+ * operator is removed and replaced by that subplan.
+ *
+ * For left join operators: if the left subplan is empty, the entire result is empty.
```
hartig: No need to talk about the case of the left subplan being empty. This case is covered by the next paragraph and mentioning it here explicitly may be confusing. In other words, the only case to be mentioned here explicitly is the case in which the right subplan is known to produce the empty result.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,120 @@
...
+ * If the right subplan is empty, the left join is removed and replaced by its left subplan.
+ *
+ * For all other operators, empty subplans are not removed individually.
+ * Instead, emptiness is propagated via cardinality estimates, meaning that
+ * if an operator is guaranteed to produce an empty result, the entire plan
+ * rooted at that operator is replaced with a {@link LogicalPlanWithoutResult}.
```
hartig: I understand that putting `LogicalPlanWithoutResult` into the plan is something that needs to be done when going bottom up, but it is not needed when going top down, which is more efficient.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+public class RemoveSubPlansWithEmptyResults implements HeuristicForLogicalOptimization
+{
+	protected final CardinalityEstimationWorkerImpl cardEst = new CardinalityEstimationWorkerImpl();
```
hartig: Use `CardinalityEstimator` here, instead of `CardinalityEstimatorWorkerImpl`. And initialize it based on `QueryProcContext` exactly as done in `CardinalityBasedJoinOrderingBase`.

hartig (changes requested):
Some initial comments, mainly based on what you write in the Javadoc comment.

hartig:
@keski The [CI run of this PR](https://github.com/LiUSemWeb/HeFQUIN/actions/runs/24462633681/job/71480825021?pr=584) failed. The reason seem to be failures in the unit tests of `SparqlServletTest`, which is something that is not touched at all by this PR. Do you have an idea what the problem might be?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
KentThang: My bad! The actual algorithm should be top-down (since the first step after cardinality estimation is to check if the current plan is empty), it was just the description that was written incorrectly. I have rewritten it to be more accurate.

hartig: Okay. Request my review once you have pushed your commits.

keski:
Strange exception, I've not modified the servlet code and I've not seen the error before. I'm now able to replicate it so I'll begin investigating it immediately.

hartig:
Thanks @keski !
@kentthang010 don't let this stop you from continuing to work on this PR.

kentthang010:
Latest commit should resolve this first batch of PR comments. I was mistaken when I said that the algorithm uses a top-down approach but I've modified it now. 

I assume same test is failing as last time.

Note to myself mostly: Currently, one of the tests `(removeEmptySubPlansLeftJoinWithEmptyLeftBranch`) fail and is therefore ignored but closing #585 should lead to it passing.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalPlanWithoutResult;
+import se.liu.ida.hefquin.engine.queryproc.CardinalityEstimator;
+import se.liu.ida.hefquin.engine.queryproc.QueryProcContext;
+import se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator;
+import se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicForLogicalOptimization;
+
+/**
+ * Recursively removes subplans that are guaranteed to produce empty results.
+ * A subplan is considered empty only if it has a cardinality estimate of 0
+ * with ACCURATE quality.
```
hartig:
```suggestion
 * Removes every subplan that is guaranteed to produce an empty result
 * according to the cardinality estimate that is determined for it. In
 * particular, a subplan is removed if it has a cardinality estimate of
 * 0 with the quality score being ACCURATE. The subplans are removed in
 * a top-down manner (i.e., trying to remove bigger subplans first).
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+ *   If exactly one subplan remains, the union operator is removed and
+ *   replaced by that subplan.
+ *
+ * - For left join operators, if the right subplan is empty, the left join is
+ *   removed and replaced by its left subplan.
+ *
+ * For all other operators, empty subplans are not removed individually.
+ * Instead, emptiness is propagated via cardinality estimates, allowing
+ * higher-level simplifications when possible
+ *
```
hartig:
```suggestion
 * <p>
 * The heuristic performs a recursive traversal of the logical plan and simplifies
 * the plan based on emptiness information:
 * <ul>
 * <li>If the overall plan itself is provably empty, it is replaced by
 *   {@link LogicalPlanWithoutResult}.</li>
 * <li>For union and multiway union operators, empty subplans are removed.
 *   If exactly one subplan remains, the union operator is removed and
 *   replaced by that subplan.</li>
 * <li>For left join operators, if the right subplan is empty, the left join is
 *   removed and replaced by its left subplan.</li>
 * <li>For all other operators, empty subplans are not removed individually.
 * Instead, emptiness is propagated via cardinality estimates, allowing
 * higher-level simplifications when possible.</li>
 * </ul>
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+		assert cardEst != null;
+		this.cardEst = cardEst;
+	}
+
+	@Override
+	public LogicalPlan apply( final LogicalPlan inputPlan ) {
+		// Ensure all subplans have up-to-date cardinality estimates before rewriting.
+		cardEst.addCardinalities(inputPlan);
+
+		// If the entire plan is provably empty, replace it immediately.
```
hartig: The plans are not empty.
```suggestion
		// If the entire plan is guaranteed to produce the empty result,
		// replace it immediately.
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+		if ( isProvablyEmpty(inputPlan) )
+			return LogicalPlanWithoutResult.getInstance();
+
+		return _apply(inputPlan);
+	}
+
+	protected LogicalPlan _apply( final LogicalPlan inputPlan ) {
+		// Recursively rewrite all subplans before handling the current operator.
```
hartig: Why "before"?? This still sounds like a bottom-up approach.

But the issue is not only this comment here. I would like it be more obvious from the code itself that the algorithm works in a top-down fashion. To this end, move the if-block from `apply` to here, as the first thing that this recursive function does.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+			else {
+				return LogicalPlanUtils.createPlanWithSubPlans( op, null, rewrittenSubPlans );
+			}
+		}
+
+		// For left join operators, empty right collapses to left.
+		if ( op instanceof LogicalOpLeftJoin ) {
+			if ( rewrittenSubPlans.size() != 2 )
+				return rewrittenSubPlans.get(0); // if the right subplan was pruned, return the left subplan
+		}
```
hartig: I am feeling a bit uncomfortable about using `rewrittenSubPlans.size() != 2` as a condition to check whether the right subplan was pruned.

If there are no bugs in related code, I agree that this condition works. But if there are bugs, using this condition for the given purpose will make it very difficult for us to even observe any problem caused by these bugs.

A more robust approach is to move this whole if-block before the for-loop and, then, you replace the condition of the inner if-block by: `isProvablyEmpty(inputPlan.getSubPlan(1))`

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+			}
+
+			rewrittenSubPlans.add(_apply(child));
+		}
+
+		// For union operators, remove empty subplans and collapse if only one remains.
+		if ( op instanceof LogicalOpUnion || op instanceof LogicalOpMultiwayUnion ) {
+			if ( rewrittenSubPlans.size() == 1 )
+				return rewrittenSubPlans.get(0);
+			else {
```
hartig: Just for extra safety, replace this `else` as follows
```suggestion
			else if ( rewrittenSubPlans.size() > 1 ) {
```
and then add an `else` block that throws an `IllegalStateException()`. If everything is correct and there are no bugs, the code should never end up in this else-block. But if there is a bug, then this change makes it more likely that we will notice this bug.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		// Rebuild the plan with rewritten subplans, preserving operator and properties.
+		return LogicalPlanUtils.createPlanWithSubPlans( op,
+		                                                inputPlan.getQueryPlanningInfo().getProperties(),
+		                                                rewrittenSubPlans );
+	}
+
+	protected boolean isProvablyEmpty ( final LogicalPlan plan ) {
+		// Returns true if the plan is guaranteed empty
```
hartig: Again, the plan itself is not empty.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResults.java
```diff
@@ -0,0 +1,122 @@
...
+		// (via accurate cardinality 0).
+		QueryPlanningInfo info = plan.getQueryPlanningInfo();
+		if (info == null) return false;
+
+		QueryPlanProperty crd = info.getProperty(CARDINALITY);
```
hartig:
```suggestion
		final QueryPlanningInfo info = plan.getQueryPlanningInfo();
		if (info == null) return false;

		final QueryPlanProperty crd = info.getProperty(CARDINALITY);
```

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResultsTest.java
```diff
@@ -0,0 +1,480 @@
...
+import se.liu.ida.hefquin.engine.queryproc.QueryProcContext;
+import se.liu.ida.hefquin.federation.FederationMember;
+import se.liu.ida.hefquin.federation.access.FederationAccessManager;
+import se.liu.ida.hefquin.federation.access.impl.req.SPARQLRequestImpl;
+import se.liu.ida.hefquin.federation.catalog.FederationCatalog;
+
+public class RemoveSubPlansWithEmptyResultsTest extends EngineTestBase
+{
+	@Test
+	public void removeEmptySubPlansRemoveUnionKeepSubPlan() {
```
hartig: It is not necessary to prefix all these method names with `removeEmptySubPlans`. For instance, hits one here may simply be called `removeUnionKeepSubPlan`.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResultsTest.java
```diff
@@ -0,0 +1,480 @@
...
+		assertTrue( left.getRootOperator() instanceof LogicalOpRequest );
+
+		final LogicalPlan right = result.getSubPlan(1);
+		assertTrue( right.getRootOperator() instanceof LogicalOpRequest );
+	}
+
+	@Test
+	public void removeEmptySubPlansNestedUnionEmptyInner() {
+		// Nested union operator where an inner union contains an empty branch.
+		// Inner union collapses and outer union keeps the remaining non-empty branches.
```
hartig: This comment is the same as in the previous test. Somehow they should be different, given that the test are different.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResultsTest.java
```diff
@@ -0,0 +1,480 @@
...
+		// Empty request
+		final LogicalPlan reqPlan1 = new LogicalPlanWithNullaryRootImpl(reqOp1, null);
+		final QueryPlanningInfo qpInfo1 = reqPlan1.getQueryPlanningInfo();
+		qpInfo1.addProperty( QueryPlanProperty.cardinality(0, Quality.ACCURATE) );
+		qpInfo1.addProperty( QueryPlanProperty.maxCardinality(0, Quality.ACCURATE) );
+		qpInfo1.addProperty( QueryPlanProperty.minCardinality(0, Quality.ACCURATE) );
+
+		// Request with non-zero cardinality
+		final LogicalOpRequest<?,?> reqOp2 = new LogicalOpRequest<>( fm, false, new SPARQLRequestImpl(tp) );
+		List<QueryPlanProperty> qpInfo2 = new ArrayList<>();
```
hartig:
```suggestion
		final List<QueryPlanProperty> qpInfo2 = new ArrayList<>();
```

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResultsTest.java
```diff
@@ -0,0 +1,480 @@
...
+		final Var v1 = Var.alloc("x");
+		final FederationMember fm = new TPFServerForTest();
+		final TriplePattern tp = new TriplePatternImpl(v1, v1, v1);
+		final LogicalOpRequest<?,?> reqOp = new LogicalOpRequest<>( fm, false, new SPARQLRequestImpl(tp) );
+
+		final LogicalPlan leftReqPlan = new LogicalPlanWithNullaryRootImpl(reqOp, null);
+		final QueryPlanningInfo leftQPInfo = leftReqPlan.getQueryPlanningInfo();
+		leftQPInfo.addProperty( QueryPlanProperty.cardinality(0, Quality.ACCURATE) );
+		leftQPInfo.addProperty( QueryPlanProperty.maxCardinality(0, Quality.ACCURATE) );
+		leftQPInfo.addProperty( QueryPlanProperty.minCardinality(0, Quality.ACCURATE) );
```
hartig: Given that you repeat this a lot in all these tests, you may move this code into a helper method at th eend of the class. After that, you may replace these four lines by a single line as follows:
```java
addCardEstimate( 0, leftReqPlan.getQueryPlanningInfo() );
```

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResultsTest.java
hartig: Actually, also for this and the next three lines, you may introduce a helper method at the end of the class.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResultsTest.java
```diff
@@ -0,0 +1,480 @@
...
+
+		// Test
+		final LogicalPlan result = new RemoveSubPlansWithEmptyResults(new TestQueryProcContext()).apply(leftJoinPlan);
+
+		// Check
+		assertTrue( result.getRootOperator() instanceof LogicalOpRequest );
+		assertEquals( 0, result.numberOfSubPlans() );
+	}
+
+	@Ignore("This test currently fails because the cardinality estimator degrades the cardinality quality (TODO #585).")
```
hartig, line 358: Yes! Correct observation!! :-)

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/RemoveSubPlansWithEmptyResultsTest.java
```diff
@@ -0,0 +1,480 @@
...
+
+		// Test
+		final LogicalPlan result = new RemoveSubPlansWithEmptyResults(new TestQueryProcContext()).apply(leftJoinPlan);
+
+		// Check
+		assertTrue( result instanceof LogicalPlanWithoutResult );
+	}
+
+	@Test
+	public void removeEmptySubPlansRemoveEntirePlan() {
```
hartig: Can you please add two more unit tests: One of like this one here, but with ESTIMATE as quality score (hence, the plan should remain unchanged, exactly as in the next test). The other one like `removeEmptySubPlansRemoveUnionKeepSubPlan` (the very first one above), but also with ESTIMATE as quality score.

hartig (changes requested):

keski:
> Thanks @keski ! @kentthang010 don't let this stop you from continuing to work on this PR.

I've been trying to find the root cause of the exception and it doesn't seem to be related to `hefquin-service`. The tests also fail for `hefquin-cli` and when compiled without tests, queries issued via the command line fail too. However, if `RemoveSubPlansWithEmptyResults` is disabled in the engine config file everything works.

hartig:
I found the issue that caused the unit tests to fail: The changes to the config file that this PR made had been wrong. Since the new `RemoveSubPlansWithEmptyResults` class has constructors that expect an argument, the corresponding config file entry needs to specify this argument. Commit https://github.com/LiUSemWeb/HeFQUIN/pull/584/commits/2dfd9ec7f6283f52b6c80fb54ead2da90763a7d2 fixes the issue.

kentthang010:
Nice! I'll make a note to remember to do this in potential future issues.

hartig (approved):
Kent, thanks for addressing my comments on this PR. After fixing the issue with the unit test failures, I also improved the code comments in `RemoveSubPlansWithEmptyResults` a bit (commit https://github.com/LiUSemWeb/HeFQUIN/pull/584/commits/3507311ae81c62f15308b7b1f6dd8c86052ef300). Now the PR is ready to be merged.

PR #586, opened 2026-04-17
Add LogicalOpMinus

### LINKED ISSUE #564: Native support for MINUS feature of SPARQL
One of the operators of SPARQL that HeFQUIN does not yet support natively (but only through its integration into the Jena query processor) is MINUS (see the [informal description of MINUS in the SPARQL spec](https://www.w3.org/TR/sparql12-query/#neg-minus) and the [definition of the corresponding algebra operator in the spec](https://www.w3.org/TR/sparql12-query/#defn_algMinus)). The task is to add a logical operator, a physical operator, and an executable operator, and to wire them together in same way as you have done for DISTINCT.

The easiest-to-implement algorithm for this case can work in the following way. First, consume all solution mappings from the pattern inside MINUS (i.e., the solution mappings in Ω<sub>2</sub> in the definition mentioned above) and put them into a hash table. Thereafter, consume the solution mappings from the pattern outside of the MINUS (i.e., the solution mappings in Ω<sub>1</sub>); for each of them, use the hash table to check whether it has no join partners in the hash table (at least none with which it shares at least one variable) and send it to the output sink if it doesn't have such join partners. This is almost the same as what happens in `ExecOpHashJoin2`. The only difference is in the `produceOutput` method. Hence, you may implement a new class called `ExecOpHashBasedMinus` as an extension of `ExecOpHashJoin2` and override the `produceOutput` method.

kentthang010:
Addresses the logical operator part of #564.

Also adds tests for `MergeRequests` and `FilterPushDown`.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanUtils.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanUtils.java
@@ -98,6 +98,21 @@ public class LogicalPlanUtils
 	}
 
 	/**
+	 * Creates a {@link LogicalPlan} with a {@link LogicalOpMinus} as
+	 * root operator and the given plans as its subplans.
+	 *
+	 * @param qpInfo - to be copied as the query-planning info of the created
+	 *                 plan; may be {@code null}, in which case plan is created
+	 *                 without initial query-planning info
+	 */
+	public static LogicalPlan createPlanWithMinus( final boolean mayReduce,
+	                                                     final LogicalPlan subPlan1,
+	                                                     final LogicalPlan subPlan2,
+	                                                     final Iterable<QueryPlanProperty> qpInfo ) {
+		return createPlanWithSubPlans( LogicalOpMinus.getInstance(mayReduce), qpInfo, subPlan1, subPlan2 );
+	}
+
+	/**
 	 * Creates a {@link LogicalPlan} with the given operator as root operator
 	 * and the given plans as subplans.
 	 *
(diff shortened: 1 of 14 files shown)


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanUtils.java
```diff
@@ -97,6 +97,21 @@ public static LogicalPlan createPlanWithMultiwayUnion( final boolean mayReduce,
...
+	 * root operator and the given plans as its subplans.
+	 *
+	 * @param qpInfo - to be copied as the query-planning info of the created
+	 *                 plan; may be {@code null}, in which case plan is created
+	 *                 without initial query-planning info
+	 */
+	public static LogicalPlan createPlanWithMinus( final boolean mayReduce,
+	                                                     final LogicalPlan subPlan1,
+	                                                     final LogicalPlan subPlan2,
+	                                                     final Iterable<QueryPlanProperty> qpInfo ) {
```
hartig: Something is off with the indentation of these lines. Notice that I am using tabs for indenting statements, but for a multi-line statement such as a single function call or the declaration of a function with its arguments, the individual lines are internally aligned using spaces. This way, the internal alignment doesn't depend on the tab size.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/logical/LogicalPlanVisitor.java
```diff
@@ -17,12 +17,13 @@ public interface LogicalPlanVisitor
...
-	
+
 	void visit( final LogicalOpFilter op );
 	void visit( final LogicalOpBind op );
 	void visit( final LogicalOpUnfold op );
 	void visit( final LogicalOpLocalToGlobal op );
 	void visit( final LogicalOpGlobalToLocal op );
 	void visit( final LogicalOpDedup op );
 	void visit( final LogicalOpProject op );
+	void visit( final LogicalOpMinus op );
```
hartig: Please move this under the line for left join. It fits better there because those operators are more similar.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
```diff
@@ -480,6 +502,31 @@ protected SPARQLGraphPattern mergePatternWithOptPatterns( final SPARQLGraphPatte
...
 
+	protected SPARQLGraphPattern mergePatternWithMinusPatterns( final SPARQLGraphPattern pattern,
+	                                                            final SPARQLGraphPattern ... minusPatterns ) {
+		assert minusPatterns.length > 0;
+
+		final ElementGroup group = new ElementGroup();
+
+		final Element elmt = QueryPatternUtils.convertToJenaElement(pattern);
+		if ( elmt instanceof ElementGroup ) {
+			for ( final Element subElmt : ((ElementGroup) elmt).getElements() ) {
```
hartig:
```suggestion
		if ( elmt instanceof ElementGroup eg ) {
			for ( final Element subElmt : eg.getElements() ) {
```

hartig (changes requested):
Excellent job! Just some very minor points.

hartig (approved):

PR #587, opened 2026-04-20
Add ExecOpHashBasedMinus and tests

kentthang010:
Addresses the executable operator part of #564.

Tests are admittedly minimal. I initially planned to try out cases where
1) subtraction was performed on an empty solution mapping.
2) subtraction was performed with an empty solution mapping as subtrahend
but both cases seem to fail the assertion of row 59 in `SolutionMappingsHashTable`.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinus.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinus.java
@@ -7,6 +7,11 @@ import se.liu.ida.hefquin.base.data.SolutionMapping;
 import se.liu.ida.hefquin.base.query.ExpectedVariables;
 import se.liu.ida.hefquin.engine.queryplan.info.QueryPlanningInfo;
 
+/**
+ * To be used for MINUS clauses. This operator extends {@link ExecOpHashJoin2} to
+ * calculate solution mappings in the left-hand side that are not compatible with
+ * the solutions on the right-hand side.
+ */
 public class ExecOpHashBasedMinus extends ExecOpHashJoin2
 {
 	public ExecOpHashBasedMinus( final boolean mayReduce,
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinus.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinus.java
@@ -1,0 +1,32 @@
+package se.liu.ida.hefquin.engine.queryplan.executable.impl.ops;
+
+import java.util.Iterator;
+import java.util.List;
+
+import se.liu.ida.hefquin.base.data.SolutionMapping;
+import se.liu.ida.hefquin.base.query.ExpectedVariables;
+import se.liu.ida.hefquin.engine.queryplan.info.QueryPlanningInfo;
+
+public class ExecOpHashBasedMinus extends ExecOpHashJoin2
+{
+	public ExecOpHashBasedMinus( final boolean useOuterJoinSemantics,
+	                             final boolean mayReduce,
+	                             final ExpectedVariables inputVars1,
+	                             final ExpectedVariables inputVars2,
+	                             final boolean collectExceptions,
+	                             final QueryPlanningInfo qpInfo ) {
+		super(useOuterJoinSemantics, mayReduce, inputVars1, inputVars2, collectExceptions, qpInfo);
+	}
+
+	@Override
+	protected void produceOutput( final SolutionMapping inputSolMap,
+	                              final List<SolutionMapping> output) {
+		final Iterable<SolutionMapping> joinPartners = index.getJoinPartners(inputSolMap);
+		final Iterator<SolutionMapping> it = joinPartners.iterator();
+
+		if ( ! it.hasNext() ) {
+			output.add(inputSolMap);
+		}
+	}
+
+}
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinus.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinus.java
@@ -9,13 +9,12 @@ import se.liu.ida.hefquin.engine.queryplan.info.QueryPlanningInfo;
 
 public class ExecOpHashBasedMinus extends ExecOpHashJoin2
 {
-	public ExecOpHashBasedMinus( final boolean useOuterJoinSemantics,
-	                             final boolean mayReduce,
+	public ExecOpHashBasedMinus( final boolean mayReduce,
 	                             final ExpectedVariables inputVars1,
 	                             final ExpectedVariables inputVars2,
 	                             final boolean collectExceptions,
 	                             final QueryPlanningInfo qpInfo ) {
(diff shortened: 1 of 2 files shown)


hartig:
>        but both cases seem to fail the assertion of row 59 in `SolutionMappingsHashTable`.

Yes. The two hash joins (`ExecOpHashJoin1` and `ExecOpHashJoin2`) can be used only if the `ExpectedVariables` of the two subplans have at least one certain variable in common, because the hash table can only be built if there is at least one join variable that is guaranteed to be in every solution mapping (from any of the two inputs). The `supports` function of the `Factory` of `PhysicalOpHashJoin2` checks this condition, and the constructor of `BaseForExecOpHashJoin` creates the hash table based on this condition.

(I have teaching now and will try to look at the PR asap.)

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinus.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class ExecOpHashBasedMinus extends ExecOpHashJoin2
+{
+	public ExecOpHashBasedMinus( final boolean useOuterJoinSemantics,
+	                             final boolean mayReduce,
+	                             final ExpectedVariables inputVars1,
+	                             final ExpectedVariables inputVars2,
+	                             final boolean collectExceptions,
+	                             final QueryPlanningInfo qpInfo ) {
+		super(useOuterJoinSemantics, mayReduce, inputVars1, inputVars2, collectExceptions, qpInfo);
```
hartig: `useOuterJoinSemantics` is irrelevant in this case. So, there is no need to include it as an argument of the constructor. You can change it as follows.

```suggestion
	public ExecOpHashBasedMinus( final boolean mayReduce,
	                             final ExpectedVariables inputVars1,
	                             final ExpectedVariables inputVars2,
	                             final boolean collectExceptions,
	                             final QueryPlanningInfo qpInfo ) {
		super(true, mayReduce, inputVars1, inputVars2, collectExceptions, qpInfo);
```

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinusTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+		Set<Var> varsCertain1 = new HashSet<>();
+		varsCertain1.add(var1);
+		varsCertain1.add(var2);
+		Set<Var> varsPossible1 = new HashSet<>();
+
+		Set<Var> varsCertain2 = new HashSet<>();
+		varsCertain2.add(var1);
+		varsCertain2.add(var2);
+		Set<Var> varsPossible2 = new HashSet<>();
```
hartig: There is a a simpler way to write this:

```suggestion
		final Set<Var> varsCertain1 = Set.of(var1, var2);
		final Set<Var> varsPossible1 = Set.of();

		final Set<Var> varsCertain2 = Set.of(var1, var2);
		final Set<Var> varsPossible2 = Set.of();
```

hartig: The same simplification can also be applied in the next unit test (and in many similar cases; e.g., for `input1` and `input2` above).

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinusTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+		ExpectedVariables[] inputVars = getExpectedVariables(varsCertain1, varsPossible1, varsCertain2, varsPossible2);
```
hartig:
```suggestion
		final ExpectedVariables[] inputVars = getExpectedVariables(varsCertain1, varsPossible1, varsCertain2, varsPossible2);
```

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinusTest.java
```diff
@@ -0,0 +1,181 @@
...
+import org.apache.jena.sparql.core.Var;
+import org.junit.Test;
+
+import se.liu.ida.hefquin.base.data.SolutionMapping;
+import se.liu.ida.hefquin.base.data.utils.SolutionMappingUtils;
+import se.liu.ida.hefquin.base.query.ExpectedVariables;
+import se.liu.ida.hefquin.engine.queryplan.executable.impl.CollectingIntermediateResultElementSink;
+import se.liu.ida.hefquin.engine.queryproc.ExecutionException;
+
+public class ExecOpHashBasedMinusTest
```
hartig, line 22: Can you please add one more unit test in which the solution mappings from the two inputs have only one variable in common, and each input has an additional variable that the other input doesn't have. In other words, the certain variables of the left input may be {v1,v2} and of the right input: {v1,v3}.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinusTest.java
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	/**
+	 * Sends second input first.
+	 */
+	protected Iterator<SolutionMapping> runTest(
+			final List<SolutionMapping> input1,
+			final List<SolutionMapping> input2,
+			final boolean sendAllSolMapsSeparately,
+			final boolean useOuterJoinSemantics,
```
hartig: Not needed.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinusTest.java
```diff
@@ -0,0 +1,181 @@
...
+			final ExpectedVariables... inputVars ) throws ExecutionException
+	{
+		final CollectingIntermediateResultElementSink sink = new CollectingIntermediateResultElementSink();
+
+		final ExecOpHashBasedMinus op = createExecOpForTest( useOuterJoinSemantics, inputVars );
```
hartig: I think that, in this case, you don't need the extract helper method `createExecOpForTest`. Instead, you can simply create the `ExecOpHashBasedMinus` object directly here.

hartig (changes requested):
Just a few minor points.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/ExecOpHashBasedMinusTest.java
```diff
@@ -0,0 +1,210 @@
...
+	protected ExecOpHashBasedMinus createExecOpForTest(
+			final boolean useOuterJoinSemantics,
+			final ExpectedVariables... inputVars ) {
+		assert inputVars.length == 2;
+
+		return new ExecOpHashBasedMinus( false,            // mayReduce
+		                                 inputVars[0], inputVars[1],
+		                                 false,    // collectExceptions
+		                                 null );              // qpInfo
+	}
```
hartig: You forgot to remove this helper method (after inlining its content directly into the `runTest` method above).

hartig (changes requested):

hartig (approved):
Looks good. I will merge this one now and then you can move on to doing the third PR #564.

PR #589, opened 2026-04-20
Add push into pattern of gpAdd and gpOptAdd

kentthang010:
Closes #578.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDown.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDown.java
@@ -384,39 +384,6 @@ public class FilterPushDown implements HeuristicForLogicalOptimization
 	                                                     final UnaryLogicalOp childOp,
 	                                                     final LogicalPlan subPlanUnderChildOp,
 	                                                     final LogicalPlan inputPlan ) {
-		// The following snippets of code will be needed if we try to push
-		// the filter conditions also into the pattern of the xxAdd operator.
-		// Trying to do so is not necessary at the moment because we do not
-		// actually have xxAdd operators in the logical plans to which we
-		// apply the heuristics (because the initial logical plan is always
-		// a source assignment and we currently don't have any heuristics
-		// that rewrite the multiway joins of such source assignments into
-		// xxAdd operators). Even once we have heuristics that rewrite joins
-		// into xxAdd operators, as long as we apply the MergeRequest heuristic
-		// before such a heuristic (and after FilterPushDown), we would never
-		// need to try pushing filter conditions also into the pattern of the
-		// xxAdd operator.
-		// However, should we ever try to push the filter conditions also
-		// into the pattern of the xxAdd operator, notice that the xxOptAdd
-		// operators are special because not every filter condition can be
-		// pushed into the pattern of such an operator even if the certain
-		// variables of that pattern contain all of the variables mention
-		// in the filter condition. For instance, if the optional pattern
-		// has ?x as a certain variable and the filter condition is
-		// !BOUND(?x), then pushing this condition is not possible.
-
-		//final SPARQLGraphPattern patternOfAddOp;
-		//if      ( childOp instanceof LogicalOpTPAdd )     patternOfAddOp = ((LogicalOpTPAdd) childOp).getTP();
-		//else if ( childOp instanceof LogicalOpTPOptAdd )  patternOfAddOp = ((LogicalOpTPOptAdd) childOp).getTP();
-		//else if ( childOp instanceof LogicalOpBGPAdd )    patternOfAddOp = ((LogicalOpBGPAdd) childOp).getBGP();
-		//else if ( childOp instanceof LogicalOpBGPOptAdd ) patternOfAddOp = ((LogicalOpBGPOptAdd) childOp).getBGP();
-		//else if ( childOp instanceof LogicalOpGPAdd )     patternOfAddOp = ((LogicalOpGPAdd) childOp).getPattern();
-		//else if ( childOp instanceof LogicalOpGPOptAdd )  patternOfAddOp = ((LogicalOpGPOptAdd) childOp).getPattern();
-		//else throw new IllegalArgumentException( childOp.getClass().getName() );
-
-		//final ExpectedVariables expVarsInPattern = QueryPatternUtils.getExpectedVariablesInPattern(patternOfAddOp);
-		//final Set<Var> certainVarsInPattern = expVarsInPattern.getCertainVariables();
-
 		final SPARQLGraphPattern pattern;
 		final FederationMember fm;
 		if ( childOp instanceof LogicalOpGPAdd gpAddOp ) {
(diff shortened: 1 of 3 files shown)


kentthang010:
Not sure if I have fully understood the issue, so I made a draft to share the code for clarity rather than sending an email.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDown.java
```diff
@@ -440,6 +436,17 @@ protected LogicalPlan createPlanForAddOpUnderFilter( final LogicalOpFilter paren
 			return createPlanAfterPushingInSubPlan(parentFilterOp, childOp, subPlanUnderChildOp, inputPlan);
 		}
 
+		if ( childOp instanceof LogicalOpGPAdd gpAddOp
+		  && gpAddOp.getFederationMember().supportsMoreThanTriplePatterns() ) {
+			// The filter can be pushed into the pattern of the given GPAdd operator.
+			gpAddOp.getPattern().mergeWith(toBePushed);
```
hartig: That's not correct. `toBePushed` was populated specifically for the subplan under the gpAdd operator. You need to create and populate a second `toBePushed` for the graph pattern of the gpAdd operator. Take a look at `createPlanForJoinUnderFilter` which works in a similar way, creating a separate `toBePushed` for every subplan under the join. Yet, `createPlanForJoinUnderFilter` is a bit more flexible than what you need for gpAdd, for which you really need only two versions of `toBePushed`: one for the subplan under the gpAdd operator---call that one `toBePushedIntoSubPlan`---and one for the graph pattern of the gpAdd operator---call that one `toBePushedIntoPattern`. For populating the latter, you need the set of certain variables of the graph pattern, which you can simply obtain by calling `gpAddOp.getPattern().getCertainVariables()`.

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDown.java
```diff
@@ -421,40 +417,87 @@ protected LogicalPlan createPlanForAddOpUnderFilter( final LogicalOpFilter paren
...
+		// Apply pattern-level filter pushdown:
+		// expressions that are safe with the GPAdd pattern (i.e. only depend
+		// on variables guaranteed by the pattern) are merged into the SPARQL
+		// graph pattern itself, so they become part of the remote request.
+		UnaryLogicalOp newChildOp = childOp;
+		if ( ! toBePushedIntoPattern.isEmpty() ) {
+			SPARQLGraphPattern newPattern = pattern.mergeWith(toBePushedIntoPattern);
+
+			if ( childOp instanceof LogicalOpGPAdd gpAdd && gpAdd.hasParameterVariables() )
+				newChildOp = new LogicalOpGPAdd(fm, newPattern, ((LogicalOpGPAdd) childOp).getParameterVariables(), childOp.mayReduce());
```
hartig: Since the first part of the if-condition introduces the variable `gpAdd`, you can use that variable also inside the if-block:

```suggestion
			if ( childOp instanceof LogicalOpGPAdd gpAdd && gpAdd.hasParameterVariables() )
				newChildOp = new LogicalOpGPAdd(fm, newPattern, gpAdd.getParameterVariables(), childOp.mayReduce());
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDown.java
```diff
@@ -421,40 +417,87 @@ protected LogicalPlan createPlanForAddOpUnderFilter( final LogicalOpFilter paren
...
 			return createPlanAfterPushingInSubPlan(parentFilterOp, childOp, subPlanUnderChildOp, inputPlan);
 		}
 
+		// Apply pattern-level filter pushdown:
+		// expressions that are safe with the GPAdd pattern (i.e. only depend
+		// on variables guaranteed by the pattern) are merged into the SPARQL
+		// graph pattern itself, so they become part of the remote request.
+		UnaryLogicalOp newChildOp = childOp;
+		if ( ! toBePushedIntoPattern.isEmpty() ) {
+			SPARQLGraphPattern newPattern = pattern.mergeWith(toBePushedIntoPattern);
```
hartig:
```suggestion
			final SPARQLGraphPattern newPattern = pattern.mergeWith(toBePushedIntoPattern);
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDown.java
```diff
@@ -421,40 +417,87 @@ protected LogicalPlan createPlanForAddOpUnderFilter( final LogicalOpFilter paren
...
+
```
hartig: It is not enough to just check whether the federation member `supportsMoreThanTriplePatterns`. It is also necessary at this point (after having created the merged pattern), to check whether the federation member supports the merged pattern. Use the `isSupportedPattern` method of `fm` for this purpose. If `fm.isSupportedPattern(newPattern)` returns `true`, the code can proceed with creating the `newChildOp`.

However, if `fm.isSupportedPattern(newPattern)` returns `false`, the code needs to backtrack. That is, the gpAdd operator (including its graph pattern) should remain untouched; in other words, `newChildOp` should remain `childOp`. But this also has another consequence: every expression that has been put into `toBePushedIntoPattern` needs to be kept in the filter on top of the gpAdd operator _unless_ (!!!) the expression has also been put into `toBePushedIntoSubPlan` (in which case it does not need to be kept in the filter on top of the gpAdd). Instead of trying to identify these expressions after the fact, it might be better to make the distinction already when populating the `toBePushedInto...` sets. To this end, you may split `toBePushedIntoPattern` into two sets: `toBePushedIntoPatternOrKept` and `toBePushedIntoPatternOrDropped`.

Sorry, you couldn't know this.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDown.java
```diff
@@ -421,40 +417,87 @@ protected LogicalPlan createPlanForAddOpUnderFilter( final LogicalOpFilter paren
...
+			if ( childOp instanceof LogicalOpGPAdd gpAdd && gpAdd.hasParameterVariables() )
+				newChildOp = new LogicalOpGPAdd(fm, newPattern, ((LogicalOpGPAdd) childOp).getParameterVariables(), childOp.mayReduce());
+			else if ( childOp instanceof LogicalOpGPAdd gpAdd && ! gpAdd.hasParameterVariables() )
+				newChildOp = new LogicalOpGPAdd(fm, newPattern, null, childOp.mayReduce());
+			else if ( childOp instanceof LogicalOpGPOptAdd )
+				newChildOp = new LogicalOpGPOptAdd(fm, newPattern, childOp.mayReduce());
```
hartig: As I wrote in the comment at the beginning of his method, it is not always correct to push a filter into the pattern of a gpOptAdd. For the moment, I would simply not push at all if 'childOp` is a `LogicalOpGPOptAdd`. As a consequence, you do not even need to populate `toBePushedIntoPattern` (resp. `toBePushedIntoPatternOrKept` and `toBePushedIntoPatternOrDropped`) if 'childOp` is a `LogicalOpGPOptAdd`.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
```diff
@@ -392,6 +392,23 @@ public boolean isSupportedPattern( final SPARQLGraphPattern p ) {
 		}
 	}
 
+	public static class GraphCapableFederationMemberForTest extends FederationMemberBaseForTest
```
hartig: Instead of introducing this extra new class, you can simply use `SPARQLEndpointForTest` in the new unit tests.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDownTest.java
```diff
@@ -613,4 +616,48 @@ public void pushFilterUnderFixedSolMapUnsatisfied() {
...
+	@Test
+	public void pushFilterUnderGPAdd() {
+		// a filter on top of a GPAdd operator, where filter expressions
+		// are pushed into the GPAdd pattern and also pushed into the
+		// subplan request under the GPAdd operator
+
+		// set up
+		final Var v1 = Var.alloc("x");
+		final Var v2 = Var.alloc("y");
+		final FederationMember fm = new GraphCapableFederationMemberForTest(GraphFactory.createDefaultGraph());
```
hartig: You can simply do the following instead.
```suggestion
		final FederationMember fm = new SPARQLEndpointForTest();
```

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/FilterPushDownTest.java
```diff
@@ -613,4 +616,48 @@ public void pushFilterUnderFixedSolMapUnsatisfied() {
 		assertEquals( filterPlan, result );
 	}
 
+	@Test
+	public void pushFilterUnderGPAdd() {
```
hartig:
```suggestion
	public void pushFilterBothUnderAndIntoGPAdd() {
```

hartig (changes requested):

kentthang010:
Latest commit should resolve the PR comments above! It also adds more test cases, including a test that tests out partial pushdown under and into GPAdd.

I opted to keep the new mock federation member but changed its function so that it always returns `false` on `isSupportedPattern` to test the case where pushdown into pattern fails. To my knowledge, there isn't any existing mock federation members that do this?

hartig (approved):
I have edited the code a bit more: some minor reorganization and some more comment, all to make it easier to understand what is going on in this method.

Now, the PR is ready to be merged.

PR #592, opened 2026-04-22
Add PhysicalOpMinus

### LINKED ISSUE #564: Native support for MINUS feature of SPARQL

kentthang010:
Addresses the physical operator part of #564. As it is the last part of the issue left to address, this PR closes #564.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/BaseForPhysicalOpBinaryJoin.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/BaseForPhysicalOpBinaryJoin.java
@@ -17,9 +17,11 @@ public abstract class BaseForPhysicalOpBinaryJoin
 		implements BinaryPhysicalOpForLogicalOp
 {
 	protected final boolean useOuterJoinSemantics;
+	protected final boolean mayReduce;
 
-	protected BaseForPhysicalOpBinaryJoin( final boolean useOuterJoinSemantics ) {
+	protected BaseForPhysicalOpBinaryJoin( final boolean useOuterJoinSemantics, final boolean mayReduce ) {
 		this.useOuterJoinSemantics = useOuterJoinSemantics;
+		this.mayReduce = mayReduce;
 	}
 
 	@Override
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpHashJoin2.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpHashJoin2.java
@@ -38,11 +38,13 @@ public class PhysicalOpHashJoin2 extends BaseForPhysicalOpBinaryJoin
 	protected static final Factory factory = new Factory();
 	public static PhysicalOpFactory getFactory() { return factory; }
 
-	private static PhysicalOpHashJoin2 singletonForInnerJoin = null;
-	private static PhysicalOpHashJoin2 singletonForOuterJoin = null;
+	private static PhysicalOpHashJoin2 innerJoinMayReduce = null;
+	private static PhysicalOpHashJoin2 innerJoinNoReduce = null;
+	private static PhysicalOpHashJoin2 outerJoinMayReduce = null;
+	private static PhysicalOpHashJoin2 outerJoinNoReduce = null;
 
-	protected PhysicalOpHashJoin2( final boolean useOuterJoinSemantics ) {
-		super(useOuterJoinSemantics);
+	protected PhysicalOpHashJoin2( final boolean useOuterJoinSemantics, final boolean mayReduce ) {
+		super(useOuterJoinSemantics, mayReduce);
 	}
 
 	@Override
@@ -52,7 +54,7 @@ public class PhysicalOpHashJoin2 extends BaseForPhysicalOpBinaryJoin
 		assert inputVars.length == 2;
 
 		return new ExecOpHashJoin2( useOuterJoinSemantics,
-		                            getLogicalOperator().mayReduce(),
+		                            mayReduce,
 		                            inputVars[0], inputVars[1],
 		                            collectExceptions,
 		                            qpInfo );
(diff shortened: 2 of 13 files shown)


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpMinus.java
```diff
@@ -0,0 +1,80 @@
...
+		public boolean supports( final LogicalOperator lop, final ExpectedVariables... inputVars ) {
+			if ( ! (lop instanceof LogicalOpMinus) )
+				return false;
+
+			// inputVars contains null value?
+			for ( final ExpectedVariables vars : inputVars ) {
+				if ( vars == null ) return false;
+			}
+
+			return true;
```
hartig: Both `PhysicalOpHashJoin1` and `PhysicalOpHashJoin2` have an additional condition here, related to the availability of at least one certain join variable. As I wrote earlier (in an email, maybe), this is needed for the hash table and, thus, needs to be checked here as well.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpMinus.java
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		@Override
+		public PhysicalOpMinus create( final BinaryLogicalOp lop ) {
+			if ( lop instanceof LogicalOpMinus ) {
+				return new PhysicalOpMinus();
```
hartig: Instead of repeatedly creating new Java objects, please use the singleton pattern with two singletons as done in `PhysicalOpHashJoin2`. But instead of having the singletons for inner join versus outer join, here we need one for `mayReduce==true` and one for `mayReduce==false`.

In fact, I notice now that the base class (`BaseForPhysicalOpBinaryJoin`) is incomplete. It needs to have the `mayReduce` flag as an additional member variable and, thus, also as an additional constructor argument. Adding this will require you to extend `PhysicalOpHashJoin1` and `PhysicalOpHashJoin2` as well.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpMinus.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	public void visit( final PhysicalPlanVisitor visitor ) {
+		visitor.visit(this);
+	}
+
+	@Override
+	public boolean equals( final Object o ) {
+		if ( o == this ) return true;
+
+		return o instanceof PhysicalOpMinus;
```
hartig: After adding the `mayReduce`  flag to the base class (see one of my other comments),  the value of this flag needs to be taken into account here as well:
```suggestion
		return o instanceof PhysicalOpMinus other 
		    && other.mayReduce == mayReduce;
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpMinus.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	@Override
+	public int hashCode() {
+		return getClass().hashCode() ^ getLogicalOperator().hashCode();
```
hartig: After adding the `mayReduce`  flag to the base class (see one of my other comments), change this one as follows.
```suggestion
		return getClass().hashCode() ^ (mayReduce ? 1 : 0);
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpMinus.java
```diff
@@ -0,0 +1,80 @@
...
+		super(true);
+	}
+
+	@Override
+	public BinaryExecutableOp createExecOp( final boolean collectExceptions,
+	                                        final QueryPlanningInfo qpInfo,
+	                                        final ExpectedVariables ... inputVars ) {
+		assert inputVars.length == 2;
+
+		return new ExecOpHashBasedMinus( getLogicalOperator().mayReduce(),
```
hartig: After adding the `mayReduce`  flag to the base class (see one of my other comments),  change this one as follows:
```suggestion
		return new ExecOpHashBasedMinus( mayReduce,
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/ApplyVocabularyMappings.java
```diff
@@ -255,8 +256,22 @@ public void visit( final LogicalOpProject op ) {
...
-			// TODO Auto-generated method stub
-			throw new UnsupportedOperationException("Unimplemented method 'visit'");
+			final LogicalPlan leftOriginal  = inputPlan.getSubPlan(0);
+			final LogicalPlan rightOriginal = inputPlan.getSubPlan(1);
+
+			final LogicalPlan leftRewritten  = apply(leftOriginal);
+			final LogicalPlan rightRewritten = apply(rightOriginal);
+
+			final boolean rewritten = ! leftOriginal.equals(leftRewritten)
+			                       || ! rightOriginal.equals(rightRewritten);
```
hartig: Better use `samePlan` instead of `equals` for both of these cases.

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/BaseForPhysicalOpBinaryJoin.java
hartig, line 34: You need to override this method in `PhysicalOpMinus`.

Additionally, in the version here. you now also need to take the `mayReduce` flag into account: use the versions of these `getInstance` methods that have the flag as argument.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpHashJoin2.java
```diff
@@ -67,12 +69,13 @@ public void visit( final PhysicalPlanVisitor visitor ) {
 	public boolean equals( final Object o ) {
 		if ( o == this ) return true;
 
-		return o instanceof PhysicalOpHashJoin2;
+		return o instanceof PhysicalOpHashJoin2 oo
+		    && oo.mayReduce == mayReduce;
```
hartig:
```suggestion
		    && oo.mayReduce == mayReduce
		    && oo.useOuterJoinSemantics == useOuterJoinSemantics;
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/physical/impl/PhysicalOpHashJoin2.java
```diff
@@ -67,12 +69,13 @@ public void visit( final PhysicalPlanVisitor visitor ) {
...
 	}
 
 	@Override
 	public int hashCode() {
-		return getClass().hashCode() ^ getLogicalOperator().hashCode();
+		return getClass().hashCode() ^ (mayReduce ? 1 : 0);
```
hartig:
```suggestion
		return getClass().hashCode() ^ (mayReduce ? 1 : 0) ^ (useOuterJoinSemantics ? 1 : 0);
```

hartig (changes requested):
Almost there for this one now!

kentthang010:
Comments above resolved!

hartig (approved):
➖✔️ 🙂

PR #596, opened 2026-04-27
fix race condition in MappingExpressionUtils.evaluate by removing shared EvaluateWorker

### LINKED ISSUE #595: Race condition in MappingExpressionUtils.evaluate due to shared EvaluateWorker
`MappingExpressionUtils` uses a single static instance of `EvaluateWorker` (`evalWorker`) to evaluate mapping expressions. However, `EvaluateWorker` maintains a mutable state (`srMap`, `inputs`, `output`) that is modified during evaluation.

When `evaluate(...)` is called concurrently from multiple threads, this shared state can be overwritten, leading to race conditions, incorrect results, or unexpected exceptions. Evaluations should be thread-safe and independent.

The suggested fix would be to remove the static shared `evalWorker` and instead instantiate a new `EvaluateWorker` per evaluation.

keski:


hartig (commented):
Oh my! I didn't think of the fact that the `evaluate` method may be used by multiple threads. I am so sorry that I caused you headaches with this mistake!

The fix looks good. I will merge and update the CHANGELOG.

hartig (approved):

PR #600, opened 2026-04-28
Add option to push projection variables and distinct into SPARQLRequest

KentThang:
Addresses the extension of `SPARQLRequest` and `SPARQLRequestProcessor` part of #570.

### DIFF
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
@@ -1,5 +1,9 @@
 package se.liu.ida.hefquin.federation.access;
 
+import java.util.Set;
+
+import org.apache.jena.sparql.core.Var;
+
 import se.liu.ida.hefquin.base.query.SPARQLGraphPattern;
 import se.liu.ida.hefquin.base.query.SPARQLQuery;
 import se.liu.ida.hefquin.base.query.impl.SPARQLQueryImpl;
@@ -13,6 +17,28 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	 * accessed via the method {@link #getQuery()}.
 	 */
 	SPARQLGraphPattern getQueryPattern();
+
+	/**
+	 * Returns the set of variables that should be projected in the result.
+	 *
+	 * <p>If non-empty, this set represents a request-level projection that may
+	 * be applied when constructing the SPARQL query sent to an endpoint.
+	 * Implementations may ignore this information if applying it would not be
+	 * semantically safe (e.g., for queries with aggregation or expression-based
+	 * projections).</p>
+	 *
+	 */
+	Set<Var> getProjectionVars();
+
+	/**
+	 * Indicates whether the results of this request should be duplicate-free.
+	 *
+	 * <p>If {@code true}, the request processor may enforce this by issuing
+	 * a {@code DISTINCT} query to the endpoint (if supported).</p>
+	 *
+	 * @return {@code true} if duplicate elimination is requested; {@code false} otherwise
+	 */
+	boolean isDistinct();
 
 	/**
 	 * Returns the SPARQL query for which solutions
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
@@ -45,15 +45,11 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	 * should be requested.
 	 */
 	default SPARQLQuery getQuery() {
-		return convertToQuery( getQueryPattern(),
-		                       getProjectionVars(),
-		                       isDistinct() );
+		return convertToQuery( getQueryPattern() );
 	}
 
-	static SPARQLQuery convertToQuery( final SPARQLGraphPattern pattern,
-	                                   final Set<Var> projectionVars,
-	                                   final boolean isDistinct ) {
-		return new SPARQLQueryImpl( pattern, projectionVars, isDistinct );
+	static SPARQLQuery convertToQuery( final SPARQLGraphPattern pattern ) {
+		return new SPARQLQueryImpl( pattern );
 	}
 
 }
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
@@ -31,10 +31,8 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	Set<Var> getProjectionVars();
 
 	/**
-	 * Indicates whether the results of this request should be duplicate-free.
-	 *
-	 * <p>If {@code true}, the request processor may enforce this by issuing
-	 * a {@code DISTINCT} query to the endpoint (if supported).</p>
+	 * Returns {@code true} if this request <em>explicit</em> requires that the requested
+	 * result is duplicate free.
 	 *
 	 * @return {@code true} if duplicate elimination is requested; {@code false} otherwise
 	 */
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
@@ -19,14 +19,16 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	SPARQLGraphPattern getQueryPattern();
 
 	/**
-	 * Returns the set of variables that should be projected in the result.
+	 * Returns the set of variables that should be projected in the result,
+	 * or {@code null} if no projection is specified for this request.
 	 *
-	 * <p>If non-empty, this set represents a request-level projection that may
-	 * be applied when constructing the SPARQL query sent to an endpoint.
-	 * Implementations may ignore this information if applying it would not be
-	 * semantically safe (e.g., for queries with aggregation or expression-based
-	 * projections).</p>
+	 * <p>If a non-null set is returned (including the empty set), then
+	 * projection is considered an explicit part of this request and must
+	 * be respected by components that can support it.</p>
 	 *
+	 * <p>Deciding whether such a projection can be safely applied or pushed
+	 * into a request is the responsibility of the query planning and
+	 * rewriting logic, not of this interface or its implementations.</p>
 	 */
 	Set<Var> getProjectionVars();
 
@@ -43,11 +45,15 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	 * should be requested.
 	 */
 	default SPARQLQuery getQuery() {
-		return convertToQuery( getQueryPattern() );
+		return convertToQuery( getQueryPattern(),
+		                       getProjectionVars(),
+		                       isDistinct() );
 	}
 
-	static SPARQLQuery convertToQuery( final SPARQLGraphPattern pattern ) {
-		return new SPARQLQueryImpl( pattern );
+	static SPARQLQuery convertToQuery( final SPARQLGraphPattern pattern,
+	                                   final Set<Var> projectionVars,
+	                                   final boolean isDistinct ) {
+		return new SPARQLQueryImpl( pattern, projectionVars, isDistinct );
 	}
 
 }
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
@@ -33,7 +33,7 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	Set<Var> getProjectionVars();
 
 	/**
-	 * Returns {@code true} if this request <em>explicit</em> requires that the requested
+	 * Returns {@code true} if this request <em>explicitly</em> requires that the requested
 	 * result is duplicate free.
 	 *
 	 * @return {@code true} if duplicate elimination is requested; {@code false} otherwise
@@ -45,11 +45,13 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	 * should be requested.
 	 */
 	default SPARQLQuery getQuery() {
-		return convertToQuery( getQueryPattern() );
+		return convertToQuery( getQueryPattern(), getProjectionVars(), isDistinct() );
 	}
 
-	static SPARQLQuery convertToQuery( final SPARQLGraphPattern pattern ) {
-		return new SPARQLQueryImpl( pattern );
+	static SPARQLQuery convertToQuery( final SPARQLGraphPattern pattern,
+	                                   final Set<Var> projectionVars,
+	                                   final boolean isDistinct ) {
+		return new SPARQLQueryImpl( pattern, projectionVars, isDistinct );
 	}
 
 }
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
@@ -38,20 +38,20 @@ public interface SPARQLRequest extends DataRetrievalRequest
 	 *
 	 * @return {@code true} if duplicate elimination is requested; {@code false} otherwise
 	 */
-	boolean isDistinct();
+	boolean getDistinctRequired();
 
 	/**
 	 * Returns the SPARQL query for which solutions
 	 * should be requested.
 	 */
 	default SPARQLQuery getQuery() {
-		return convertToQuery( getQueryPattern(), getProjectionVars(), isDistinct() );
+		return convertToQuery( getQueryPattern(), getProjectionVars(), getDistinctRequired() );
 	}
 
 	static SPARQLQuery convertToQuery( final SPARQLGraphPattern pattern,
 	                                   final Set<Var> projectionVars,
-	                                   final boolean isDistinct ) {
-		return new SPARQLQueryImpl( pattern, projectionVars, isDistinct );
+	                                   final boolean distinctRequired ) {
+		return new SPARQLQueryImpl( pattern, projectionVars, distinctRequired );
 	}
 
 }
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
@@ -1,5 +1,10 @@
 package se.liu.ida.hefquin.federation.access.impl.req;
 
+import java.util.Collections;
+import java.util.Set;
+
+import org.apache.jena.sparql.core.Var;
+
 import se.liu.ida.hefquin.base.query.BGP;
 import se.liu.ida.hefquin.base.query.ExpectedVariables;
 import se.liu.ida.hefquin.federation.access.BGPRequest;
@@ -36,6 +41,16 @@ public class BGPRequestImpl implements BGPRequest
 	@Override
 	public String toString(){
 		return bgp.toString();
+	}
+
+	@Override
+	public Set<Var> getProjectionVars() {
+		return Collections.emptySet();
+	}
+
+	@Override
+	public boolean isDistinct() {
+		return false;
 	}
 
 }
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
@@ -45,7 +45,7 @@ public class BGPRequestImpl implements BGPRequest
 
 	@Override
 	public Set<Var> getProjectionVars() {
-		return Collections.emptySet();
+		return null;
 	}
 
 	@Override
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
@@ -1,6 +1,5 @@
 package se.liu.ida.hefquin.federation.access.impl.req;
 
-import java.util.Collections;
 import java.util.Set;
 
 import org.apache.jena.sparql.core.Var;
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
@@ -48,7 +48,7 @@ public class BGPRequestImpl implements BGPRequest
 	}
 
 	@Override
-	public boolean isDistinct() {
+	public boolean getDistinctRequired() {
 		return false;
 	}
 
--- a/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
+++ b/hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
@@ -1,7 +1,6 @@
 package se.liu.ida.hefquin.federation.access.impl.req;
 
-import java.util.Collections;
-import java.util.Objects;
+import java.util.HashSet;
 import java.util.Set;
 
 import org.apache.jena.sparql.core.Var;
@@ -23,8 +22,22 @@ public class SPARQLRequestImpl implements SPARQLRequest
 		assert pattern != null;
 		this.pattern = pattern;
 		this.query = null;
-		this.projectionVars = Collections.emptySet();
+		this.projectionVars = null;
 		this.isDistinct = false;
+
+		// we materialize the ExpectedVariables in this case
+		// to avoid producing it again whenever it is used
+		expectedVars = pattern.getExpectedVariables();
+	}
+
+	public SPARQLRequestImpl( final SPARQLGraphPattern pattern,
+	                          final Set<Var> projectionVars,
+	                          final boolean distinct ) {
+		assert pattern != null;
+		this.pattern = pattern;
+		this.query = null;
+		this.projectionVars = projectionVars;
+		this.isDistinct = distinct;
 
 		// we materialize the ExpectedVariables in this case
 		// to avoid producing it again whenever it is used
@@ -35,22 +48,8 @@ public class SPARQLRequestImpl implements SPARQLRequest
 		assert query != null;
 		this.query = query;
 		this.pattern = null;
-		this.projectionVars = query.getProjectionVars();
-		this.isDistinct = query.isDistinct();
-
-		// we materialize the ExpectedVariables in this case
-		// to avoid producing it again whenever it is used
-		expectedVars = query.getExpectedVariables();
-	}
-
-	public SPARQLRequestImpl( final SPARQLQuery query,
-	                          final Set<Var> projectionVars,
-	                          final boolean distinct ) {
-		assert query != null;
-		this.query = query;
-		this.pattern = null;
-		this.projectionVars = projectionVars;
-		this.isDistinct = distinct;
+		this.projectionVars = new HashSet<>( query.asJenaQuery().getProjectVars() );
+		this.isDistinct = query.asJenaQuery().isDistinct();
 
 		// we materialize the ExpectedVariables in this case
 		// to avoid producing it again whenever it is used
@@ -64,18 +63,25 @@ public class SPARQLRequestImpl implements SPARQLRequest
 
 		final SPARQLRequest oo = (SPARQLRequest) o;
 
-		return Objects.equals( getQueryPattern(), oo.getQueryPattern() )
-			&& getProjectionVars().equals( oo.getProjectionVars() )
-			&& isDistinct() == oo.isDistinct();
+		if ( pattern == null ) {
+			return oo.getQueryPattern() == null
+			    && query.equals( oo.getQuery() )
+			    && getProjectionVars().equals(oo.getProjectionVars())
+			    && isDistinct() == oo.isDistinct();
+		}
+		else {
+			return pattern.equals( oo.getQueryPattern() )
+			    && getProjectionVars().equals(oo.getProjectionVars())
+			    && isDistinct() == oo.isDistinct();
+		}
 	}
 
 	@Override
 	public int hashCode() {
-		return Objects.hash(
-			getQueryPattern(),
-			getProjectionVars(),
-			isDistinct()
-		);
+		if ( pattern == null )
+			return query.hashCode() ^ getProjectionVars().hashCode() ^ (isDistinct() ? 1 : 0);
+		else
+			return pattern.hashCode() ^ getProjectionVars().hashCode() ^ (isDistinct() ? 1 : 0);
 	}
 
 	@Override
@@ -98,7 +104,7 @@ public class SPARQLRequestImpl implements SPARQLRequest
 		if ( query != null )
 			return query;
 		else
-			return SPARQLRequest.convertToQuery( getQueryPattern() );
+			return SPARQLRequest.convertToQuery( getQueryPattern(), projectionVars, isDistinct );
 	}
 
 	@Override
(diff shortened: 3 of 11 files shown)


kentthang010:
Beyond rewriting `equals()` and the `hashcode()` for `SPARQLRequestImpl` to include the newly introduced `projectionVars` and `distinct`, I also replaced the conditional logic with a uniform comparison. The thought with this was to avoid subtle inconsistencies where logically equivalent requests could be treated differently depending on their internal representation but I've been debating whether this is actually correct or not. Let me know!

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
```diff
@@ -14,6 +18,28 @@ public interface SPARQLRequest extends DataRetrievalRequest
...
 	SPARQLGraphPattern getQueryPattern();
 
+	/**
+	 * Returns the set of variables that should be projected in the result.
+	 *
+	 * <p>If non-empty, this set represents a request-level projection that may
+	 * be applied when constructing the SPARQL query sent to an endpoint.
+	 * Implementations may ignore this information if applying it would not be
+	 * semantically safe (e.g., for queries with aggregation or expression-based
+	 * projections).</p>
```
hartig: No. This should be different now from the `mayReduce` flag. If there is a set of variables here--empty or not--then the projection is really meant to be a part of the request. Whether it would be semantically safe to push a projection into such a request is not something that the implementation of the request can decide; instead, that needs to be decided by the plan rewriting code (the other part of #570 that you will do in a separate PR). In fact, the question is actually not so much about whether pushing projection into a request is semantically safe, but more about whether requests with projection can even be expressed for the corresponding federation member.

For requests without projection, this method here should return `null`.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
```diff
@@ -14,6 +18,28 @@ public interface SPARQLRequest extends DataRetrievalRequest
...
+	 */
+	Set<Var> getProjectionVars();
+
+	/**
+	 * Indicates whether the results of this request should be duplicate-free.
+	 *
+	 * <p>If {@code true}, the request processor may enforce this by issuing
+	 * a {@code DISTINCT} query to the endpoint (if supported).</p>
+	 *
+	 * @return {@code true} if duplicate elimination is requested; {@code false} otherwise
```
hartig:
```suggestion
	 * Returns {@code true} if this request <em>explicit</em> requires that the requested
	 * result is duplicate free.
	 *
	 * @return {@code true} if duplicate elimination is requested; {@code false} otherwise
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
hartig, line 55: These should be extended as well, and so should `SPARQLQuery` and `SPARQLQueryImpl`. Otherwise, these methods may give us a query that doesn't reflect what `getDistinctRequired()` and `getProjectionVars()` return.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -10,11 +16,15 @@ public class SPARQLRequestImpl implements SPARQLRequest
...
 	protected final ExpectedVariables expectedVars;
+	protected final Set<Var> projectionVars;
+	protected final boolean isDistinct;
 
 	public SPARQLRequestImpl( final SPARQLGraphPattern pattern ) {
 		assert pattern != null;
 		this.pattern = pattern;
 		this.query = null;
+		this.projectionVars = Collections.emptySet();
+		this.isDistinct = false;
```
hartig: It should be possible to provide these via arguments of the constructor.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -25,6 +35,22 @@ public SPARQLRequestImpl( final SPARQLQuery query ) {
...
+		this.isDistinct = query.isDistinct();
+
+		// we materialize the ExpectedVariables in this case
+		// to avoid producing it again whenever it is used
+		expectedVars = query.getExpectedVariables();
+	}
+
+	public SPARQLRequestImpl( final SPARQLQuery query,
+	                          final Set<Var> projectionVars,
+	                          final boolean distinct ) {
```
hartig: There is no need for such a constructor. It would only be confusing if the given `projectionVars` and `distinct` are different from what the given `query` has.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,27 +63,36 @@ public boolean equals( final Object o ) {
...
-		if ( pattern == null ) {
-			return oo.getQueryPattern() == null && query.equals( oo.getQuery() );
-		}
-		else {
-			return pattern.equals( oo.getQueryPattern() );
-		}
+
+		return Objects.equals( getQueryPattern(), oo.getQueryPattern() )
+			&& getProjectionVars().equals( oo.getProjectionVars() )
+			&& isDistinct() == oo.isDistinct();
```
hartig: This is not correct for cases in which both `oo` and `this` have their respective `pattern == null` but have different `query` members.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,27 +63,36 @@ public boolean equals( final Object o ) {
...
-	public int hashCode(){
-		if ( pattern == null )
-			return query.hashCode();
-		else
-			return pattern.hashCode();
+	public int hashCode() {
+		return Objects.hash(
+			getQueryPattern(),
+			getProjectionVars(),
+			isDistinct()
```
hartig: Same issue here as for the `equals` method.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
```diff
@@ -38,4 +43,14 @@ public String toString(){
 		return bgp.toString();
 	}
 
+	@Override
+	public Set<Var> getProjectionVars() {
+		return Collections.emptySet();
```
hartig: As mentioned in my comment n this method in `SPARQLRequest` interface, this should be:
```suggestion
		return null;
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/TriplePatternRequestImpl.java
```diff
@@ -41,4 +45,14 @@ public String toString(){
 		return tp.toString();
 	}
 
+	@Override
+	public Set<Var> getProjectionVars() {
+		return Collections.emptySet();
```
hartig: As mentioned in my comment n this method in `SPARQLRequest` interface, this should be:
```suggestion
		return null;
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImpl.java
```diff
@@ -57,12 +57,32 @@ protected SolMapsResponse performRequestWithQueryExecutionHTTP( final SPARQLRequ
...
+			q.setQueryResultStar(false);
+			q.getProject().clear();
+			req.getProjectionVars().forEach(q::addResultVar);
+		}
+
+		// Apply DISTINCT if requested.
+		// This enforces duplicate elimination at the endpoint level.
+		if ( req.isDistinct() )
+			q.setDistinct( true );
+
```
hartig: These changes should not be necessary if, instead, you extend the `default` implementation of `getQuery()` in the `SPARQLRequest` interface, as mentioned in my comment there. In fact, the code here can be used for that implementation.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImpl.java
```diff
@@ -141,7 +161,43 @@ protected static class MySolutionConsumer implements Consumer<QuerySolution>
...
+	 *   <li>GROUP BY (projection affects grouping semantics)</li>
+	 *   <li>HAVING clauses (depends on grouped results)</li>
+	 *   <li>Aggregators (e.g., COUNT, SUM), which rely on specific projection expressions</li>
+	 * </ul>
+	 * </p>
+	 *
+	 * @param q the query to inspect
+	 * @return {@code true} if projection can be safely overridden; {@code false} otherwise
+	 */
+	protected static boolean isSafeToOverrideProjection(final Query q)
```
hartig: This one would also not be needed because, I believe, there needs to be no case in which the projection and the distinct _of a given_ `Query` object would be overridden.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/query/SPARQLQuery.java
```diff
@@ -14,6 +14,27 @@ public interface SPARQLQuery extends Query
...
+	/**
+	 * Indicates whether this query enforces duplicate elimination.
+	 *
+	 * <p>This corresponds to the presence of the {@code DISTINCT} modifier
+	 * in the SELECT clause.</p>
+	 *
+	 * @return {@code true} if the query is DISTINCT; {@code false} otherwise
+	 */
+	boolean isDistinct();
+
```
hartig: I don't think these are needed. If we have code that uses this information, it can pick it up directly from the `Query` object returned by `asJenaQuery()`.

hartig (changes requested):

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
KentThang: I am making a new constructor instead of changing this one!

kentthang010:
I am a little bit confused so I am marking this as a draft for now. The latest commit moves the projection/distinct logic from the `SPARQLRequestProcessorImpl` to the `getQuery()` method of `SPARQLRequestImpl`. The reason that it's in there instead of the default method of `SPARQLRequest`, is because SPARQLRequestImpl is used to initialize the requests in the tests and therefore the `getQuery()` method of `SPARQLRequestImpl` is used instead. You mean that it should instead be in the default `getQuery()` of SPARQLRequest?

I also kept `isSafeToOverrideProjection` because it keeps this logic in `getQuery()` from incorrectly overwriting the query in the `performRequestWithQueryOnDBpedia` test but as I understand your comment, this is not necessary if I were to implement this correctly?

Please check the latest commit when you have the time!

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
```diff
@@ -14,6 +18,28 @@ public interface SPARQLRequest extends DataRetrievalRequest
...
+	 * be respected by components that can support it.</p>
+	 *
+	 * <p>Deciding whether such a projection can be safely applied or pushed
+	 * into a request is the responsibility of the query planning and
+	 * rewriting logic, not of this interface or its implementations.</p>
+	 */
+	Set<Var> getProjectionVars();
+
+	/**
+	 * Returns {@code true} if this request <em>explicit</em> requires that the requested
```
hartig: Typo. My bad.
```suggestion
	 * Returns {@code true} if this request <em>explicitly</em> requires that the requested
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
hartig: The `default` implementation of `getQuery()` here needs to be extended to take `getProjectionVars()` and `isDistinct()` into account. As a consequence, the `convertToQuery` function here needs to be extended as well, and so does the corresponding constructor of `SPARQLQueryImpl`.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/query/impl/SPARQLQueryImpl.java
```diff
@@ -23,10 +23,10 @@ public SPARQLQueryImpl( final SPARQLGraphPattern p ) {
 
 	protected SPARQLQueryImpl( final Element jenaElement ) {
```
hartig: Related to my comment about the `default` implementation of `getQuery()` in `SPARQLRequest`, you need to extend these two constructors with two more arguments: one for a set of projection variables or `null` and one for `isDistinct`.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/query/impl/SPARQLQueryImpl.java
hartig: While looking at this file, I notice that this method can be improved. Please change it as follows.
```
		if ( o == this ) return true;

		return o instanceof SPARQLQuery oo && oo.asJenaQuery().equals(jenaQuery);
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -1,20 +1,45 @@
...
 		assert pattern != null;
 		this.pattern = pattern;
 		this.query = null;
+		this.projectionVars = null;
+		this.isDistinct = false;
+
+		// we materialize the ExpectedVariables in this case
+		// to avoid producing it again whenever it is used
+		expectedVars = pattern.getExpectedVariables();
+	}
```
hartig, line 27:
```suggestion
	public SPARQLRequestImpl( final SPARQLGraphPattern pattern ) {
		this(pattern, null, false);
	}
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -25,6 +50,8 @@ public SPARQLRequestImpl( final SPARQLQuery query ) {
 		assert query != null;
 		this.query = query;
 		this.pattern = null;
+		this.projectionVars = new HashSet<>( query.asJenaQuery().getProjectVars() );
```
hartig: To be consistent with the contract for `getProjectionVars()`, you need to check whether the given query has a `*` instead of projection variables. I believe you can use a function called `getQueryResultStar()` (or similar) for this purpose. If it is a `*`, then `projectionVars` must be initialized to `null`.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,33 +64,69 @@ public boolean equals( final Object o ) {
 			return false;
 
 		final SPARQLRequest oo = (SPARQLRequest) o;
+
 		if ( pattern == null ) {
-			return oo.getQueryPattern() == null && query.equals( oo.getQuery() );
+			return oo.getQueryPattern() == null
+			    && query.equals( oo.getQuery() )
+			    && getProjectionVars().equals(oo.getProjectionVars())
```
hartig: Since any of the two `getProjectionVars()` methods may return `null`, you must do this comparison using: `Objects.equals( projectionVars, oo.getProjectionVars() )`

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,33 +64,69 @@ public boolean equals( final Object o ) {
...
+			    && isDistinct() == oo.isDistinct();
 		}
 		else {
-			return pattern.equals( oo.getQueryPattern() );
+			return pattern.equals( oo.getQueryPattern() )
+			    && getProjectionVars().equals(oo.getProjectionVars())
```
hartig: Also here, the comparison must be done by: `Objects.equals( projectionVars, oo.getProjectionVars() )`

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,33 +64,69 @@ public boolean equals( final Object o ) {
...
+			    && isDistinct() == oo.isDistinct();
 		}
 	}
 
 	@Override
-	public int hashCode(){
+	public int hashCode() {
 		if ( pattern == null )
-			return query.hashCode();
+			return query.hashCode() ^ getProjectionVars().hashCode() ^ (isDistinct() ? 1 : 0);
```
hartig: In case that `query` has been initialized, `query` has anything needed for the distinction. Hence, you can change this one back to:

```suggestion
			return query.hashCode();
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,33 +64,69 @@ public boolean equals( final Object o ) {
...
 		else
-			return pattern.hashCode();
+			return pattern.hashCode() ^ getProjectionVars().hashCode() ^ (isDistinct() ? 1 : 0);
```
hartig: Include the `hashCode` of `getProjectionVars()` only if `getProjectionVars() != null`

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -1,20 +1,45 @@
...
+import se.liu.ida.hefquin.base.query.impl.SPARQLQueryImpl;
 import se.liu.ida.hefquin.federation.access.SPARQLRequest;
 
 public class SPARQLRequestImpl implements SPARQLRequest
 {
 	protected final SPARQLGraphPattern pattern;
 	protected final SPARQLQuery query;
 	protected final ExpectedVariables expectedVars;
+	protected final Set<Var> projectionVars;
+	protected final boolean isDistinct;
```
hartig: Logically, the two new member variables belong to `pattern` and the other member variables are each somehow separate from the other ones. Hence:

```suggestion
	protected final SPARQLGraphPattern pattern;
	protected final Set<Var> projectionVars;
	protected final boolean isDistinct;

	protected final SPARQLQuery query;

	protected final ExpectedVariables expectedVars;
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,33 +64,69 @@ public boolean equals( final Object o ) {
...
+	public boolean isDistinct() {
+		return isDistinct;
+	}
+
 	@Override
 	public SPARQLQuery getQuery() {
-		if ( query != null )
-			return query;
-		else
-			return SPARQLRequest.convertToQuery( getQueryPattern() );
```
hartig, line 66: After you have extended `convertToQuery` with the two additional arguments (as mentioned in my comment there), you go back the original implementation here, with the addition of passing `getProjectionVars()` and `isDistinct()` as argument to `convertToQuery`.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -79,4 +142,40 @@ public String toString(){
 			return "SPARQLRequest with pattern: " + pattern.toString();
```
hartig: Extend this one to include the projection variables (be careful about `null`) is `isDistinct`.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/TriplePatternRequestImpl.java
```diff
@@ -4,7 +4,11 @@
 import se.liu.ida.hefquin.base.query.TriplePattern;
 import se.liu.ida.hefquin.federation.access.TriplePatternRequest;
 
+import java.util.Collections;
```
hartig: Please double check whether this import is needed.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/BGPRequestImpl.java
```diff
@@ -1,5 +1,10 @@
 package se.liu.ida.hefquin.federation.access.impl.req;
 
+import java.util.Collections;
```
hartig: Is this import needed?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImpl.java
```diff
@@ -141,7 +141,43 @@ protected static class MySolutionConsumer implements Consumer<QuerySolution>
...
```
hartig: I still think that this is not needed.

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java
```diff
@@ -9,10 +9,10 @@
 public abstract class FederationTestBase
 {
 	/**
-	 * Change this flag to true if you also want to run the
-	 * unit tests that access servers on the actual Web.
+	 * If this flag is true, tests that access servers on the actual
+	 * Web will be skipped.
 	 */
-	public static boolean skipLiveWebTests = true;
+	public static boolean skipLiveWebTests = false;
```
hartig: Please revert these changes back (including the Javadoc comment).

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImplTest.java
```diff
@@ -108,6 +112,100 @@ public void performRequestWithQueryOnDBpedia() throws FederationAccessException
...
+			final SolMapsResponse resp1 = recProc.performRequest(req1, fm);
+			final SolMapsResponse resp2 = recProc.performRequest(req2, fm);
+
+			// checking
+			// the base query does not contain DISTINCT
+			assertFalse( baseQuery.asJenaQuery().isDistinct() );
+
+			// req1 inherits DISTINCT=false, req2 explicitly enforces DISTINCT
+			assertFalse( req1.isDistinct() );
+			assertTrue( req2.isDistinct() );
```
hartig: These tests are actually testing the functionality of `SPARQLRequestImpl` and not the one of `SPARQLRequestProcessorImpl`. In this sense, it would indeed be good if you added `SPARQLRequestImplTest` with it (which can then also be run with `skipLiveWebTests == true`).

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImplTest.java
```diff
@@ -108,6 +112,100 @@ public void performRequestWithQueryOnDBpedia() throws FederationAccessException
...
+			final SPARQLEndpoint fm = new SPARQLEndpointForTest("http://dbpedia.org/sparql");
+
+			final SPARQLRequestProcessor recProc = new SPARQLRequestProcessorImpl();
+
+			final SolMapsResponse resp1 = recProc.performRequest(req1, fm);
+			final SolMapsResponse resp2 = recProc.performRequest(req2, fm);
+
+			// checking
+			// ensure projection is applied
+			for ( SolutionMapping sm : resp2.getResponseData() ) {
```
hartig:
```suggestion
			for ( final SolutionMapping sm : resp2.getResponseData() ) {
```

hartig (changes requested):

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java
KentThang: But is the Javadoc comment correct though? From what I have observed in `SPARQLRequestProcessorImplTest,` the opposite is true. That is to say, when it is set to true, the tests in `SPARQLRequestProcessorImplTest` are not run which contradicts the Javadoc comment.

Beyond that, the current naming suggests that true means “skip”, which aligns with the implementation but not the Javadoc. Changing the comment would also make it consistent with the other flags `skipLocalNeo4jTests` and `skipLocalGraphQLTests`.

Either the Javadoc comment needs to be changed or the variable name and flag usage (in `SPARQLRequestProcessorImplTest` needs to be changed (or I've misunderstood something).

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImpl.java
KentThang: You were right, resolving the other PR comments has led to this method being unnecessary!

KentThang:
Latest commit resolves the comments above, except for the one regarding the Javadoc comment. See my reply above!

hefquin-base/src/main/java/se/liu/ida/hefquin/base/query/impl/SPARQLQueryImpl.java
```diff
@@ -17,21 +20,36 @@ public SPARQLQueryImpl( final Query jenaQuery ) {
...
+
+		jenaQuery.setDistinct( isDistinct );
 		jenaQuery.setQueryPattern( jenaElement );
 	}
 
 	@Override
 	public boolean equals( final Object o ) {
+		if ( o == this ) return true;
+
 		return o instanceof SPARQLQuery && ((SPARQLQuery) o).asJenaQuery().equals(jenaQuery);
```
hartig:
```suggestion
		return o instanceof SPARQLQuery oo && oo.asJenaQuery().equals(jenaQuery);
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/SPARQLRequest.java
```diff
@@ -14,16 +18,40 @@ public interface SPARQLRequest extends DataRetrievalRequest
...
+	 */
+	Set<Var> getProjectionVars();
+
+	/**
+	 * Returns {@code true} if this request <em>explicitly</em> requires that the requested
+	 * result is duplicate free.
+	 *
+	 * @return {@code true} if duplicate elimination is requested; {@code false} otherwise
+	 */
+	boolean isDistinct();
```
hartig: A better name for this one is `getDistinctRequired()`

(Sorry, I made this comment already during my previous review, but it somehow didn't become part of the actual review in the end.)

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,33 +60,51 @@ public boolean equals( final Object o ) {
 			return false;
 
 		final SPARQLRequest oo = (SPARQLRequest) o;
+
 		if ( pattern == null ) {
-			return oo.getQueryPattern() == null && query.equals( oo.getQuery() );
+			return oo.getQueryPattern() == null
+			    && query.equals( oo.getQuery() )
+			    && Objects.equals( getProjectionVars(), (oo.getProjectionVars()) )
+			    && isDistinct() == oo.isDistinct();
```
hartig: This can be simplified to how it was before, because the `query` has all relevant information to check for equality.
```suggestion
			return oo.getQueryPattern() == null && query.equals( oo.getQuery() );
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -37,33 +60,51 @@ public boolean equals( final Object o ) {
...
 		}
 		else {
-			return pattern.equals( oo.getQueryPattern() );
+			return pattern.equals( oo.getQueryPattern() )
+			    && Objects.equals( getProjectionVars(), (oo.getProjectionVars()) )
```
hartig:
```suggestion
			    && Objects.equals( getProjectionVars(), oo.getProjectionVars() )
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -74,9 +115,12 @@ public ExpectedVariables getExpectedVariables() {
 	@Override
 	public String toString(){
 		if ( query != null )
-			return "SPARQLRequest with query: " + query.toString();
+			return "SPARQLRequest with query: " + query.toString()
+			     + ", projection variables: " + ( projectionVars == null ? "null" : projectionVars.toString() )
+			     + ", distinct: " + isDistinct;
```
hartig: Also, here, the `query` should have everything that is needed.
```suggestion
			return "SPARQLRequest with query: " + query.toString();
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImpl.java
```diff
@@ -74,9 +115,12 @@ public ExpectedVariables getExpectedVariables() {
...
 		else
-			return "SPARQLRequest with pattern: " + pattern.toString();
+			return "SPARQLRequest with pattern: " + pattern.toString()
+			     + ", projection variables: " + ( projectionVars == null ? "null" : projectionVars.toString() )
```
hartig: Actually, the following will do exactly the same (but is shorter).
```suggestion
			     + ", projection variables: " + projectionVars
```

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java
hartig: You are totally right. Sorry!

Now, please change the Javadoc comment back but replace "true" by "false".

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImplTest.java
hartig: I see that you misunderstood my comment here. My comment was not about the whole test method but only about these `assertFalse` and `assertTrue` statements. In other words, I did _not_ mean that you should completely move the two new test methods from `SPARQLRequestProcessorImplTest` to `SPARQLRequestImplTest`. Please move them back where you had them before, and then create new test methods in `SPARQLRequestImplTest` that test only the new functionality of `SPARQLRequestImpl`. Hence, these new methods do not need to run actual queries; they only need to test that the constructor of `SPARQLRequestImpl` that has a `SPARQLQuery` as argument does the right thing in terms of initializing `projectionVars` and `isDistinct`.

hartig (changes requested):
We are getting there ;-)

KentThang:
Comments have been resolved! The tests that I added back to `SPARQLRequestProcessorImplTest` are unchanged (so they include the initialization correctness checks) while the tests in `SPARQLRequestImplTest` are the original tests but without the request and request checks.

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/req/SPARQLRequestImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+import se.liu.ida.hefquin.base.query.impl.SPARQLQueryImpl;
+import se.liu.ida.hefquin.federation.FederationTestBase;
+import se.liu.ida.hefquin.federation.access.FederationAccessException;
+import se.liu.ida.hefquin.federation.access.SPARQLRequest;
+
+public class SPARQLRequestImplTest extends FederationTestBase
+{
+	@Test
+	public void testDistinctRequestInitialization() throws FederationAccessException {
+		if ( ! skipLiveWebTests ) {
```
hartig: This check should be removed (both here and also in the other unit test in this file).

hartig (changes requested):
One last issue to be fixed.

hartig (approved):
Now this PR is ready. I will merge it, and then you can go ahead with adapting the rewriting rules (i.e., the second part of #570).

PR #601, opened 2026-04-29
Add MaterializeRDFViewFromRML

### LINKED ISSUE #598: CLI for executing RML mappings
Name the main Java class: `MaterializeRDFViewFromRML`
As a summary of our chat yesterday, here is a laundry list of the bits and pieces needed:
* Probably the most similar CLI tool is `MaterializeRDFViewOfLPG`
* The tool should have one mandatory argument: an RDF file (in any possible format), which is assumed to contain an RML mapping consisting of one or more triples maps.
* Two optional arguments: one to specify the file into which the resulting RDF data should be written (if omitted, then write to stdout) and one to specify a base URI to be used for the mapping process (if omitted, create a hardcoded one like "http://example.org/FixedBaseIRI/HardcodedInMaterializeRDFViewFromRML /").
* Read the RDF file into a Jena `Model` object (see the first `parseFedDescr` in `FederationDescriptionReader`).
* Find all `Resource` objects in the `Model` that are of rdf:type rml:TriplesMap; use the `listResourcesWithProperty` method of `Model` for that; as the `Property` for rdf:type, use the one in `org.apache.jena.vocabulary.RDF` and, for rml:TriplesMap, use the one in `se.liu.ida.hefquin.mappings.rml.vocabulary.RMLVocab`.
* Convert each of these triples map `Resource` objects into a `MappingExpression` by using `RML2MappingAlgebra.convert`, similar to how it is done in lines 291-307 of `FederationDescriptionReader`.
* Put a `MappingOpProject` on top of each of the mapping expressions and, then, combine them all via a `MappingOpUnion`, as done in `createWrappedRESTEndpoint` of `FederationDescriptionReader`.
* The resulting `MappingExpression` needs to be converted into a physical mapping plan now. To this end, use `evaluate` of `MappingRelationUtils`, which returns a `MappingRelation` that implicitly represents the physical plan. The second argument of `evaluate` is a `Map`. To create it, do the following:
   * Load a JSON file from a hard-coded location.
   * Create a `JsonObject` using a `String` that contains the content of the JSON file.
   * Extract every `SourceReference` from the `MappingExpression` to be converted, for which you can use `extractAllSrcRefs` of `MappingRelationUtils`.
   * Create a `Map` that maps each such `SourceReference` to the `JsonObject`.
* Now the `MappingRelation` needs to be consumed and converted into an RDF dataset, which can be done via `convertToRDF` of `MappingRelationUtils`. The result is a Jena `Dataset`.
* The default graph of this `Dataset` can be accessed as a Jena `Model` by using the `getDefaultModel()` method of the `Dataset`.
* This default graph `Model` needs to be written. Use the `write(OutputStream out, Model model, Lang lang)` method of Jena's `RDFDataMgr` for that.

kentthang010:
Closes #598.

Not sure if I should've included `sources.json,` the hardcoded `.json` file, I ended up excluding it.


hartig:
> Not sure if I should've included `sources.json,` the hardcoded `.json` file, I ended up excluding it.

Please add the file in the examples directory and adapt the code such that it refers to the file there.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
hartig, line 1: Please try to avoid such "non-changing" changes  in the future, just for the sake of keeping the PRs focused on the relevant stuff.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+	protected final ModTime modTime =            new ModTime();
+	protected final ModLangOutput modLangOut =   new ModLangOutput();
+
+	protected final ArgDecl argRdfFile =      new ArgDecl( ArgDecl.HasValue, "mapping" );
+	protected final ArgDecl argOutputToFile = new ArgDecl( ArgDecl.HasValue, "output" );
+	protected final ArgDecl argBaseIRI =      new ArgDecl( ArgDecl.HasValue, "baseIRI" );
+
+	/**
+	 * Main entry point of the tool, accepting command-line arguments to specify the
+	 * Neo4j connection details and output format options.
```
hartig: There is a copy&paste error here ("Neo4j connection").

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+	 * optional arguments.
+	 *
+	 * @return A string that describes the usage of the command.
+	 */
+	@Override
+	protected String getSummary() {
+		return "Usage: " + getCommandName() + " " +
+			"--mapping=<rdf-file> " +
+			"--output=<file-name> " +
+			"--baseIRI=<iri>";
```
hartig:
```suggestion
			"[--output=<file-name>] " +
			"[--baseIRI=<iri>]";
```
To indicate that these two arguments are optional.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+	 */
+	protected MaterializeRDFViewFromRML( final String[] argv ) {
+		super(argv);
+
+		addModule(modTime);
+		addModule(modLangOut);
+
+		add( argRdfFile, "--mapping", "RML mapping file" );
+		add( argOutputToFile, "--output", "Output file" );
+		add( argBaseIRI, "--baseIRI", "Base IRI for mapping" );
```
hartig:
```suggestion
		add( argOutputToFile, "--output", "Output file (optional, printing to stdout if omitted)" );
		add( argBaseIRI, "--baseIRI", "Base IRI for mapping (optional, hardcoded IRI used if omitted)" );
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+
+		final Node chosenBaseIRI =
+			contains(argBaseIRI) ? NodeFactory.createURI( getValue( argBaseIRI ) ) :
+		                           NodeFactory.createURI( "http://example.org/FixedBaseIRI/HardcodedInMaterializeRDFViewFromRML/" );
+
+		final ResIterator iter = rdfModel.listResourcesWithProperty( RDF.type, RMLVocab.TriplesMap );
+		final List<MappingExpression> trMaps = new ArrayList<>();
+		while ( iter.hasNext() ) {
+			final Resource tm = iter.next();
+			if ( tm.isResource() ) {
```
hartig: In this case, there is no need for the if-check because `tm` is a `Resource` by construction.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+				final MappingExpression trMap;
+				try {
+					trMap = ( RML2MappingAlgebra.convert( tm,
```
hartig: No need for the double parentheses.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+
+		final MappingExpression expr = MappingExpressionFactory.create(
+				MappingOpUnion.getInstance(),
+				exprs );
+
+		// NOTE: currently using a fixed JSON source for evaluation.
+		final String jsonString;
+		try {
+			jsonString = Files.readString(Path.of("sources.json"));
+		} catch ( Exception e ) {
```
hartig:
```suggestion
		}
		catch ( final Exception e ) {
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+			cmdError( "Failed to read sources.json: " + e.getMessage(), true );
+			return; // Primarily used to avoid "variable not initialized" compiler error
+		}
+
+		final JsonObject jsonObject = new JsonObject(jsonString);
```
hartig: Doesn't this constructor throw an exception as well?

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,202 @@
...
+
+		final Dataset dataset = MappingRelationUtils.convertToRDF(mappingRelation);
+
+		OutputStream outputStream = System.out;
+		if ( contains(argOutputToFile) ) {
+			try {
+				// Appends to file rather than overwriting
+				outputStream = new FileOutputStream( getValue( argOutputToFile ), true );
+			} catch ( final FileNotFoundException e ) {
+				cmdError( "Failed to create print stream for output destination: " + getValue( argOutputToFile ), false );
```
hartig: This error should be raised earlier, before attempting to evaluate the mapping. In fact, it is better to move the actual functionality into a helper method which is called after the arguments of the tool have been processed. Hence, the OutputStream should have been set successfully when that helper method is called.

hartig (changes requested):
Good job! Just a few smaller things to be improved.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
KentThang: Noted!

kentthang010:
PR comments have been resolved! I also realized I had to change the `output` flag name to something different (ended up with using `outputToFile`) as it clashed with the `output` flag of `ModLangOut`.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,224 @@
...
+		catch ( Exception e ) {
+			cmdError( "Failed to read sources.json: " + e.getMessage(), true );
+			return; // Primarily used to avoid "variable not initialized" compiler error
+		}
+
+		final JsonObject jsonObject;
+		try {
+			jsonObject = new JsonObject(jsonString);
+		}
+		catch ( JsonPathException e ) {
```
hartig:
```suggestion
		catch ( final JsonPathException e ) {
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -0,0 +1,224 @@
...
+		while ( iter.hasNext() ) {
+			final Resource tm = iter.next();
+			final MappingExpression trMap;
+			try {
+				trMap = RML2MappingAlgebra.convert( tm,
+				                                    rdfModel,
+				                                    chosenBaseIRI );
+			}
+			catch ( final RMLParserException e ) {
+				throw new IllegalArgumentException("There is a problem in the RML mapping: " +  e.getMessage(), e );
```
hartig: It is better to terminate here via `cmdError`. The message can be the same, but it would make the program print this message as an error message rather than printing the exception with its stack trace (which always looks a bit ugly for users).

hartig (changes requested):
My previous comment about the helper method was actually meant the other way around, not to create a helper method for setting up the `outputStream` argument but for performing the actual functionality after the relevant input variables for the functionality have been initialized. So, concretely, create a helper method hat may also be called `exec` but that takes three arguments: `rmlDescr` (the `Model`), `baseIRI`, and `out` (the `OutputStream`).

hartig (approved):
This looks good now! I will merge it and write another issue about extending the functionality of the tool.

Issue #594, opened 2026-04-23
Try to keep query planning info if possible


hartig:
Since PR #584, `RemoveSubPlansWithEmptyResults` performs the very first rewriting of logical plans, and it runs the cardinality estimator to do so. As a result, after applying this rewriting rule, all subplans have a `QueryPlanningInfo` object with a cardinality estimate. Most (if not all?) of the other rewriters, whenever they fire and do create a new subplan, they simply ignore the `QueryPlanningInfo` object of the original subplan and pass `null` for the `QueryPlanningInfo` argument of the subplans that they create. In some cases, it is indeed necessary to do so because the new subplans may have a different cardinality estimate, but there are also cases in which it is safe to copy over the `QueryPlanningInfo` object from a original subplan to the newly-created subplan. For instance, when pushing a project operator under a bind operator, the resulting plan that now has the bind operator as root is guaranteed to produce the exact same result (with the same cardinality) as the original plan and, thus, the resulting plan can be created with a `QueryPlanningInfo` object that is copied from the `QueryPlanningInfo` object of the original plan.

This issue is about going through all the rewriters (implementations of `HeuristicForLogicalOptimization` and look for cases in which the query planning info can be preserved. For some of the rewriters, no change is needed because they do not contain such cases. For every rewriter that contains such cases, create a PR with the changes. The rewriters to look at are:
* [ ] RemoveSubPlansWithEmptyResult
* [x] FilterPushDown
* [ ] ProjectPushDown (once the PR for it is merged)
* [x] MergeRequests
* [ ] PullUpLtgOverJoin
* [ ] PullUpLtgOverUnion
* [ ] RemovePairsOfG2lAndL2g
* [ ] RemoveUnnecessaryL2gAndG2l
* [ ] PushJoinUnderUnionWithRequests
* [ ] UnionPullUp
* [ ] ApplyVocabularyMappings

kentthang010:
I believe no changes need to be made in `RemoveSubPlansWithEmptyResult`. I briefly considered changing this:
```
if ( rewrittenSubPlans.size() > 1 )
	return LogicalPlanUtils.createPlanWithSubPlans(rootOp, null, rewrittenSubPlans);
```
(rows 137-138)
from `null` to `inputPlan.getQueryPlanningInfo().getProperties()` with the reasoning that removing empty subplans should not change the cardinality of the root union operator but I am not sure if this is safe, even if makes sense "mathematically".

[...]

PR #603, opened 2026-05-04
Keep QP info for various cases in ProjectPushDown

kentthang010:
Addresses the `ProjectPushDown` part of #594.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
@@ -145,7 +145,9 @@ public class ProjectPushDown implements HeuristicForLogicalOptimization
 
 		@Override
 		public void visit( final LogicalOpFixedSolMap op ) {
-			createdPlan = createPlanForFixedSolMapUnderProject(projectOp, op, inputPlan);
+			createdPlan = createPlanForFixedSolMapUnderProject( projectOp,
+			                                                    op,
+			                                                    inputPlan );
 		}
 
 		@Override
@@ -229,37 +231,38 @@ public class ProjectPushDown implements HeuristicForLogicalOptimization
 
 		@Override
 		public void visit( final LogicalOpLocalToGlobal op ) {
-			createdPlan = createPlanForUnaryOpUnderProject( projectOp,
-			                                                op,
-			                                                subPlanUnderProject.getSubPlan(0) );
+			createdPlan = createPlanForUnaryOpUnderProjectDropQP( projectOp,
+			                                                      op,
+			                                                      subPlanUnderProject.getSubPlan(0) );
 		}
 
 		@Override
 		public void visit( final LogicalOpGlobalToLocal op ) {
 			createdPlan = createPlanForL2GOrG2LUnderProject( projectOp,
-			                                                op,
-			                                                subPlanUnderProject.getSubPlan(0) );
+			                                                 op,
+			                                                 subPlanUnderProject.getSubPlan(0) );
 		}
 
 		@Override
 		public void visit( final LogicalOpDedup op ) {
-			createdPlan = createPlanForUnaryOpUnderProject( projectOp,
-			                                                op,
-			                                                subPlanUnderProject.getSubPlan(0) );
+			createdPlan = createPlanForUnaryOpUnderProjectDropQP( projectOp,
+			                                                      op,
+			                                                      subPlanUnderProject.getSubPlan(0) );
 		}
 
 		@Override
 		public void visit( final LogicalOpProject op ) {
 			createdPlan = createPlanForProjectUnderProject( projectOp,
 			                                                op,
+			                                                subPlanUnderProject,
 			                                                subPlanUnderProject.getSubPlan(0) );
 		}
 
 		@Override
 		public void visit( final LogicalOpMinus op ) {
 			createdPlan = createPlanForJoinLikeOpUnderProject( projectOp,
-			                                                 subPlanUnderProject,
-			                                                 inputPlan );
+			                                                   subPlanUnderProject,
+			                                                   inputPlan );
 		}
 
 	} // end of Worker
@@ -357,6 +360,10 @@ public class ProjectPushDown implements HeuristicForLogicalOptimization
 	 *
 	 * Otherwise, the project is split: the adjusted projection is pushed below
 	 * the bind, while the original projection remains on top.
+	 *
+	 * Query planning information is preserved for the bind subtree because this
+	 * transformation is cardinality-preserving and does not change the number of
+	 * solution mappings produced by the bind operator.
 	 */
 	protected LogicalPlan createPlanForBindUnderProject( final LogicalOpProject projectOp,
 	                                                     final LogicalOpBind bindOp,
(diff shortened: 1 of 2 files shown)


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -229,37 +231,38 @@ public void visit( final LogicalOpUnfold op ) {
 
 		@Override
 		public void visit( final LogicalOpLocalToGlobal op ) {
-			createdPlan = createPlanForUnaryOpUnderProject( projectOp,
-			                                                op,
-			                                                subPlanUnderProject.getSubPlan(0) );
+			createdPlan = createPlanForUnaryOpUnderProjectDropQP( projectOp,
```
hartig: I think we made a mistake in one of the other recent PRs. This method should call `createPlanForL2GOrG2LUnderProject` instead of `createPlanForUnaryOpUnderProjectDropQP`

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -677,32 +700,50 @@ protected LogicalPlan createPlanForJoinLikeOpUnderProject( final LogicalOpProjec
...
+	 * The parent and child projections are combined by computing the intersection of their
+	 * variable sets. The resulting projection is applied above the child subplan.
+	 *
+	 * If the parent projection does not further restrict the child projection (i.e., the
+	 * intersection equals the child's projection), the parent project is redundant and is
+	 * removed, leaving the child project as the new root.
+	 *
+	 * Query planning information from the child subtree is preserved in the rewritten plan
+	 * since this transformation is cardinality-preserving and does not affect the structure
+	 * or size of the underlying solution mappings.
```
hartig: I don't think that this is true in all cases, especially when considering the `mayReduce` flag.

For instance, assume the child project operator has a projection variable, say ?x, that is not in the set of projection variables of the parent project operator. Then, there may be a case in which the child project operator outputs two solution mappings that differ only regarding the binding that they have for ?x. After projecting away they respective binding for ?x in the parent project operator, the two solution mappings become a duplicate of one another. If the parent project operator has the `mayReduce` flag set to `true`, the result cardinality would be affected and it would be incorrect to keep the qpInfo of the child project for the resulting merged project operator.

Any extreme case of the above example is if the intersection of the two sets of projection variables is empty. In this case, the resulting merged project operator has the empty set of projection variables. Then, with `mayReduce` set to `true`, that merged project operator may return only a single output solution mapping (which would be the empty solution mapping). Only if `mayReduce == false`, the duplicates are guaranteed to be preserved and, thus, the output cardinality is the same as the input cardinality.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -677,32 +700,50 @@ protected LogicalPlan createPlanForJoinLikeOpUnderProject( final LogicalOpProjec
...
+		// If the parent projection does not further restrict the variables,
+		// it is redundant and can be removed. Continue optimization from the child project.
+		if ( intersectionOfVars.equals(childProjectOp.getVariables()) ) {
+			return apply(subPlanUnderParentProjectOp.getSubPlan(0));
+		}
+
+		final LogicalPlan rewritten = new LogicalPlanWithUnaryRootImpl(
 			new LogicalOpProject(intersectionOfVars, parentProjectOp.mayReduce()),
-			null,
+			subPlanUnderChildProjectOp.getQueryPlanningInfo().getProperties(),
```
hartig: Here is the issue pointed out by my previous comment (in the Javadoc of this method). You can copy qpInfo, but only if `parentProjectOp.mayReduce() == false`; otherwise the qpInfo needs to be dropped.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -419,6 +428,10 @@ protected LogicalPlan createPlanForBindUnderProject( final LogicalOpProject proj
 	 *
 	 * Otherwise, the project is split: the adjusted projection is pushed below
 	 * the unfold, while the original projection remains on top.
+	 *
+	 * Query planning information is preserved for the unfold subtree under the
+	 * assumption that unfold is cardinality-preserving in the same way as bind,
+	 * i.e., it does not change the number of solution mappings, only their shape.
```
hartig: The assumption is incorrect. UNFOLD is not cardinality preserving; see: https://awslabs.github.io/SPARQL-CDTs/spec/latest.html#description-of-unfold

Having said that, is is still possible to preserve qpInfo, but only if the `mayReduce` flag is not set.

hartig (changes requested):
I didn't manage to go through everything, but I added some comments. After addressing them, please revisit the other cases (for which I don't have comments because I didn't manage to look at them) and think about them once more in the light of my comments about the `mayReduce` flag. Some of them might also be affected by the `mayReduce` flag (or may be not--just double check).

KentThang:
Latest commit resolves the comments above.

I also went over the other cases and considered the `mayReduce` flag.

In summary, the pushdowns that now consider `mayReduce` are:
- `createPlanForUnaryOpUnderProjectPreserveQP`
- Case 1 (full pushdown) for `createPlanForAddOpUnderProject`
- `createPlanForProjectUnderProject`
- `createPlanForUnfoldUnderProject`

I don't think I have to check mayReduce for `createPlanForBindUnderProject`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -713,6 +759,35 @@ protected LogicalPlan createPlanForUnaryOpUnderProject( final LogicalOpProject p
...
+	 * given unary operator, with the subplan underneath being a version of
+	 * the given subplan in which the project push down heuristic has been
+	 * applied recursively.
+	 *
+	 * The resulting plan preserves the query planning info of the input plan
+	 * only if the project operator is duplicate-preserving (i.e., {@code mayReduce == false});
+	 * otherwise, the query planning info must be dropped since pushing the projection
+	 * may affect result cardinality.
+	 */
+	protected LogicalPlan createPlanForUnaryOpUnderProjectPreserveQP( final LogicalOpProject projectOp,
```
hartig, line 773: Thinking about this yet again, I notice that there is another condition to consider: Given that this function also applies the rewriting recursively to `newSubPlan1`, and then further into `subPlanUnderOp`, the projection push down somewhere further down in that sub-plan might lead to a cardinality change there. If so, then that change may trickle up all the way to the root of `newSubPlan2`. In this case, the input to `op` would have a different cardinality and, therefore, also the output of `op`. In other words, qpInfo should not be copied over, at least not if `newSubPlan2` is different from `newSubPlan1`.

Having said that, I am really not sure anymore whether it was a good idea at all to try to keep the qpInfo objects if possible. Or, no. The idea was good but it turns out to be very complex to get it right, because there are so many dependencies and special cases to consider. Even if we get it right, this makes the code here much more complex than it already was, adding a whole bunch of things that do not actually have anything to do with the project push down itself. So, even if we have already spent quite a bit of time on this PR, and on the overall issue #594, I think it is better to simply drop this PR, and the issue as well. What do you think?

KentThang: Yes, I think it is fair to drop this PR and close the issue, but I assume the existing cases we have added won't be reverted?

hartig: Yes, the cases that you added in the already merged PR related to the issue do not need to be reverted.

I will close the PR and the issue now.

PR #616, opened 2026-05-12
Add CLI tool to execute queries via HTTP

### LINKED ISSUE #606: Add CLI to execute queries via HTTP
Currently, queries can be executed either:
- through the HTTP API against a running HeFQUIN server, or
- through the existing query CLI, which creates a local HeFQUINEngine instance for each invocation.

We want to add a new CLI command that executes queries against an already running HeFQUIN server via HTTP.

KentThang:
Closes #606.

I copied a segment of `integrateEngineIntoJena()` from `HeFQUINEngineBuilder` so the custom syntax for HeFQUIN can be used in the query factory.


KentThang:
Maybe mentionable that I initially tried to use `QueryExecutionHTTPBuilder` but I opted to just use `HttpRequest` since it felt more lightweight.

bin/hefquin-client
hartig, line 1: Please rename this to `hefquin-client` (to be consistent with the `hefquin-server` program that we have).

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	/**
+	 * Returns the command name used to invoke the tool.
+	 *
+	 * @return The name of the command.
+	 */
+	@Override
+	protected String getCommandName() {
+		return "hefquin-httpquery";
```
hartig:
```suggestion
		return "hefquin-client";
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -0,0 +1,252 @@
...
+ * A command-line tool that executes SPARQL queries against a remote SPARQL
+ * endpoint using HTTP and prints the resulting query result set.
+ *
+ * The tool sends a SPARQL SELECT query to a given HTTP endpoint using a
+ * standard HTTP POST request with SPARQL Protocol encoding. It expects the
+ * endpoint to return results in SPARQL JSON Results format, which is then
+ * parsed into a Jena {@code ResultSet}.
+ *
+ * The resulting bindings are printed to standard output or optionally written
+ * to a file in a user-specified results format.
```
hartig:
```suggestion
 * A command-line tool that interacts with a HeFQUIN service to execute a SPARQL
 * query over the federation for which the service is set up, and prints the query result
 * returned by the service.
 *
 * The tool sends a SPARQL SELECT query to a given HTTP endpoint using a
 * standard HTTP POST request with SPARQL Protocol encoding. It expects the
 * endpoint to return results in SPARQL JSON Results format, which is then
 * parsed into a Jena {@code ResultSet}.
 *
 * The resulting solution mappings are printed to standard output or written
 * to a file in a user-selected results format.
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	public RunHttpQuery( final String[] argv ) {
+		super( argv );
+
+		registerHeFQUINJenaIntegration();
+
+		addModule( modResults );
+
+		add( argServerAddress, "--server", "SPARQL endpoint URI" );
```
hartig:
```suggestion
		add( argServerAddress, "--server", "Address of HeFQUIN service" );
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -0,0 +1,252 @@
...
+	 * Returns the usage summary string of the command, showing the required arguments.
+	 *
+	 * @return A string that describes the usage of the command.
+	 */
+	@Override
+	protected String getSummary() {
+		return getCommandName() + " " +
+		"--query=<query> " +
+		"--server=<server-address> " +
+		"[--outputToFile=<file-name>]";
```
hartig:
```suggestion
		"--server=<server address> " +
		"[--outputToFile=<file name>]";
```

hartig (changes requested):
Looks good for the initial version! I have just a few minor edit suggestions.

hartig (approved):

Issue #607, opened 2026-05-07
Add INFO-level logging for executable operators


KentThang:
Add SLF4J logging support to all executable operators. All logging messages should use the `INFO` level. The goal is to improve visibility into executable operator execution and lifecycle events. 

Potential logging points include:
- operator start/end
- execution details
- input/output summaries
- important state transitions
- warnings about unexpected but non-failing conditions

Inspect `SparqlServlet.java` to see how it utilizes the logger (avoid string concatenation).

List of executable operators (list does not include superclasses but those should probably be logged too):
- [x] ExecOpBind
- [x] ExecOpBinaryUnion
- [x] ExecOpDuplicateRemoval
- [x] ExecOpFilter
- [x] ExecOpGlobalToLocal
- [x] ExecOpHashJoin1
- [x] ExecOpHashJoin2
- [x] ExecOpIndexNestedLoopsJoinSPARQL
- [x] ExecOpIndexNestedLoopsJoinTPF
- [x] ExecOpLocalToGlobal
- [x] ExecOpLookupJoinViaWrapperWithoutParamVars
- [x] ExecOpLookupJoinViaWrapperWithParamVars
- [x] ExecOpMultiwayUnion
- [x] ExecOpNaiveNestedLoopsJoin
- [x] ExecOpParallelBindJoinSPARQLwithFILTER
- [x] ExecOpParallelBindJoinSPARQLwithUNION
- [x] ExecOpParallelBindJoinSPARQLwithVALUES
- [x] ExecOpParallelMultiwayLeftJoin
- [x] ExecOpProject
- [x] ExecOpRequestBRTPF
- [x] ExecOpRequestOther
- [x] ExecOpRequestSPARQL
- [x] ExecOpRequestTPF
- [x] ExecOpSequentialBindJoinBRTPF
- [x] ExecOpSequentialBindJoinSPARQLwithFILTER
- [x] ExecOpSequentialBindJoinSPARQLwithUNION
- [x] ExecOpSequentialBindJoinSPARQLwithVALUES
- [x] ExecOpSequentialBindJoinSPARQLwithVALUESorFILTER
- [x] ExecOpSequentialBindJoinSPARQLwithVarRenaming
- [x] ExecOpSymmetricHashJoin
- [x] ExecOpUnfold
- [x] ExecOpHashBasedMinus

hartig:
> (list does not include superclasses but those should probably be logged too)

[...]

PR #617, opened 2026-05-12
Add logging for BindJoin executable operators

KentThang:
Addresses #607 by adding logging for all bind-join related executable operators and base classes.


KentThang:
Some logs here might be way too noisy and should instead be in `DEBUG` level. See for instance, `ExecOpSequentialBindJoinSPARQLwithVarRenaming` where logs in `createUnion()`, `send()` or `renameVar()` might be spammy.

Not sure if I should change it to `DEBUG` level or just remove those logs altogether.

hartig (approved):
Some of the logging does indeed seem a bit too much, but let's keep it like that for the moment. I may reconsider when actually using it.

Also, it feels like almost all of these log messages should be on the debug level (also in the other related PRs). I will think about it.

PR #624, opened 2026-05-15
Change certain logs to DEBUG level

KentThang:
Changes certain logs added in #607 to `DEBUG` level to reduce verbosity.

Changes were made according to these rules:

> 1. The only log messages at the info level should be
> i. at the beginning of the process.. methods to say something like "Beginning to process X input solution mappings"
> ii. in the conclude..methods to say that the operator has completed
> (I believe that both of these things can be done in the corresponding base classes, but the log messages should also say which actual operator it is)
> 2. Everything else should be at the debug level.
> (Debug-level) log messages about initialization should be only in the constructors of the actual operators, not in the base classes.

Beyond those rules, I kept the logs for `wrapUpForChildX` methods at `INFO` level, with the same reasoning that conclude methods should be in INFO level.

For some operators, I have some logs both at the beginning and the end of `process` methods, these are also `INFO` level.


hartig (approved):
I will merge this PR even if it is not doing exactly what I was expecting. I will explain more in #607 .

PR #625, opened 2026-05-18
Move INFO level logs to superclasses

### LINKED ISSUE #607: Add INFO-level logging for executable operators

KentThang:
Addresses #607 by moving the `INFO` level logs of each executable operator to its superclasses. 

However, I kept some `INFO` level logs in `ExecOpSymmetricHashJoin` since `wrapUp()` isn't a method that overrides a method in any superclass. Let me know if these should be deleted or if there is some other solution!

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/BaseForUnaryExecOpWithCollectedInput.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/BaseForUnaryExecOpWithCollectedInput.java
@@ -70,9 +70,9 @@ public abstract class BaseForUnaryExecOpWithCollectedInput extends UnaryExecutab
 		// If enough solution mappings have been collected, process them now
 		// and, afterwards, remove them from the collection.
 		if ( collectedInputSolMaps.size() == minimumCollectionSize ) {
-			log.info( "Processing collected batch of {} solution mappings in {}.", collectedInputSolMaps.size(), getClass().getSimpleName() );
+			log.debug( "Processing collected batch of {} solution mappings in {}.", collectedInputSolMaps.size(), getClass().getSimpleName() );
 			_processCollectedInput(collectedInputSolMaps, sink, execCxt);
-			log.info( "Finished batch processing in {}.", getClass().getSimpleName() );
+			log.debug( "Finished batch processing in {}.", getClass().getSimpleName() );
 			collectedInputSolMaps.clear();
 			numberOfCollectionsProcessed++;
 		}
@@ -91,9 +91,9 @@ public abstract class BaseForUnaryExecOpWithCollectedInput extends UnaryExecutab
 		// next collection to be processed.
 		if (    inputSolMaps.size() >= minimumCollectionSize
 		     && collectedInputSolMaps.isEmpty()  ) {
-			log.info( "Passing through batch of {} mappings directly in {}.", inputSolMaps.size(), getClass().getSimpleName() );
+			log.debug( "Passing through batch of {} mappings directly in {}.", inputSolMaps.size(), getClass().getSimpleName() );
 			_processCollectedInput(inputSolMaps, sink, execCxt);
-			log.info( "Finished batch processing in {}.", getClass().getSimpleName() );
+			log.debug( "Finished batch processing in {}.", getClass().getSimpleName() );
 			numberOfCollectionsProcessed++;
 			return;
 		}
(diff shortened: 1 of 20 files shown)


hartig (changes requested):
Please change this PR to do the info-level logging really only in the _top-level_ base classes (`BinaryExecutableOpBase`, `NaryExecutableOpBase`, `NullaryExecutableOpBase`, and `UnaryExecutableOpBase`). The other base classes, which are subclasses of these four top-level ones, should produce log messages only at the debug level.

> However, I kept some `INFO` level logs in `ExecOpSymmetricHashJoin` since `wrapUp()` isn't a method that overrides a method in any superclass. Let me know if these should be deleted or if there is some other solution!

These messages should be at the debug level. The corresponding top-level base class (`BinaryExecutableOpBase`) has the info-level logging about wrapping up for the two input plans (i.e., in `wrapUpForChild1` and in `wrapUpForChild2`).

Also, I notice that there is some inconsistency in the format of the info-level log messages of the four top-level base classes. For instance:
* `UnaryExecutableOpBase` prints `"Processing solution mapping in {}."` whereas `NaryExecutableOpBase` prints `"Processing input solution mapping from child{} in {}."` (i.e., "solution mapping" versus "input solution mapping");
* `NaryExecutableOpBase` prints only one message at the beginning of each of the relevant methods whereas `UnaryExecutableOpBase` prints two messages per each of the relevant methods (the first message at the beginning of the method and the second message at the end of the method).

Please make that behavior consistent across all four top-level base classes.

Finally, `NullaryExecutableOpBase` does not have any info-level logging in its `execute` method. It should have something (consistent with the format used by the other three top-level base classes).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/BaseForUnaryExecOpWithCollectedInput.java
```diff
@@ -67,12 +67,12 @@ protected final void _process( final SolutionMapping inputSolMap,
 		// Add the given solution mapping to the current collection.
 		collectedInputSolMaps.add(inputSolMap);
 
-		log.info( "Processing collected batch of {} solution mappings.", collectedInputSolMaps.size() );
-
 		// If enough solution mappings have been collected, process them now
 		// and, afterwards, remove them from the collection.
 		if ( collectedInputSolMaps.size() == minimumCollectionSize ) {
+			log.info( "Processing collected batch of {} solution mappings in {}.", collectedInputSolMaps.size(), getClass().getSimpleName() );
```
hartig: The log messages in this file (i.e., not only this one here) should be at the debug level.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryplan/executable/impl/ops/UnaryExecutableOpBaseWithoutBlocking.java
```diff
@@ -31,6 +34,7 @@
  */
 public abstract class UnaryExecutableOpBaseWithoutBlocking extends UnaryExecutableOpBase
 {
+	private static final Logger log = LoggerFactory.getLogger( UnaryExecutableOpBaseWithoutBlocking.class );
```
hartig: The log messages in this file should be at the debug level. In fact, they can be removed altogether because they are essentially reporting the same as the log messages in the superclass of this class.

hartig (changes requested):
Just two things remaining.

hartig (approved):

PR #629, opened 2026-05-18
Push project operator into requests

KentThang:
Closes #570.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
@@ -2,10 +2,13 @@ package se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
+import org.apache.jena.sparql.core.Var;
 import org.apache.jena.sparql.expr.ExprList;
 import org.apache.jena.sparql.syntax.Element;
 import org.apache.jena.sparql.syntax.ElementGroup;
@@ -264,8 +267,8 @@ public class MergeRequests implements HeuristicForLogicalOptimization
 					returnPlan = newSubPlans.get(0);
 				else
 					returnPlan = LogicalPlanUtils.createPlanWithSubPlans( op,
-																	null,
-																	newSubPlans );
+					                                                      null,
+					                                                      newSubPlans );
 			}
 		}
 
@@ -384,7 +387,18 @@ public class MergeRequests implements HeuristicForLogicalOptimization
 
 		@Override
 		public void visit( final LogicalOpProject op ) {
-			// nothing to do here
+			// A project can be merged into a request operator if that request
+			// is for a SPARQL endpoint.
+			final LogicalOperator childOp = rewrittenSubPlans.get(0).getRootOperator();
+			if (    childOp instanceof LogicalOpRequest reqOp
+					&& reqOp.getRequest() instanceof SPARQLRequest req )
+			{
+				final Set<Var> newProj = new HashSet<>( op.getVariables() );
+				final SPARQLRequest newReq = new SPARQLRequestImpl( req.getQueryPattern(), newProj, req.getDistinctRequired() );
+
+				final LogicalOpRequest<?,?> mergedReqOp = new LogicalOpRequest<>( reqOp.getFederationMember(), op.mayReduce(), newReq );
+				returnPlan = new LogicalPlanWithNullaryRootImpl(mergedReqOp, null);
+			}
 		}
 
 		@Override
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
@@ -394,9 +394,14 @@ public class MergeRequests implements HeuristicForLogicalOptimization
 					&& reqOp.getFederationMember() instanceof SPARQLEndpoint
 					&& reqOp.getRequest() instanceof SPARQLRequest req )
 			{
-				final Set<Var> newProj = op.getVariables();
-				if ( req.getProjectionVars() != null )
+				final Set<Var> newProj;
+				if ( req.getProjectionVars() != null ) {
+					newProj = new HashSet<>( op.getVariables() );
 					newProj.retainAll( req.getProjectionVars() );
+				}
+				else {
+					newProj = op.getVariables();
+				}
 
 				final SPARQLRequest newReq = new SPARQLRequestImpl( req.getQueryPattern(), newProj, req.getDistinctRequired() );
 
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
@@ -2,7 +2,6 @@ package se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics;
 
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -32,6 +31,7 @@ import se.liu.ida.hefquin.federation.access.SPARQLRequest;
 import se.liu.ida.hefquin.federation.access.impl.req.BGPRequestImpl;
 import se.liu.ida.hefquin.federation.access.impl.req.SPARQLRequestImpl;
 import se.liu.ida.hefquin.federation.access.impl.req.TriplePatternRequestImpl;
+import se.liu.ida.hefquin.federation.members.SPARQLEndpoint;
 
 /**
  * Merges subplans that consists of multiple requests to the same federation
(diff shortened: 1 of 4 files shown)


KentThang:
I forgot to change the other tests that are affected by this change (project merging into request). Will fix that now.

KentThang:
Ready for review!

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
```diff
@@ -384,7 +387,18 @@ public void visit( final LogicalOpDedup op ) {
 
 		@Override
 		public void visit( final LogicalOpProject op ) {
-			// nothing to do here
+			// A project can be merged into a request operator if that request
+			// is for a SPARQL endpoint.
+			final LogicalOperator childOp = rewrittenSubPlans.get(0).getRootOperator();
+			if (    childOp instanceof LogicalOpRequest reqOp
+					&& reqOp.getRequest() instanceof SPARQLRequest req )
+			{
```
hartig, line 397: The comment is correct, but the if condition does not actually test that the request is for a SPARQL endpoint.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
```diff
@@ -384,7 +387,18 @@ public void visit( final LogicalOpDedup op ) {
...
+				final Set<Var> newProj = new HashSet<>( op.getVariables() );
+				final SPARQLRequest newReq = new SPARQLRequestImpl( req.getQueryPattern(), newProj, req.getDistinctRequired() );
```
hartig: There is no need to create an extra `HashSet` object here. You can pass `op.getVariables()` directly into the constructor.

Another issue here: `req` may already have a set of projection variables. In this case, the new `SPARQLRequestImpl` should be created with a set of projection variables that is the intersection of the previous set of `req` and `op.getVariables()`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -265,11 +267,27 @@ public void visit( final LogicalOpMinus op ) {
 	} // end of Worker
 
 	/**
-	 * Pushing a project operator into request operator is not possible yet TODO #570
-	*/
+	 * Pushes a project operator into a request operator by creating a new
+	 * request with the projection variables from the project operator.
```
hartig: Extend this comment to indicate that this is done only if the request is for a SPARQL endpoint (and then adapt the code of this function accordingly).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -265,11 +267,27 @@ public void visit( final LogicalOpMinus op ) {
...
+	 * of the original request while replacing any existing request-level
+	 * projection with the variables required by the project operator.
+	 *
+	 * A new request operator containing the rewritten request is then returned
+	 * as a plan with a nullary root.
+	 */
 	protected LogicalPlan createPlanForRequestUnderProject( final LogicalOpProject projectOp,
 	                                                        final LogicalOpRequest<?,?> reqOp,
 	                                                        final LogicalPlan inputPlan ) {
+		if ( reqOp.getRequest() instanceof SPARQLRequest oldReq ) {
```
hartig: You also need to check that the federation member of `reqOp` is a SPARQL endpoint.

```suggestion
		if (    reqOp.getFederationMember instanceof SPARQLEndpoint
		     && reqOp.getRequest() instanceof SPARQLRequest oldReq ) {
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -265,11 +267,27 @@ public void visit( final LogicalOpMinus op ) {
...
 
 	/**
-	 * Pushing a project operator into request operator is not possible yet TODO #570
-	*/
+	 * Pushes a project operator into a request operator by creating a new
+	 * request with the projection variables from the project operator.
+	 *
+	 * The resulting request preserves the duplicate-elimination requirement
+	 * of the original request while replacing any existing request-level
+	 * projection with the variables required by the project operator.
```
hartig: Why replacing? If another projection has already been pushed into the request before, then that projection still remains relevant. In other words, the new request operator should have the intersection of the two sets of projection variables.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -265,11 +267,27 @@ public void visit( final LogicalOpMinus op ) {
...
+	 *
+	 * A new request operator containing the rewritten request is then returned
+	 * as a plan with a nullary root.
+	 */
 	protected LogicalPlan createPlanForRequestUnderProject( final LogicalOpProject projectOp,
 	                                                        final LogicalOpRequest<?,?> reqOp,
 	                                                        final LogicalPlan inputPlan ) {
+		if ( reqOp.getRequest() instanceof SPARQLRequest oldReq ) {
+			final Set<Var> newProj = new HashSet<>( projectOp.getVariables() );
```
hartig: No need to create another copy of this set.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -265,11 +267,27 @@ public void visit( final LogicalOpMinus op ) {
...
+
+			final SPARQLRequest newReq = new SPARQLRequestImpl( oldReq.getQueryPattern(), newProj, oldReq.getDistinctRequired() );
```
hartig, line 300: As mentioned in my comment above, you need to cover the case in which `oldReq` already has a set of projection variables.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDownTest.java
```diff
@@ -44,13 +44,47 @@
...
+
+		// Left request produces x
+		final TriplePattern tp1 = new TriplePatternImpl(v1, v1, v1);
+		final LogicalOpRequest<?,?> reqOp1 = new LogicalOpRequest<>(
+			new SPARQLEndpointForTest("http://exA.org"),
+			false,
+			new SPARQLRequestImpl(tp1) );
+		final LogicalPlan reqPlan = new LogicalPlanWithNullaryRootImpl(reqOp1, null);
+
+		// Project keeps z
```
hartig: z?

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDownTest.java
```diff
@@ -566,10 +600,10 @@ public void pushProjectUnderBindWhole() {
 		assertTrue( result.getRootOperator() instanceof LogicalOpBind );
 
 		final LogicalPlan subResult = result.getSubPlan(0);
-		assertTrue( subResult.getRootOperator() instanceof LogicalOpProject );
+		assertTrue( subResult.getRootOperator() instanceof LogicalOpRequest );
```
hartig: That's wrong. Given that the federation member in this test is a TPF server, not a SPARQL endpoint, the projection should not be pushed into the request.

This issue is related to my other comments about actually checking that the federation member is indeed a SPARQL endpoint.

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
```diff
@@ -384,7 +387,22 @@ public void visit( final LogicalOpDedup op ) {
...
+			// A project can be merged into a request operator if that request
+			// is for a SPARQL endpoint.
+			final LogicalOperator childOp = rewrittenSubPlans.get(0).getRootOperator();
+			if (    childOp instanceof LogicalOpRequest reqOp
+					&& reqOp.getFederationMember() instanceof SPARQLEndpoint
+					&& reqOp.getRequest() instanceof SPARQLRequest req )
+			{
+				final Set<Var> newProj = op.getVariables();
+				if ( req.getProjectionVars() != null )
+					newProj.retainAll( req.getProjectionVars() );
```
hartig: My earlier comment about not having to create an extra `HashSet` object was under the assumption that you wouldn't change the set. If you change the set, then a new `HashSet` object should be created. Do it this way:

```suggestion
				final Set<Var> newProj;
				if ( req.getProjectionVars() != null ) {
					newProj = new HashSet<>( op.getVariables() );
					newProj.retainAll( req.getProjectionVars() );
				}
				else {
					newProj = op.getVariables();
				}
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectPushDown.java
```diff
@@ -265,11 +268,35 @@ public void visit( final LogicalOpMinus op ) {
...
+	 * as a plan with a nullary root.
+	 */
 	protected LogicalPlan createPlanForRequestUnderProject( final LogicalOpProject projectOp,
 	                                                        final LogicalOpRequest<?,?> reqOp,
 	                                                        final LogicalPlan inputPlan ) {
+		if (    reqOp.getFederationMember() instanceof SPARQLEndpoint
+			 && reqOp.getRequest() instanceof SPARQLRequest oldReq ) {
+			final Set<Var> newProj = projectOp.getVariables();
+			if ( oldReq.getProjectionVars() != null )
+				newProj.retainAll( oldReq.getProjectionVars() );
```
hartig: Same issue here. In the case in which the set is changes, a new `HashSet` object needs to be created. Do it in the same way as proposed in my other comment.

hartig (changes requested):
Two minor things remaining.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/MergeRequests.java
KentThang: Ah, I see! I will try to remember this for next time!

hartig (approved):
Looks good now. Thanks!

PR #642, opened 2026-05-27
Push DISTINCT into requests in more cases

KentThang:
Closes #570.


hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerBase1.java
```diff
@@ -78,8 +78,13 @@ public CompletableFuture<CardinalityResponse> issueCardinalityRequest(
 
 		// initialize the SELECT clause of the query
 		// (it needs to be a COUNT(*) without DISTINCT,
```
hartig: Can you please double check. It seems you have this comment twice now.

hartig (changes requested):
Excellent job! Just minor thing that I would like you to check.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerBase1.java
KentThang: You are right. It has been removed.

hartig (approved):

Issue #641, opened 2026-05-27
Extend 'hefquin-client' with features of 'hefquin'


hartig:
As a follow up to #606, the `bin/hefquin-client` program should be extended to have the same kinds of features as the main CLI tool (`bin/hefquin`).

---
* [x] --base
* [x] --time
* [x] --suppressResultPrintout
---
* [x] --skipExecution
---
* [x] --printSourceAssignment
* [x] --printSourceAssignmentToFile
* [x] --printLogicalPlan
* [x] --printLogicalPlanToFile
* [x] --printPhysicalPlan
* [x] --printPhysicalPlanToFile
---
* [x] --printQueryProcMeasurements
* [x] --printQueryProcMeasurementsToFile
---
* [x] --printQueryProcStats
* [x] --printQueryProcStatsToFile
---
* [x] --printFedAccessStats
* [x] --printFedAccessStatsToFile
---
For the first group above, this can be added by simply extending the program of the CLI.

The second group (with only --skipExecution) requires an extension of the request protocol in order to communicate the flag from the client program to the HeFQUIN service, and a corresponding extension of the code for the service (main entry point: `SparqlServlet` in the `se.liu.ida.hefquin.service` package).

The other groups require further extensions to the request protocol and to the code for the service, as well as extensions to the serialization format(s) for the query results!

You can take this step by step: First PR for group 1, second PR for group 2, third PR for only one of the points of group 3, fourth PR for the rest of group 3, and then one more PR per each of the remaining groups.

KentThang:
For `--skipExecution` (and the rest of the flags really), I assume that the desired behaviour of `bin/hefquin-client` should match the user-visible behaviour of `bin/hefquin` as closely as possible. However, does the same assumption apply to the internal execution flow as well?

For instance, is it sufficient to check for a `Skip Execution` header in `SparqlServlet` and if so, return a response that matches the output of `--skipExecution` for `bin/hefquin`? Or is it preferable to somehow involve the HeFQUIN Engine similar to how it is done in `bin/hefquin`?

hartig:
Yes, for these arguments, the behavior of the two programs should match.

[...]

PR #643, opened 2026-05-28
Extend hefquin-client with flags --base, --time and --suppressResultPrintout

KentThang:
Addresses "group 1" of #641.

### DIFF
--- a/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
+++ b/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
@@ -10,6 +10,7 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.time.Duration;
+import java.util.regex.Pattern;
 
 import org.apache.commons.io.output.NullPrintStream;
 import org.apache.jena.atlas.io.IndentedWriter;
@@ -60,6 +61,8 @@ public class RunHttpQuery extends CmdARQ
 	protected final ArgDecl argSuppressResultPrintout = new ArgDecl( ArgDecl.NoValue, "suppressResultPrintout" );
 	protected final ArgDecl argServerAddress = new ArgDecl( ArgDecl.HasValue, "server" );
 	protected final ArgDecl argOutputToFile =  new ArgDecl( ArgDecl.HasValue, "outputToFile" );
+
+	protected static final Pattern BASE_PATTERN = Pattern.compile( "\\bBASE\\b\\s*<[^>]+>", Pattern.CASE_INSENSITIVE );
 
 	public static void main( final String[] argv ) {
 		new RunHttpQuery( argv ).mainRun();
@@ -119,11 +122,6 @@ public class RunHttpQuery extends CmdARQ
 
 		PrintStream out = System.out;
 		PrintStream ownedStream = null;
-		if ( contains( argSuppressResultPrintout ) ) {
-			out = NullPrintStream.INSTANCE;
-		}
-
-		// File output takes precedence over result printout suppression
 		if ( contains(argOutputToFile) ) {
 			try {
 				// Appends to file rather than overwriting
@@ -134,6 +132,11 @@ public class RunHttpQuery extends CmdARQ
 			} catch ( final FileNotFoundException e ) {
 				cmdError( "Failed to create print stream for output destination: " + getValue( argOutputToFile ), false );
 			}
+		}
+
+		// Result printout suppression takes precedence over file output
+		if ( contains( argSuppressResultPrintout ) ) {
+			out = NullPrintStream.INSTANCE;
 		}
 
 		final String serverURI = getValue( argServerAddress );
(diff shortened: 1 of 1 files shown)


hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -113,6 +119,11 @@ protected void exec() {
 
 		PrintStream out = System.out;
 		PrintStream ownedStream = null;
+		if ( contains( argSuppressResultPrintout ) ) {
+			out = NullPrintStream.INSTANCE;
```
hartig: You override `out` below (in the case that `argOutputToFile` is given). `argSuppressResultPrintout` should have priority.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -250,4 +268,18 @@ protected Query getQuery() {
...
+	/**
+	 * Prepares a SPARQL query for HTTP transmission by ensuring that any base URI
+	 * defined in the query prologue is explicitly included in the serialized query.
+	 */
+	protected static String prepareQuery( final Query query ) {
+		final String base = query.getPrologue().getBaseURI();
+		String q = query.toString();
+		if ( base != null && ! q.toLowerCase().contains("base") ) {
+			q = "BASE <" + base + ">\n" + q;
+		}
```
hartig: Doing `q.toLowerCase().contains("base")` is not reliable. The query may contain a variable named `?base` or it may use some RDF vocabulary in which one of the URIs contains the substring "base".

More generally, I wonder whether this whole method is needed at all. `ModQuery` processes the `--base` argument, and the `Query` that it returns is created by taking a given base URI into account.

Have you tried using this program with the `--base` argument. Does it not work without a method like this one?

hartig (changes requested):

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
KentThang: Okay, I assumed that the opposite behaviour was desired. Will change!

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
KentThang: Yes, I noticed that `ModQuery` processes the `--base` argument when I started this issue and hoped that it worked by itself but that wasn't the case. 

Running the same query with `--base` for both `bin/hefquin` and `bin/hefquin-client`, I got results with `bin/hefquin` but no results in `bin/hefquin-client`. Maybe I should further investigate this for a cleaner solution.

KentThang: Okay after printing out the serialized query in both CLIs, I think what's happening is that Jena doesn't serialize base URIs provided in `--base`. But `bin/hefquin` still works because the `Query` object retains the base info internally. In `bin/hefquin-client` however, remote execution reparses `query.toString()` and therefore loses the base context during the process.

Should I try to figure out an alternative solution or improve on this `prepareQuery()` idea?

hartig: Yes, please investigate this a bit further. For instance, does the query string `q = query.toString()` contain the `BASE` entry of you provide it in the query file itself versus providing it via the `--base` argument? What does the query string look like that arrives at the service?

hartig: > Okay after printing out [...]

Thanks for investigating!

> Should I try to figure out an alternative solution or improve on this `prepareQuery()` idea?

Improve on this `prepareQuery()` idea. For instance, check what happens if you provide the `BASE` entry directly in the query file (rather than via the `--base` argument)---is it still present in the string that you obtain by `query.toString()`? If not, then you don't need to do the `contains("base")` check but simply add a `BASE` entry in case a base URI is given.

KentThang: If you provide a `BASE` entry in the query file, it will be present in `query.toString()`. So I am thinking some kind of `contains("base")` check needs to exist, but maybe something more refined?

Maybe I can use regex or check if the `query.ToString()` starts with "base" or something else?

hartig: > If you provide a `BASE` entry in the query file, it will be present in `query.toString()`.

Interesting. Is that base URI (provided via the `BASE` entry) also returned by `query.getBaseURI()`?

> So I am thinking some kind of `contains("base")` check needs to exist, but maybe something more refined?
> 
> Maybe I can use regex or check if the `query.ToString()` starts with "base" or something else?

Yes, some form of regex. See the following Stack Overflow question to get an idea of how this can be done in Java.

https://stackoverflow.com/questions/7938033/string-pattern-matching-in-java

KentThang: > Interesting. Is that base URI (provided via the BASE entry) also returned by query.getBaseURI()?

Yup!

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -125,6 +134,11 @@ protected void exec() {
 			}
 		}
 
+		// Result printout suppression takes precedence over file output
+		if ( contains( argSuppressResultPrintout ) ) {
+			out = NullPrintStream.INSTANCE;
+		}
```
hartig: That's not particularly elegant either. Now we may end up creating a `PrintStream` object together with a `FileOutputStream` object even if we do not need these objects.

I suggest you use the pattern:
```java
final PrintStream out;
final PrintStream ownedStream;
if ( ...
```
This way, the order of precedence must be reflected explicitly by the actual code, and will also make it easier to read the code.

hartig (changes requested):
Just one more (smaller) issue.

hartig (approved):

PR #644, opened 2026-06-10
Resolve JSON sources from RML logical sources during mapping evaluation

### LINKED ISSUE #604: Resolve JSON sources from RML logical sources during mapping evaluation
Currently, `MaterializeRDFViewFromRML` assigns a single JSON source uniformly to all extracted `SourceReferences` during evaluation, regardless of how sources are defined in the RML mapping.

The evaluation pipeline should instead follow the logical source definitions provided in the RML mapping itself:
RML mapping -> contains logical source(s) -> converted into SourceReference(s) -> evaluator loads matching source(s)

This means that:
logical sources defined in individual `rml:TriplesMaps` should be preserved during conversion to mapping algebra,
generated `SourceReferences` should retain enough information to identify their corresponding source,
the evaluator should resolve and load the correct JSON source for each `SourceReference` instead of assigning the same `JsonObject` to all sources.

KentThang:
Closes #604. 

From reading the RML specs, I believe only `rml:RelativeSourcePath` and `rml:FilePath` are relevant to this issue. 

I opted to remove the use of the hardcoded JSON file. 

The current implementation only supports `FileSourceReferences`, if anything else is returned while looping through all the source references, the program will exit prematurely. The thinking behind that is that we want to avoid silently producing incomplete or incorrect mappings.

Is it worthwhile to include the mapping I used for testing? And should I perhaps delete the hardcoded JSON file, now that it is unused?


hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/vocabulary/RMLVocab.java
```diff
@@ -1,513 +1,519 @@
...
     public static final Resource LogicalView = M_MODEL.createResource( "http://w3id.org/rml/LogicalView" );
-    
-    /** <p>A logical view join is an operation that adapts the logical iterations of 
-     *  one logical view (the child logical view) with fields from another logical 
+
+    /** <p>Denotes a relative path source.</p> */
+    public static final Resource RelativePathSource = M_MODEL.createResource( "http://w3id.org/rml/RelativePathSource" );
+
+    /** <p>Denotes a file path.</p> */
+    public static final Resource FilePath = M_MODEL.createResource( "http://w3id.org/rml/FilePath" );
```
hartig: Did you add these manually? Or did you regenerate this file from the RML ontology?

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -181,28 +182,40 @@ protected void exec( final Model rmlDescr,
...
+					cmdError( "Failed to read " + file.getPath() + ":" + e.getMessage(), true );
+					return; // Primarily used to avoid "variable not initialized" compiler error
+				}
+
+				final JsonObject jsonObject;
+				try {
+					jsonObject = new JsonObject(jsonString);
+				}
+				catch ( final JsonPathException e ) {
+					cmdError( "Invalid JSON input: failed to parse JSON string.", true );
```
hartig: It is useful to mention the file that caused the issue.
```suggestion
					cmdError( "Parsing JSON string from " + file.getPath() + " failed:" + e.getMessage(), true );
```

hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/RML2MappingAlgebra.java
```diff
@@ -299,7 +302,15 @@ public static SourceReference createSourceReference( final Resource tm )
...
 		}
 
-		return new MySourceReference(src);
+		if ( src.hasProperty( RDF.type, FDVocab.TemplateBasedInterface ) || src.hasProperty( RDF.type, FDVocab.FixedEndpointInterface ) )
+			return new MySourceReference(src);
+		else if ( src.hasProperty( RDF.type, RMLVocab.RelativePathSource ) || src.hasProperty( RDF.type, RMLVocab.FilePath ) )
+			return new FileSourceReference(src);
+		else {
+			throw new RMLParserException(
+				"Unsupported rml:source type: " + src
```
hartig:
```suggestion
				"The following object of an rml:source triple has an unsupported or unspecified type: " + src
```

hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/RML2MappingAlgebra.java
```diff
@@ -643,4 +654,23 @@ public boolean equals( final Object other ) {
 		}
 	}
 
+	public static class FileSourceReference implements SourceReference {
+		public final File file;
+		public FileSourceReference( final File file ) { this.file = file; }
+		public FileSourceReference( final Resource r ) {
+			final String path = r.getProperty( RMLVocab.path ).getString();
```
hartig: Please make this part more robust to produce descriptive error messages in case the RML description is incorrect (e.g., `r` does not have an `rml:path` property).

Also, this doesn't cover cases in which `rml:root` is specified (see https://kg-construct.github.io/rml-io/spec/docs/#source-examples).

hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/RML2MappingAlgebra.java
```diff
@@ -643,4 +654,23 @@ public boolean equals( final Object other ) {
...
+
+		@Override
+		public boolean equals( final Object other ) {
+			if ( other == this ) return true;
+
+			return (    other instanceof FileSourceReference sr
+			         && sr.file.equals(file) );
+		}
+
+		public File getFile() { return this.file; }
```
hartig:
```suggestion
		public File getFile() { return file; }
```

hartig (commented):
Apologies for the late review.

In addition to the in-code comments of this review, here are also responses to your text in the PR description itself:

> From reading the RML specs, I believe only `rml:RelativeSourcePath` and `rml:FilePath` are relevant to this issue.

Yes, I think so too.

By the way, I find it a bit strange that the [RML-IO spec](https://kg-construct.github.io/rml-io/spec/docs/) doesn't mention `rml:RelativeSourcePath` at all.

> I opted to remove the use of the hardcoded JSON file.

Makes sense.

> The current implementation only supports `FileSourceReferences`, if anything else is returned while looping through all the source references, the program will exit prematurely. The thinking behind that is that we want to avoid silently producing incomplete or incorrect mappings.

I agree!

> Is it worthwhile to include the mapping I used for testing?

In fact, can you please add some unit tests for `MaterializeRDFViewFromRML`. You can take inspiration from `RunQueryWithoutSrcSelTest`, and the mapping file that you use for testing would then go into https://github.com/LiUSemWeb/HeFQUIN/tree/main/hefquin-cli/src/test/resources

> And should I perhaps delete the hardcoded JSON file, now that it is unused?

Yes.

hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/vocabulary/RMLVocab.java
KentThang: Oh, I added these manually! How would I regenerate the file from the RML ontology?

hartig: > Oh, I added these manually! How would I regenerate the file from the RML ontology?

Generally: Enter the directory `./hefquin-vocabs/`, edit `build.sh` and uncomment the line about RML around the very end of the file, run `build.sh` (for which you need `JENA_HOME` set with the path of a directory that contains a recent version of Apache Jena), edit `build.sh` again and comment the line about RML again.

Let me do it because there might be a few extra steps.

hartig: I regenerated the file from the RML ontology but the two entries that you added manually are not there. It turns out that the RML ontology does not actually define `rml:RelativePathSource` and `rml:FilePath`. I created two issues for it: [rml-core/271](https://github.com/kg-construct/rml-core/issues/271) and [rml-io/170](https://github.com/kg-construct/rml-io/issues/170)

hartig: As per the discussion in the two RML-related issues that I created, `rml:RelativePathSource` is obsolete now, and it may take a bit of time until `rml:FilePath` shows up in the ontology. Therefore, it is okay to proceed with your manually-added `FilePath` entry for this PR, but please remove the `RelativePathSource` entry.

hartig:
Is this one ready for review? Probably, [my comment](https://github.com/LiUSemWeb/HeFQUIN/pull/644#discussion_r3422671742) about removing `RelativePathSource` still needs to be addressed.

KentThang:
Not quite yet! I am adding some tests for it right now as well as handling if root is specified.

KentThang:
Latest commit resolves the PR comments, adds tests (should I add a test for `--baseIRI` too?) and adds support for when `rml:root` is specified. In fact, the current example mapping uses `rml:root rml:MappingDirectory`. 

Right now, the mapping directory is passed kind of clumsily from the CLI, all the way to the `FileSourceReference` constructor. Perhaps I should make a context object instead?

Furthermore, `MaterializeRDFViewFromRML`  doesn't exit anymore if a mapping/RDF file wasn't provided. This was to mimic the behaviour of `RunQueryWithoutSrcSel` and for the `runWithMissingMapping()` test to pass. Is that the reason why `RunQueryWithoutSrcSel` doesn't exit early if a query or federation description isn't provided, for the tests?

hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/RML2MappingAlgebra.java
```diff
@@ -643,4 +657,51 @@ public boolean equals( final Object other ) {
...
+	public static class FileSourceReference implements SourceReference {
+		public final File file;
+		public FileSourceReference( final File file ) { this.file = file; }
+		public FileSourceReference( final Resource r, final File mappingDir ) {
+			final Statement s = r.getProperty( RMLVocab.path );
+			if ( s == null || ! s.getObject().isLiteral() )
+				throw new IllegalArgumentException( "Missing or invalid rml:path on " + r );
+
+			final Statement rootStatement = r.getProperty( RMLVocab.root );
+			String rootPath = System.getProperty( "user.dir" );
```
hartig: Can you please make this variable `final` and do the actual assignment within the following if-block. I am always getting a bit nervous when seeing cases in which such kind of a variable is reassigned later, because I am worried that we may miss possible cases for reassignment (either now or when potentially extending the code in the future).

hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/RML2MappingAlgebra.java
```diff
@@ -643,4 +657,51 @@ public boolean equals( final Object other ) {
...
+				final RDFNode rootNode = rootStatement.getObject();
+
+				if ( rootNode.isLiteral() ) {
+					rootPath = rootNode.asLiteral().getString();
+				}
+				else if ( rootNode.isResource() ) {
+					final Resource rootRes = rootNode.asResource();
+
+					// handle special cases
+					final String uri = rootRes.getURI();
```
hartig: This would fail if `rootRes` is not a URI, right? To avoid this issue, change the else-if condition above to use `isURIResource()`.

hefquin-cli/src/test/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRMLTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.PrintStream;
+import java.nio.file.Files;
+
+import org.junit.Test;
+
+public class MaterializeRDFViewFromRMLTest
+{
+	private static final String mappingFile = "src/test/resources/ExampleMappingTest.ttl";
```
hartig, line 16: Are you sure you need all these path components for the file?

I see that `RunQueryWithoutSrcSelTest` only has the names of the files that are in the `resources` directory:
```java
	private static final String queryFile = "TestQuery.rq";
	private static final String fedCatFile = "TestFedCat.ttl";
```

hefquin-cli/src/test/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRMLTest.java
```diff
@@ -0,0 +1,78 @@
...
+		System.setOut( new PrintStream(baos) );
+
+		// Run CLI
+		final String[] args = new String[] { "--mapping=" + mappingFile };
+		final int exitCode = new MaterializeRDFViewFromRML(args).mainRun(false, false);
+		assertEquals( 0, exitCode );
+
+		// Check result
+		final String result = baos.toString().replaceAll("\r", "");
+		assertEquals( "_:BSweden <http://example.org/totalCases> \"999\"^^<http://www.w3.org/2001/XMLSchema#integer> .\n", result );
```
hartig: There is no guarantee that the (file-local) identifier for the blank node in the subject position is `BSweden`. Better do the check in the way as done in `outputToFile_createsExpectedResult`.

hartig (changes requested):

hartig:
> Latest commit resolves the PR comments, adds tests (should I add a test for `--baseIRI` too?) [...]

Not needed for now.

> Right now, the mapping directory is passed kind of clumsily from the CLI, all the way to the `FileSourceReference` constructor. Perhaps I should make a context object instead?

No need. It is okay as is (for now).

> Furthermore, `MaterializeRDFViewFromRML` doesn't exit anymore if a mapping/RDF file wasn't provided. This was to mimic the behaviour of `RunQueryWithoutSrcSel` and for the `runWithMissingMapping()` test to pass. Is that the reason why `RunQueryWithoutSrcSel` doesn't exit early if a query or federation description isn't provided, for the tests?

Honestly, I don't know. But it doesn't sound right. If the argument for the mapping file is not provided, the tool should terminate immediately. How is that a problem for the `runWithMissingMapping()` test?

KentThang:
> Honestly, I don't know. But it doesn't sound right. If the argument for the mapping file is not provided, the tool should terminate immediately. How is that a problem for the runWithMissingMapping() test?

The test just simply terminates before it ever reaches the assertions. I assume the same thing would happen to the tests in `RunQueryWithoutSrcSelTest`.

hartig:
> The test just simply terminates before it ever reaches the assertions. I assume the same thing would happen to the tests in `RunQueryWithoutSrcSelTest`.

Just to double check: Do you mean that, for the following two lines from `runWithMissingMapping()`, the call of `mainRun` in the first line terminates the test and, thus, the second of these two lines is not reached?
```java
final int exitCode = new MaterializeRDFViewFromRML(args).mainRun(true, false);
assertEquals( 2, exitCode );
```

If yes, do you see a way of changing this behavior?

KentThang:
> Just to double check: Do you mean that, for the following two lines from runWithMissingMapping(), the call of mainRun in the first line terminates the test and, thus, the second of these two lines is not reached?

Yup, more specifically what happens is that `cmdError()` throws a `TerminationException` and `mainRun` runs `System.exit(ex.getCode());` if such exceptions are caught which subsequently terminates the JVM. As such, changing the args of `mainRun` won't do anything. 

I've explored different options and the one I landed on was extracting the `if ( ! contains(argRdfFile) )` to its own method, throw the same `IllegalArgumentException` but without `exit: true` and then in the `exec` method, have a try-catch block and if the exception is thrown, exit. Doing it in this roundabout way means that in the test, I can do this assertion:
`assertThrows( IllegalArgumentException.class, cmd::validateMappingArg );` without it terminating prematurely. What do you think?

hartig:
Yes, that's a useful solution. Please implement it in this way.

hefquin-cli/src/test/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRMLTest.java
KentThang: I'm not quite sure what allows `RunQueryWithoutSrcSelTest` to work by only specifying the files in the resources directory but the way it is resolved here, I need the mapping directory to correctly resolve things like `rml:path`.

Should I try to figure out a way to solve it like `RunQueryWithoutSrcSelTest`?

KentThang:
Latest commit resolves all PR comments except the one left for discussion.

hefquin-rml/src/main/java/se/liu/ida/hefquin/mappings/rml/RML2MappingAlgebra.java
```diff
@@ -643,4 +657,53 @@ public boolean equals( final Object other ) {
...
+
+				if ( rootNode.isLiteral() ) {
+					rootPath = rootNode.asLiteral().getString();
+				}
+				else if ( rootNode.isURIResource() ) {
+					final Resource rootRes = rootNode.asResource();
+
+					// handle special cases
+					final String uri = rootRes.getURI();
+
```
hartig: This can be simplified now:

```suggestion
					final String uri = rootNode.asResource().getURI();
```

hefquin-cli/src/test/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRMLTest.java
hartig: > [...]
> Should I try to figure out a way to solve it like `RunQueryWithoutSrcSelTest`?

No need.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/MaterializeRDFViewFromRML.java
```diff
@@ -222,4 +242,15 @@ protected void exec( final Model rmlDescr,
...
 	}
+
+	/**
+	 * Validates that the required RDF mapping file argument is provided.
+	 *
+	 * @throws IllegalArgumentException if no RDF input file argument is provided
+	 */
+	protected void validateMappingArg() {
+		if ( ! contains(argRdfFile) ) {
+			throw new IllegalArgumentException( "Must give an RDF input file" );
```
hartig:
```suggestion
			throw new IllegalArgumentException( "No RML mapping file provided." );
```

hartig (approved):
Just two minor edit suggestions. I will commit them right away and merge the PR afterwards.

PR #646, opened 2026-06-15
Add skipExecution header and use QueryProcContext in server

KentThang:
Addresses #641 by extending `bin/hefquin-client` with the `skipExecution` feature.

### DIFF
--- a/hefquin-cli/pom.xml
+++ b/hefquin-cli/pom.xml
@@ -47,12 +47,6 @@
       <artifactId>hefquin-service</artifactId>
       <version>0.0.12-SNAPSHOT</version>
     </dependency>
-	  <!-- hefquin-base -->
-    <dependency>
-      <groupId>se.liu.research.hefquin</groupId>
-      <artifactId>hefquin-base</artifactId>
-      <version>0.0.12-SNAPSHOT</version>
-    </dependency>
     <dependency>
       <groupId>org.apache.jena</groupId>
       <artifactId>jena-cmds</artifactId>
--- a/hefquin-cli/pom.xml
+++ b/hefquin-cli/pom.xml
@@ -7,7 +7,7 @@
     <artifactId>hefquin</artifactId>
     <version>0.0.12-SNAPSHOT</version>
   </parent>
-  
+
   <artifactId>hefquin-cli</artifactId>
 
   <!-- Project Meta Data -->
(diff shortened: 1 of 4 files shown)


KentThang:
I would like to add a file that has all the HTTP headers as constants but I am not sure where I would put it. 

Instead of the magic strings in `SparqlServlet` and `RunHttpQuery` like `"X-HeFQUIN-Skip-Execution"` , I'd have something like `SKIP_EXECUTION` or later on for the third group of extensions `PRINT_LOGICAL_PLAN` etc. 

Would `hefquin-service/src/test/resources/` be an appropriate place to put this? Or do I just add these constants to both `RunHttpQuery` and `SparqlServlet` in a similar way to how constants are currently declared in `SparqlServlet` for the accept headers?

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -175,12 +177,18 @@ protected void exec( final String serverURI,
...
 			.header( "Accept", "application/sparql-results+json" )
 			.timeout( Duration.ofSeconds(60) )
-			.POST( HttpRequest.BodyPublishers.ofString(prepareQuery( query )) )
+			.POST( HttpRequest.BodyPublishers.ofString(prepareQuery( query )) );
+
+		if ( contains(argSkipExecution) )
+			builder.header( "X-HeFQUIN-Skip-Execution", "true" );
+
+		final HttpRequest request = builder
+			.POST( HttpRequest.BodyPublishers.ofString(query.toString()) )
```
hartig: I think that this can be removed. You have it already above.

hartig (changes requested):
Just one minor comment in the code.

Regarding the names for the extra HTTP headers (`"X-HeFQUIN-Skip-Execution"` and the like), I agree that these should be defined as constants. The best option would be to create a new class called `HttpConstants` that you put into the `se.liu.ida.hefquin.base.net.http` package of the hefquin-base module. And within this class you do:
```java
public static final String X_HEADER_SKIP_EXECUTION = "X-HeFQUIN-Skip-Execution";
```

hefquin-cli/pom.xml
```diff
@@ -46,6 +46,12 @@
       <groupId>se.liu.research.hefquin</groupId>
       <artifactId>hefquin-service</artifactId>
       <version>0.0.12-SNAPSHOT</version>
+    </dependency>
+	  <!-- hefquin-base -->
+    <dependency>
+      <groupId>se.liu.research.hefquin</groupId>
+      <artifactId>hefquin-base</artifactId>
+      <version>0.0.12-SNAPSHOT</version>
```
hartig: I don't think this dependency needs to be declared explicitly. It should come indirectly through the dependency on hefquin-service. Or did this not work?

hartig (commented):
Looks good now. Just one question about the dependency in the POM file.

KentThang:
Latest commit removes the dependency from `pom.xml`! 

I had initially put the new `HttpConstants` in the test directory ( `hefquin-base/src/test/java/se/liu/ida/hefquin/base/net/http/HttpClientProviderTest.java`) which led to the import not working. To fix it, I thought I had to add the dependency to the `pom.xml` but once I solved it, I forgot to remove it from there.

hartig (approved):

PR #647, opened 2026-06-17
Extend bin/hefquin-client to include plan printing flags

KentThang:
Addresses the third group of extension in #641 (also including `--printExecutablePlan` & `--printExecutablePlanToFile`)

### DIFF
--- a/hefquin-service/src/main/java/se/liu/ida/hefquin/service/SparqlServlet.java
+++ b/hefquin-service/src/main/java/se/liu/ida/hefquin/service/SparqlServlet.java
@@ -148,14 +148,22 @@ public class SparqlServlet extends HttpServlet {
 					.setSkipExecution( Boolean.parseBoolean( request.getHeader(HttpConstants.X_HEADER_SKIP_EXECUTION) ) );
 
 		final QueryResponseBuffers queryResBuf = new QueryResponseBuffers();
-		if ( Boolean.parseBoolean( request.getHeader(HttpConstants.X_HEADER_PRINT_SOURCE_ASSIGNMENT)) )
+		if ( Boolean.parseBoolean( request.getHeader(HttpConstants.X_HEADER_PRINT_SOURCE_ASSIGNMENT)) ) {
+			queryResBuf.sourceAssignment = new ByteArrayOutputStream();
 			ctxBuilder.setSourceAssignmentPrinter( new TextBasedLogicalPlanPrinterImpl( new PrintStream( queryResBuf.sourceAssignment, true, StandardCharsets.UTF_8 ) ) );
-		if ( Boolean.parseBoolean( request.getHeader(HttpConstants.X_HEADER_PRINT_LOGICAL_PLAN)) )
+		}
+		if ( Boolean.parseBoolean( request.getHeader(HttpConstants.X_HEADER_PRINT_LOGICAL_PLAN)) ) {
+			queryResBuf.logicalPlan = new ByteArrayOutputStream();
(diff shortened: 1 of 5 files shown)


hefquin-service/src/main/java/se/liu/ida/hefquin/service/SparqlServlet.java
```diff
@@ -143,20 +146,40 @@ private void executeRequest( final String query,
...
-			response.getWriter().write( result.getString( "result" ) );
+
+			if ( mimeType.equals( ACCEPT_CSV ) ||
+				 mimeType.equals( ACCEPT_TSV ) ||
+				 mimeType.equals( ACCEPT_SPARQL_RESULTS_XML ) ) {
+
+				response.getWriter().write( result.getString( HttpConstants.JSON_RESULT ) );
+				return;
+			}
+			else
```
hartig: Instead of just doing `else` and, thus, simply assuming it is the right time type, do an `else if` here instead and check that it is indeed the right mime type. And then add a new `else` branch that throws an exception.

hefquin-service/src/main/java/se/liu/ida/hefquin/service/SparqlServlet.java
```diff
@@ -204,26 +227,31 @@ private static String findSupportedMimeType( final Enumeration<String> acceptHea
...
 		}
 
 		final JsonObject res = new JsonObject();
-		res.put( "result", baos.toString() );
-		res.put( "exceptions", ServletUtils.getExceptions(statsAndExceptions) );
+		res.put( HttpConstants.JSON_RESULT, resultBaos.toString() );
+		res.put( HttpConstants.JSON_SOURCE_ASSIGNMENT, queryResponseBuffers.sourceAssignment.toString() );
+		res.put( HttpConstants.JSON_LOGICAL_PLAN, queryResponseBuffers.logicalPlan.toString() );
+		res.put( HttpConstants.JSON_PHYSICAL_PLAN, queryResponseBuffers.physicalPlan.toString() );
+		res.put( HttpConstants.JSON_EXECUTABLE_PLAN, queryResponseBuffers.executablePlan.toString() );
```
hartig: Shouldn't each of them be added only if requested?

hartig (changes requested):
Excellent! Just two small things.

hefquin-service/src/main/java/se/liu/ida/hefquin/service/SparqlServlet.java
KentThang: Good point! Thanks.

KentThang:
Latest commit resolves the two comments. How do you feel about the way I solved the second comment regarding only adding the plans if they were requested? I think I like the solution I pushed but an alternative solution can make it more explicit by checking if the headers were received at server-side. I think I would have to include the headers in an object though, to pass them along from `executeRequest()` to `execute()`.

hartig:
> How do you feel about the way I solved the second comment regarding only adding the plans if they were requested? I think I like the solution I pushed but an alternative solution can make it more explicit by checking if the headers were received at server-side.

Isn't that what happens implicitly now? I mean, for each of these headers that is not received at the server side, the corresponding plan printer is not created and, thus, the corresponding `ByteArrayOutputStream` is never written to. Right?

One additional optimization would be to create each of the `ByteArrayOutputStream` in `QueryResponseBuffers` only if needed.

KentThang:
> Isn't that what happens implicitly now? I mean, for each of these headers that is not received at the server side, the corresponding plan printer is not created and, thus, the corresponding `ByteArrayOutputStream` is never written to. Right?

Yup! I was just wondering if it would be desired to make it more explicit.

> One additional optimization would be to create each of the ByteArrayOutputStream in QueryResponseBuffers only if needed.

I'll get on that.

KentThang:
I have added the optimization in the latest commit!

hartig (approved):

PR #648, opened 2026-06-18
599 mapdb for persistent caching with cli

### LINKED ISSUE #599: MapDB for persistent caching with CLI
Using MapDB could significantly improve startup performance compared with Chronicle Map for the CLI. Chronicle Map has a high initialization cost, since we start a new JVM for each run.

While the maximum throughput of MapDB is reportedly not quite as high as Chronicle Map, it provides much faster startup times, typically in the range of a few milliseconds as opposed to hundreds of milliseconds, because it avoids heavy off-heap memory allocation and large pre-initialized structures. In general, it seems better suited for short-lived processes.

keski:
This PR adds a MapDB-based persistent cache implementation and corresponding federation access manager. It also refactors cache serialization, cache keys, cache entries, and cache entry factories into reusable components, simplifying maintenance and reducing duplication across persistent cache implementations.

MapDB offers significantly lower startup overhead than Chronicle Map, making it a better fit for CLI workloads where a new JVM is started for each invocation. The PR makes this the new default cache implementation used by the engine.


hefquin-service/src/main/java/se/liu/ida/hefquin/service/SharedResourceInitializer.java
```diff
@@ -67,4 +67,13 @@ private void check( final String filenameOrURI ) {
 			throw new RuntimeException( "Failed to access: " + filenameOrURI, e );
 		}
 	}
+
+	@Override
+	public void contextDestroyed( ServletContextEvent sce ) {
```
hartig:
```suggestion
	public void contextDestroyed( final ServletContextEvent sce ) {
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithMapDBCache.java
```diff
@@ -0,0 +1,405 @@
...
+import se.liu.ida.hefquin.federation.access.impl.cache.PersistentCacheEntryFactory;
+import se.liu.ida.hefquin.federation.access.impl.cache.PersistentCacheKey;
+import se.liu.ida.hefquin.federation.access.impl.cache.PersistentCacheKey.ResponseMode;
+import se.liu.ida.hefquin.federation.access.impl.cache.mapdb.MapDBCache;
+import se.liu.ida.hefquin.federation.members.BRTPFServer;
+import se.liu.ida.hefquin.federation.members.SPARQLEndpoint;
+import se.liu.ida.hefquin.federation.members.TPFServer;
+
+/**
+ * Federation access manager with a ChronicleMap-backed persistent cache.
```
hartig: MapDB?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithMapDBCache.java
hartig, line 1: Is there any difference between this class and `FederationAccessManagerWithChronicleMapCache` except for the constructors initialize the member variable for the cache (using `MapDBCache` here and `ChronicleMapCache` in the other one) ?
If not, is it possible to have just one class where the respective `Cache` object is passed as an argument to the constructor?

hartig (changes requested):
A few comments and questions. Additionally, the PR has merge conflicts now :-(

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithMapDBCache.java
keski: Indeed, this is a typo.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithMapDBCache.java
keski: Yes! This is exactly how it is handled after refactoring in https://github.com/LiUSemWeb/HeFQUIN/issues/651, preparing a separate PR for that.

hartig: Okay. Then, we can go ahead with the code duplication in this PR for now and you will deduplicate in the other PR later.

Let me know when you have removed the merge conflicts and I will take another look at this one here.

keski: Merge conflicts removed, please have another look.

hartig (approved):
Looks good now.

PR #655, opened 2026-06-24
Create ModResultsOutExt and reduce duplication

### LINKED ISSUE #641: Extend 'hefquin-client' with features of 'hefquin'

KentThang:
Closes #641.

The new `ModResultsOutExt` features some static methods that should maybe be moved to a utils file?

### DIFF
--- a/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
+++ b/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
@@ -38,7 +38,6 @@ import org.apache.jena.sparql.serializer.SerializerRegistry;
 import org.apache.jena.sparql.util.QueryExecUtils;
 
 import arq.cmdline.CmdARQ;
-import arq.cmdline.ModResultsOut;
 import arq.cmdline.ModTime;
 import se.liu.ida.hefquin.base.net.http.HttpConstants;
 import se.liu.ida.hefquin.cli.modules.ModPlanPrinting;
@@ -64,7 +63,6 @@ public class RunHttpQuery extends CmdARQ
 {
 	protected final ModTime          modTime =          new ModTime();
 	protected final ModPlanPrinting  modPlanPrinting =  new ModPlanPrinting();
-	protected final ModResultsOut    modResults =       new ModResultsOut();
 	protected final ModQuery         modQuery =         new ModQuery();
 	protected final ModResultsOutExt modResultsExt =    new ModResultsOutExt();
 
@@ -84,7 +82,6 @@ public class RunHttpQuery extends CmdARQ
 
 		addModule( modTime );
 		addModule( modPlanPrinting );
-		addModule( modResults );
 		addModule( modResultsExt );
 
 		add( argServerAddress, "--server", "Address of HeFQUIN service" );
@@ -244,7 +241,7 @@ public class RunHttpQuery extends CmdARQ
 			)
 		);
 
-		QueryExecUtils.outputResultSet( rs, query.getPrologue(), modResults.getResultsFormat(), out );
+		QueryExecUtils.outputResultSet( rs, query.getPrologue(), modResultsExt.getResultsFormat(), out );
 
 		if ( modTime.timingEnabled() ) {
 			final long time = modTime.endTimer();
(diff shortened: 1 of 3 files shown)


hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/modules/ModResultsOutExt.java
```diff
@@ -0,0 +1,181 @@
...
+	protected boolean printQueryProcStats;
+	protected boolean printFedAccessStats;
+	protected boolean printOnelineTimeStats;
+	protected String queryProcStatsFile;
+	protected String fedAccessStatsFile;
+	protected String onelineTimeStatsFile;
+
+	@Override
+	public void registerWith( final CmdGeneral cmdLine ) {
+		cmdLine.getUsage().startCategory("Execution Statistics");
```
hartig: Starting this new category is not needed. The arguments can be put under the category introduced by the superclass.

Speaking of superclass, you need to call the superclass version of this method at the beginning:
```suggestion
	public void registerWith( final CmdGeneral cmdLine ) {
		super.registerWith(cmdLine);
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/modules/ModResultsOutExt.java
```diff
@@ -0,0 +1,181 @@
...
+		cmdLine.add( argQueryProcStatsToFile, "--printQueryProcStatsToFile", "Print out statistics about the query execution process to a file" );
+		cmdLine.add( argOnelineTimeStats, "--printQueryProcMeasurements",
+			"Print out measurements about the query processing time in one line that can be used for a CSV file" );
+		cmdLine.add( argOnelineTimeStatsToFile, "--printQueryProcMeasurementsToFile", "Print out measurements about the query processing time to a file" );
+		cmdLine.add( argFedAccessStats, "--printFedAccessStats", "Print out statistics of the federation access manager" );
+		cmdLine.add( argFedAccessStatsToFile, "--printFedAccessStatsToFile", "Print out statistics of the federation access manager to a file" );
+	}
+
+	@Override
+	public void processArgs( final CmdArgModule cmdLine ) {
```
hartig, line 57: This method also needs to call the superclass version of it.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -65,17 +66,10 @@ public class RunHttpQuery extends CmdARQ
 	protected final ModPlanPrinting  modPlanPrinting =  new ModPlanPrinting();
 	protected final ModResultsOut    modResults =       new ModResultsOut();
```
hartig: This can be removed now.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -267,52 +253,45 @@ protected void exec( final String serverURI,
 
 		final JsonValue queryProcStatsValue = obj.get(HttpConstants.JSON_QUERY_PROC_STATS);
 		if ( queryProcStatsValue != null ) {
-			if ( contains( argQueryProcStats ) ) {
+			if ( modResultsExt.isPrintQueryProcStats() ) {
 				System.err.println( queryProcStatsValue.toString() );
 				System.err.println();
 			}
```
hartig: I think you can wrap this together with the file-based version of the next if-block into another method of the new module, where you then simply pass `queryProcStatsValue` and the error handler to that method.

hartig: Similarly for `oneLineTimeStats` and `fedAccessStats`.

hartig (changes requested):
Good job!!
Nonetheless, a few change requests.

KentThang:
Latest commit resolves all PR comments. 

I wasn't sure whether it would've been better to extract both of the `extractOnelineTimeStats` methods to the new module and have two `handleOnelineTimeStats` methods in there too. I ended up keeping the `extractOnelineTimeStats` methods where they were so there's only one `handleOnelineTimeStats` in `ModResultsOutExt` which takes in a `string`.

hartig (approved):
Nice!

PR #658, opened 2026-06-25
651 decouple in memory caching from persistent cache implementations

### LINKED ISSUE #651: Decouple in-memory caching from persistent cache implementations
The current persistent cache implementations combine two distinct cache layers:
- A persistent cache backed by a persistent map (e.g., `Chronicle Map` or `MapDB`)
- An in-memory cache

This issue proposes separating the in-memory cache layer so that it can be reused across persistent cache implementations. Additionally, it should be possible to define different cache capacities for the two layers. For example, the in-memory cache is limited by the available RAM, whereas the persistent store may use, e.g., memory mapped files or do disk-based lookups. 

The refactoring should preserve the current behavior while making the cache components independently reusable and composable.

keski:
This PR refactors the persistent cache implementation by introducing cache layers and a generic hierarchical cache (L1 and L2), along with a hierarchical federation access manager. The in-memory cache is now separated from the persistent cache layer, enabling composition of different cache backends (e.g., `ChronicleMapCache` and `MapDBCache`), independent configuration of cache capacities, and reduced implementation-specific code while preserving existing caching behavior.

The hierarchical cache is managed via the the engine configuration file.


hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/HierarchicalCache.java
```diff
@@ -0,0 +1,177 @@
...
+	 *
+	 * @param l1 primary cache layer consulted first during lookups
+	 * @param l2 secondary cache layer consulted when entries are not found in the
+	 *           primary cache layer
+	 */
+	public HierarchicalCache( final Cache<IdType, ObjectType> l1,
+	                          final Cache<IdType, ObjectType> l2 ) {
+		this.l1 = l1;
+		this.l2 = l2;
+		new HashMap<>();
```
hartig: ?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/HierarchicalCache.java
```diff
@@ -0,0 +1,177 @@
...
+ * @param <EntryType>  cache entry type used by underlying cache layers
+ */
+public class HierarchicalCache<IdType,
+                               ObjectType,
+                               EntryType extends CacheEntry<ObjectType>
+							   > implements Cache<IdType,ObjectType>
+{
+	protected final Cache<IdType, ObjectType> l1;
+	protected final Cache<IdType, ObjectType> l2;
+	private final Object lock = new Object();
```
hartig, line 37: I don't think this is needed.

Given that, within every method in this implementation, you have the complete code within a `synchronized` block, you can instead make the methods themselves `synchronized`. See:
https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html

But maybe this is not possible because these methods are declared in the `Cache` interface, where they are not declared as `synchronized`. In this case, instead of making the methods `synchronized`, keep the current structure (with `synchronized` blocks inside the methods), but synchronize on `this`.

hartig (changes requested):
Just two comments at this point.

Before I continue, please remove the merge conflicts. For instance, I notice that `MapDBCache.java` seems to be a new file in this PR, which shouldn't be the case right?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/HierarchicalCache.java
keski: Oops, I'll remove it.

keski:
> Just two comments at this point.
> 
> Before I continue, please remove the merge conflicts. For instance, I notice that `MapDBCache.java` seems to be a new file in this PR, which shouldn't be the case right?

Indeed, this is just a miss on my part when I created a duplicate. I'll fix it.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/HierarchicalCache.java
keski: It is possible to add `synchronized` on the method level, but I think that would remove synchronization between `evict`, `get`, and `put`, right?

With respect to synchronizing on `this`, my understanding is the it's usually preferable to use an explicit lock, since it avoids exposing the synchronization mechanism to other callers. However, I also don't see any risks with synchronizing on `this` in this case, so it's really just a matter of preference.

Let me know what you prefer.

hartig:
There are still (or again?) merge conflict here. Can you fix them?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/HierarchicalCache.java
hartig: I was thinking that `synchronized` methods are implicitly like putting `synchronized(this)` around the complete method body.

Anyways, I am fine now if you keep the extra `lock` object.

keski: > I was thinking that `synchronized` methods are implicitly like putting `synchronized(this)` around the complete method body.

You are correct, it is equivalent to applying the lock on `this`, I had completely misunderstood.

keski:
> There are still (or again?) merge conflict here. Can you fix them?

Now the PR is ready again.

config/DefaultConfDescrExtended.ttl
```diff
@@ -6,17 +6,34 @@ PREFIX ex:     <http://example.org/>
 
 [] rdf:type ec:HeFQUINEngineConfiguration ;
    ec:fedAccessMgr [
-        ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.FederationAccessManagerWithMapDBCache" ;
+    ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.FederationAccessManagerWithHierarchicalCache" ;
         ec:constructorArguments (
             [ ec:argumentTypeName "se.liu.ida.hefquin.federation.access.FederationAccessManager" ;
               ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.AsyncFederationAccessManagerImpl" ;
+              ec:constructorArguments ( ec:value:ExecServiceForFedAccess ) ]
```
hartig, line 13:
```suggestion
              ec:constructorArguments (
                ec:value:ExecServiceForFedAccess
                10  # defaultMaxParallelRequests
              )
            ]
```

config/DefaultConfDescr.ttl
```diff
@@ -18,17 +18,34 @@
    ] .
 
 _:bFederationAccessManager rdf:type ec:FederationAccessManager ;
-    ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.FederationAccessManagerWithMapDBCache" ;
+    ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.FederationAccessManagerWithHierarchicalCache" ;
     ec:constructorArguments (
         [ ec:argumentTypeName "se.liu.ida.hefquin.federation.access.FederationAccessManager" ;
           ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.AsyncFederationAccessManagerImpl" ;
+          ec:constructorArguments ( ec:value:ExecServiceForFedAccess ) ]
```
hartig, line 25:
```suggestion
          ec:constructorArguments (
            ec:value:ExecServiceForFedAccess
            10  # defaultMaxParallelRequests
          )
        ]
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheLayer.java
```diff
@@ -0,0 +1,253 @@
...
+ * </p>
+ *
+ * @param <IdType>     key type used to identify cached objects
+ * @param <ObjectType> type of objects stored in the cache
+ * @param <EntryType>  cache entry type used internally
+ */
+public class CacheLayer<IdType,
+                        ObjectType,
+                        EntryType extends CacheEntry<ObjectType>
+                        > implements Cache<IdType, ObjectType>
```
hartig, line 49:
```suggestion
public class CacheLayer< IdType,
                         ObjectType,
                         EntryType extends CacheEntry<ObjectType> >
                   implements Cache<IdType, ObjectType>
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheLayer.java
```diff
@@ -0,0 +1,253 @@
...
+{
+	protected final Object lock = new Object();
+	protected final Map<IdType, EntryType> map;
```
hartig, line 52: In this case, you can synchronize on the `map` and the `lock` object can be removed, no?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/HierarchicalCache.java
```diff
@@ -0,0 +1,174 @@
...
+ * </p>
+ *
+ * @param <IdType>     key type used to identify cached objects
+ * @param <ObjectType> type of objects stored in the cache
+ * @param <EntryType>  cache entry type used by underlying cache layers
+ */
+public class HierarchicalCache<IdType,
+                               ObjectType,
+                               EntryType extends CacheEntry<ObjectType>
+							   > implements Cache<IdType,ObjectType>
```
hartig, line 33:
```suggestion
public class HierarchicalCache< IdType,
                                ObjectType,
                                EntryType extends CacheEntry<ObjectType> >
					implements Cache<IdType,ObjectType>
```

CHANGELOG.md
hartig, line 12:
```suggestion
- Support for *two* persistent caches for federation requests (including cardinality requests): a MapDB-based one and a ChronicleMap-based one, both with support for cache replacement and invalidation policies ([#546](https://github.com/LiUSemWeb/HeFQUIN/pull/546), [#648](https://github.com/LiUSemWeb/HeFQUIN/pull/648), [#658](https://github.com/LiUSemWeb/HeFQUIN/pull/658)).
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - CLI program to use the RML component of HeFQUIN explicitly ([#601](https://github.com/LiUSemWeb/HeFQUIN/pull/601), [#644](https://github.com/LiUSemWeb/HeFQUIN/issues/644)).
 ### Changed
 - Refactoring issueCardinalityRequest to make it generic ([#630](https://github.com/LiUSemWeb/HeFQUIN/pull/630)).
+- Refactoring of the persistent cache implementations by introducing cache layer abstraction and a generic hierarchical cache [#658](https://github.com/LiUSemWeb/HeFQUIN/pull/658).
 - Refactoring maxParallelRequests handling to eliminate redundant constructors in federation access managers ([#650](https://github.com/LiUSemWeb/HeFQUIN/pull/650)).
 - Refactoring of persistent cache components to improve reuse and simplify maintenance across cache implementations ([#648](https://github.com/LiUSemWeb/HeFQUIN/pull/648)).
```
hartig, line 19:
```suggestion
- Refactoring maxParallelRequests handling to eliminate redundant constructors in federation access managers ([#650](https://github.com/LiUSemWeb/HeFQUIN/pull/650)).
```

hartig (changes requested):
Excellent work! Also with the detailed Javadoc comments.

Just a few very minor comments (mostly edit suggestions).

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheLayer.java
keski: The problem is that synchronizing on a MapDB `map` can cause blocking due to MapDB's internal locking mechanisms. Using a dedicated `lock` object keeps the application-level lock separate and avoids this. For `ChronicleMap`, this is not an issue.

CHANGELOG.md
```diff
@@ -7,16 +7,14 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - New CLI program to issue queries to the HeFQUIN service ([#616](https://github.com/LiUSemWeb/HeFQUIN/issues/616), [#643](https://github.com/LiUSemWeb/HeFQUIN/issues/643), [#646](https://github.com/LiUSemWeb/HeFQUIN/issues/646), [#647](https://github.com/LiUSemWeb/HeFQUIN/issues/647), [#653](https://github.com/LiUSemWeb/HeFQUIN/issues/653), [#655](https://github.com/LiUSemWeb/HeFQUIN/issues/655), [#671](https://github.com/LiUSemWeb/HeFQUIN/issues/671)).
 - New CLI program to use the RML component of HeFQUIN explicitly ([#601](https://github.com/LiUSemWeb/HeFQUIN/pull/601), [#644](https://github.com/LiUSemWeb/HeFQUIN/issues/644), [#670](https://github.com/LiUSemWeb/HeFQUIN/issues/670)).
+- Support for *two* persistent cache implementations for federation requests (including cardinality requests): a MapDB-based one and a ChronicleMap-based one, both with support for cache replacement and invalidation policies ([#546](https://github.com/LiUSemWeb/HeFQUIN/pull/546), [#648](https://github.com/LiUSemWeb/HeFQUIN/pull/648), [#658](https://github.com/LiUSemWeb/HeFQUIN/pull/658)).
```
hartig:
```suggestion
- Support for *two* persistent cache implementations for federation requests (including cardinality requests): a MapDB-based one and a ChronicleMap-based one, both with support for cache replacement and invalidation policies ([#546](https://github.com/LiUSemWeb/HeFQUIN/pull/546), [#648](https://github.com/LiUSemWeb/HeFQUIN/pull/648), [#658](https://github.com/LiUSemWeb/HeFQUIN/pull/658), [#666](https://github.com/LiUSemWeb/HeFQUIN/pull/666)).
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheLayer.java
```diff
@@ -0,0 +1,253 @@
...
+ * @param <IdType>     key type used to identify cached objects
+ * @param <ObjectType> type of objects stored in the cache
+ * @param <EntryType>  cache entry type used internally
+ */
+public class CacheLayer< IdType,
+                         ObjectType,
+                         EntryType extends CacheEntry<ObjectType> >
+                   implements Cache<IdType, ObjectType>
+{
+	protected final Object lock = new Object();
```
hartig, line 56:
```suggestion
	// We introduce a separate object for synchronization, instead of simply
	// synchronizing on the 'map' object below. For the MapDB implementation,
	// synchronizing on 'map' can cause blocking due to MapDB's internal locking
	// mechanisms. Using a dedicated lock object keeps the application-level
	// lock separate and avoids this. For ChronicleMap, this is not an issue.
	protected final Object lock = new Object();
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheLayer.java
hartig: > > In this case, you can synchronize on the `map` and the `lock` object can be removed, no?
>
> The problem is that synchronizing on a MapDB `map` can cause blocking due to MapDB's internal locking mechanisms. [...]

Thanks for the explanation! I added this explanation as a comment for the `lock` object (commit https://github.com/LiUSemWeb/HeFQUIN/pull/658/commits/4ff5594122fb6a192fdf3ee72d0a3c68c5210aa9)

hartig (approved):

PR #660, opened 2026-06-26
Implement translateExpressions and add tests

### LINKED ISSUE #257: Consider vocabulary mappings in filter conditions
Currently, when applying vocabulary mappings, filter expressions are ignored (see around line 86
ApplyVocabularyMappings). This is is about, first, defining and, second, implementing an approach to deal with vocabulary mappings in the context of filter conditions.

KentThang:
Addresses #257 by implementing the functionally-limited first version of translateExpressions described in the latest comment of the issue thread.

To briefly summarize:
This PR adds support for rewriting rewritable SPARQL filter expressions during vocabulary mapping. Currently, equality, inequality and recursively composed AND/OR expressions are supported. Global IRIs in these expressions are translated to their local equivalents, expanding to logical OR (for =) or AND (for !=) when multiple local mappings exist. An exception is thrown if an unsupported filter expression is encountered.


KentThang:
@keski I seem to be getting some errors in `FederationAccessManagerWithMapDBCache`. Do you happen to know what might be causing them? 

I don't think I touched any of the related files in this issue and I don't think it's due to a bad merge, so I'm not sure where they're coming from.

keski:
> @keski I seem to be getting some errors in `FederationAccessManagerWithMapDBCache`. Do you happen to know what might be causing them?
> 
> I don't think I touched any of the related files in this issue and I don't think it's due to a bad merge, so I'm not sure where they're coming from.

It seems the `main` branch is currently failing the CI. I've created a PR to fix the issue: https://github.com/LiUSemWeb/HeFQUIN/pull/662

KentThang:
Thanks for looking into it!

hartig:
I have merged @keski's fix into `main` now. Probably you have rebase the branch of this PR in order to pick up the fix.

KentThang:
Forgot to make a comment, this is ready for review now.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/VocabularyMappingUtils.java
```diff
@@ -201,8 +208,114 @@ public static SPARQLGraphPattern translateGraphPattern( final OpFilter op,
...
+
+			if ( rewritten == null )
+				throw new UnsupportedOperationException( "Filter expression " + e + " cannot be rewritten" );
+			else
+				rewrittenExpressions.add(rewritten);
+		}
+		return rewrittenExpressions;
+	}
+
+	private static Expr rewrite( final Expr expr, final VocabularyMapping vm ) {
```
hartig: No need to make this one `private`.

```suggestion
	public static Expr rewrite( final Expr expr, final VocabularyMapping vm ) {
```

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/VocabularyMappingUtils.java
```diff
@@ -201,8 +208,114 @@ public static SPARQLGraphPattern translateGraphPattern( final OpFilter op,
...
-		throw new UnsupportedOperationException("Applying vocabulary mappings into FILTER expressions is not implemented yet.");
+		final ExprList rewrittenExpressions = new ExprList();
+
+		for ( final Expr e : exprs ) {
+			final Expr rewritten = rewrite(e, vm);
+
+			if ( rewritten == null )
+				throw new UnsupportedOperationException( "Filter expression " + e + " cannot be rewritten" );
+			else
+				rewrittenExpressions.add(rewritten);
```
hartig: Instead of doing these `null` check here and also in some of the helper functions below, simply throw the `UnsupportedOperationException` directly at each point where you are currently returning `null`.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/VocabularyMappingUtils.java
```diff
@@ -201,8 +208,114 @@ public static SPARQLGraphPattern translateGraphPattern( final OpFilter op,
...
+		}
+		return rewrittenExpressions;
+	}
+
+	private static Expr rewrite( final Expr expr, final VocabularyMapping vm ) {
+		if ( expr instanceof E_LogicalAnd and ) {
+			final Expr l = rewrite( and.getArg1(), vm );
+			final Expr r = rewrite( and.getArg2(), vm );
+
+			if ( l == null || r == null ) return null;
```
hartig: This is one of the points where you can avoid the `null` check if you would throw the `UnsupportedOperationException` already at the root cause.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/VocabularyMappingUtils.java
```diff
@@ -201,8 +208,114 @@ public static SPARQLGraphPattern translateGraphPattern( final OpFilter op,
...
+
+			return new E_LogicalAnd( l, r );
```
hartig: Avoid creating new objects like that if possible. In this case, if `and.getArg1().equals(l)` and `and.getArg2().equals(r)`, you can simply return `expr`, instead of creating this new object.

Same comment for the `E_LogicalOr` case below.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/EntityMapping.java
```diff
@@ -45,4 +47,14 @@ public interface EntityMapping
...
+
+	/**
+	 * Applies this entity mapping to the given node and returns the
+	 * resulting set of mapped nodes. If this entity mapping is not
+	 * relevant for the given node (i.e., applying this entity mapping
+	 * to the node does not have any effect), then the result of this
+	 * function is simply a singleton set that contains the given node
+	 * without any changes.
+	 */
+	Set<Node> applyToNode( Node n );
```
hartig: Change this to:
```
Expr applyToExpression( Expr e );
```
... and indicate in the Javadoc that the assumptions is that `e` is assumed to use the global representation of the entities mentioned in it.

... and move it below `applyToTriplePattern` (where, logically, it fits better).

Then, move the relevant functionality that you currently have in `VocabularyMappingUtils` to the implementation of this interface.

Likewise, for `SchemaMapping`. It doesn't matter if this means there is a bit of code duplication in the two implementation classes.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/VocabularyMapping.java
```diff
@@ -42,6 +44,15 @@ public interface VocabularyMapping
...
 
+	/**
+	 * Applies this vocabulary mapping to the given node and returns the
+	 * resulting set of translated nodes.
+	 * If this mapping is not relevant for the given node (i.e., applying the
+	 * mapping to the node does not have any effect), then the result of this
+	 * function is simply a singleton set that contains the given node without
+	 * any changes.
+	 */
+	Set<Node> translateNode( Node n );
```
hartig: Similarly here, change this to
```
Expr translateExpression( Expr e );
```
... and move it below `translateTriplePattern` (where, logically, it fits better).

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/VocabularyMappingUtils.java
```diff
@@ -201,8 +208,114 @@ public static SPARQLGraphPattern translateGraphPattern( final OpFilter op,
 
 	public static ExprList translateExpressions( final ExprList exprs,
 	                                             final VocabularyMapping vm ) {
-		// TODO: translate the filter expressions
-		throw new UnsupportedOperationException("Applying vocabulary mappings into FILTER expressions is not implemented yet.");
+		final ExprList rewrittenExpressions = new ExprList();
+
+		for ( final Expr e : exprs ) {
+			final Expr rewritten = rewrite(e, vm);
```
hartig: Instead of using this helper function, use `translateExpression` of `VocabularyMapping` here.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/VocabularyMappingUtils.java
```diff
@@ -201,8 +208,114 @@ public static SPARQLGraphPattern translateGraphPattern( final OpFilter op,
...
+		else if ( expr instanceof E_NotEquals neq ) {
+			left = neq.getArg1();
+			right = neq.getArg2();
+			equals = false;
+		}
+		else {
+			return null;
+		}
+
+		if ( ! right.isConstant() )
```
hartig: You should not only focus on the second argument. The first argument may be a URI as well.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/VocabularyMappingUtils.java
```diff
@@ -201,8 +208,114 @@ public static SPARQLGraphPattern translateGraphPattern( final OpFilter op,
...
+
+		final Node node = ( (NodeValue) right ).asNode();
+		if ( ! node.isURI() )
+			return null;
+
+		final Set<Node> nodes = vm.translateNode(node);
+		if ( nodes.isEmpty() )
+			return null;
+
+		final List<Expr> rewritten = new ArrayList<>();
```
hartig: Before creating this `ArrayList`, add an if-block that handles the special case of `nodes.size() == 1` (in which case there is no need to call `buildOr`/`buildAnd`).

hartig (changes requested):

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/EntityMapping.java
```diff
@@ -22,6 +24,16 @@ public interface EntityMapping
 	 */
 	Set<TriplePattern> applyToTriplePattern( TriplePattern tp );
 
+	/**
+	 * Applies this entity mapping to the given filter expression, which is
+	 * assumed to use the global representation of the entities mentioned in it
+	 * and returns the translated expression.
```
hartig:
```suggestion
	 * assumed to use the global representation of the entities mentioned in
	 * it, and returns the translated expression.
```

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/EntityMapping.java
```diff
@@ -22,6 +24,16 @@ public interface EntityMapping
...
+	 * If this entity mapping is not relevant for the given expression (i.e.,
+	 * applying this entity mapping does not change the expression), then the
+	 * result of this function is simply the given expression itself.
+	 */
```
hartig, line 38: Add the following additional comment here, and also in `SchemaMapping` and in `VocabularyMapping`.
```
	 * result of this function is simply the given expression itself.
	 * <p>
	 * It may not be possible to translate every expression; if the
	 * given expression cannot be translated, then this method
	 * throws an {@link UnsupportedOperationException}.
	 */
```

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/SchemaMapping.java
```diff
@@ -21,6 +23,16 @@ public interface SchemaMapping
 	 */
 	SPARQLGraphPattern applyToTriplePattern( TriplePattern tp );
 
+	/**
+	 * Applies this schema mapping to the given filter expression, which is
+	 * assumed to use the global representation of the schema vocabulary terms
+	 * mentioned in it and returns the translated expression.
```
hartig:
```suggestion
	 * assumed to use the global schema for the vocabulary terms mentioned
	 * in it, and returns the translated expression.
```

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/VocabularyMapping.java
```diff
@@ -19,7 +21,17 @@ public interface VocabularyMapping
 	 * pattern itself.
 	 */
 	SPARQLGraphPattern translateTriplePattern( TriplePattern tp );
-	
+
+	/**
+	 * Applies this vocabulary mapping to the given filter expression, which is
+	 * assumed to use the global representation of the vocabulary terms mentioned
+	 * in it and returns the translated expression.
```
hartig:
```suggestion
	 * assumed to use the global vocabulary, and returns the translated expression.
```

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +201,137 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
 	}
 
+	@Override
+	public Expr applyToExpression( final Expr expr ) {
+		if ( expr instanceof E_LogicalAnd and ) {
+			final Expr l = applyToExpression( and.getArg1() );
+			final Expr r = applyToExpression( and.getArg2() );
+
+			if ( l == null || r == null )
+				throw new UnsupportedOperationException( "Filter expression " + expr + " cannot be rewritten" );
```
hartig: This check should not be needed anymore, right?

(Same in the case of `E_LogicalOr` in the next if-block)

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +201,137 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
+		final Node node;
+		final boolean rewriteLeft;
+		if ( left.isConstant() ) {
+			rewriteLeft = true;
+			node = ( (NodeValue) left ).asNode();
+		}
+		else if ( right.isConstant() ) {
+			rewriteLeft = false;
+			node = ( (NodeValue) right ).asNode();
+		}
```
hartig: What if both arguments are constants? It is probably not a particularly useful case, but should still be supported correctly. For instance, consider the filter expression
```
FILTER ( ex:everyonesDarling = ex:everyonesDarling )
```
and a mapping that says the global `ex:everyonesDarling` is the same as the local `ex:myDarling`. The current implementation would produce the following filter expression
```
FILTER ( ex:myDarling = ex:everyonesDarling )
```
which evaluates to false, whereas the original filter would evaluate to true.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +201,137 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
+		else
+			throw new UnsupportedOperationException( "Filter expression " + expr + " cannot be rewritten" );
+
+		if ( ! node.isURI() )
+			throw new UnsupportedOperationException( "Filter expression " + expr + " cannot be rewritten" );
```
hartig: Remove this. Something like
```
FILTER ( ?x = "hello" )
```
should pass the translation unchanged.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +201,137 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
+
+		final Set<Node> nodes = mapGlobalTermToLocalTerms(node);
```
hartig: Better remove that extra function and move its code directly to here.

hartig (changes requested):
Here are change requests on the latest version.

All the comments that I have for `EntityMappingImpl` apply to `SchemaMappingImpl` as well.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
KentThang: I've been thinking about a solution to this and the current one I have might revert some resolved PR comments. In particular, I am thinking about extracting back out the `mapGlobalTermToLocalTerms()` function again and then if either the left or right arguments are constants, use this function (`mapGlobalTermToLocalTerms()`) to get their local terms. Thereafter, similarly to what is done now,  I add new expressions to the `Expr<List> rewritten` but we now instead do some loop that could look like this:
```
List<Expr> rewritten = new ArrayList<>();

for (Expr l : leftExprs) {
    for (Expr r : rightExprs) {
        rewritten.add(
            equals
                ? new E_Equals(l, r)
                : new E_NotEquals(l, r)
        );
    }
}
```
What are your thoughts on this solution?

hartig: Yes, this sounds like a reasonable solution. I will wait with looking at the current updates in this PR until you have implemented this solution. Simply request my review again when you are done with it.

KentThang:
Latest commit resolves the PR comments above, except for the one regarding removing the extra function (as it is now needed after some rewrites). Also note how the `nodes.size == 1` check was removed as I believe it isn't necessary anymore since the rewrite now handles both operands uniformly.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +231,94 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
+						? new E_Equals(l, r)
+						: new E_NotEquals(l, r)
+				);
+			}
+		}
+
+		return equals ? buildOr(rewritten) : buildAnd(rewritten);
+	}
+
+	private static Expr buildOr( final List<Expr> exprList ) {
```
hartig: Since you have this helper function twice now, and also the next one, it makes sense to move them somewhere else: Within the `hefquin-base` module, create a class `ExprUtils` in the (new) package `se.liu.ida.hefquin.jenaext.sparql.expr`, move these two helper functions there, and add a Javadoc comment to each of them.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +231,94 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
+			}
+		} else
+			leftExprs.add(left);
+
+		if ( right.isConstant() ) {
+			for ( final Node n : mapGlobalTermToLocalTerms(((NodeValue) right).asNode()) ) {
+				rightExprs.add(NodeValue.makeNode(n));
+			}
+		} else
+			rightExprs.add(right);
```
hartig: Now that you keep `mapGlobalTermToLocalTerms`, you can actually move more of the functionality from here into that helper function, which should reduce the duplication of the code that you have here for left and for right. So, the signature of `mapGlobalTermToLocalTerms` should be:
```java
protected List<Expr> mapGlobalTermToLocalTerms( final Expr expr )
```
... which means that all that would remain here is to do:
```java
final List<Expr> leftExprs = mapGlobalTermToLocalTerms(left);
final List<Expr> rightExprs = mapGlobalTermToLocalTerms(right);
```
Within that helper function, instead of simply casting to `NodeValue`, better use the `getConstant()` method.

Moreover, create the new `ArrayList` only within the if-branch of `isConstant()`. For the else case, you can simply do `return List.of(expr);`.

And, perhaps a bigger issue (?), I realize only now that the else case is actually not that simple. Only because `expr` is not a constant does not mean that it may not contain constants as subexpressions. For instance, `expr` may be a function call with URIs as arguments. So, you need to create a more complicated check for `expr`: recursively extract all URIs from it (do this within a new helper function in the `ExprUtils` class mentioned in my other comment), check whether any of these URIs is a global URI in this `EntityMapping` (i.e., is a key in `g2lMap`), and only if none of them is (or there are no URIs at all within `expr`), then it is safe to do `return List.of(expr);`. Otherwise, throw another `UnsupportedOperationException`.

hartig (changes requested):
Sorry, there is a bit more work to do for this one (which became clear to me only now).

KentThang:
> Sorry, there is a bit more work to do for this one (which became clear to me only now).

Don't apologize haha, I appreciate the feedback and to have more work!

Latest commit should resolve your comments above, let me know if I managed to implement the idea you had in mind. As you will notice, I opted to send `g2lMap::get` and `g2lMap::containsKey` as args to the functions in `ExtrUtils`, as opposed to the actual `g2lMap`, because I thought it would be less clean to send the whole map.

I also added another test to make sure this new logic worked as intended.

hefquin-base/src/main/java/se/liu/ida/hefquin/jenaext/sparql/expr/ExprUtils.java
```diff
@@ -0,0 +1,150 @@
...
+	 * returned unchanged as a singleton list. Non-constant expressions are only returned
+	 * unchanged if they do not contain any mapped global URIs anywhere in their
+	 * structure.
+	 *
+	 * @param e the expression to expand
+	 * @param lookup lookup function that returns the local terms for a given
+	 *               global entity
+	 * @return the translated expressions
+	 */
+	public static List<Expr> expandExpressionUsingEntityMapping( final Expr e,
```
hartig: This function captures functionality specifically related to entity mappings, which means it doesn't really belong here. My intention was that you keep it in `EntityMappingImpl`. Please move it back there.

Likewise for the next function (related to schema mappings).

hefquin-base/src/main/java/se/liu/ida/hefquin/jenaext/sparql/expr/ExprUtils.java
```diff
@@ -0,0 +1,150 @@
...
+	                                                             final Function<Node, Set<Node>> lookup,
+	                                                             final Predicate<Node> containsGlobalTerm ) {
+		if ( e.isConstant() ) {
+			final List<Expr> exprs = new ArrayList<>();
+			final Set<Node> nodes = new HashSet<>();
```
hartig: Delete the second of these and move the first one into the else block below, where you can directly populate it without having to populate the intermediate `Node` set first.

In fact, you can then even remove the `else` and put its code directly after the if-block.

hefquin-base/src/main/java/se/liu/ida/hefquin/jenaext/sparql/expr/ExprUtils.java
```diff
@@ -0,0 +1,150 @@
...
+			final Node n = e.getConstant().asNode();
+
+			if ( ! n.isURI() )
+				nodes.add(n);
```
hartig: Here you can immediately do `return List.of(e);`

hefquin-base/src/main/java/se/liu/ida/hefquin/jenaext/sparql/expr/ExprUtils.java
```diff
@@ -0,0 +1,150 @@
...
+	public static List<Expr> expandExpressionUsingSchemaMapping( final Expr e,
+	                                                             final Function<Node, Set<TermMapping>> lookup,
+	                                                             final Predicate<Node> containsGlobalTerm ) {
+		if ( e.isConstant() ) {
+			final List<Expr> exprs = new ArrayList<>();
+			final Set<Node> nodes = new HashSet<>();
+			final Node n = e.getConstant().asNode();
+
+			if ( ! n.isURI() )
+				nodes.add(n);
```
hartig: Same comments on these parts as for the corresponding parts of the previous function.

hefquin-base/src/main/java/se/liu/ida/hefquin/jenaext/sparql/expr/ExprUtils.java
```diff
@@ -0,0 +1,150 @@
...
+	 * This includes URIs appearing anywhere in the expression tree, not only
+	 * top-level constants. Variables are ignored. Function arguments are
+	 * traversed recursively.
+	 *
+	 * @param expr the expression to inspect
+	 * @param isMapped predicate that returns true if a given URI is part of
+	 *                 the global vocabulary mapping
+	 * @return true if any mapped global URI is present in the expression
+	 */
+	private static boolean containsMappedGlobalTerm( final Expr expr, final Predicate<Node> isMapped )
```
hartig: When moving the previous two functions back into their original classes, you will also have to change this function. What you can do is keep a helper function here that simply collects all `Node` objects that represent URIs within the given `Expr`.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +231,68 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
 		return result;
 	}
 
+	private Expr rewriteComparison( final Expr expr ) {
```
hartig: Make this method `protected` so that it can be used and overridden by possible subclasses.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
hartig: Why didn't you move this into the new `ExprUtils` class?

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -236,6 +272,70 @@ protected Set<SolutionMapping> applyToSolutionMapping( final SolutionMapping sol
 		return result;
 	}
 
+	private Expr rewriteComparison( final Expr expr ) {
```
hartig: Make this method `protected`

hartig (changes requested):

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
KentThang: Ah, my bad. I completely misunderstood / saw the wrong thing when resolving the comment.

KentThang:
Alright, the latest commit should resolve your PR comments correctly this time. Sorry!

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +232,85 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
+	 * If the expression is a constant URI, it is translated using the given
+	 * lookup function. If multiple local terms exist, a corresponding expression
+	 * is created for each of them. Expressions that are not constant URIs are
+	 * returned unchanged as a singleton list. Non-constant expressions are only returned
+	 * unchanged if they do not contain any mapped global URIs anywhere in their
+	 * structure.
+	 *
+	 * @param e the expression to expand
+	 * @param lookup lookup function that returns the local terms for a given
+	 *               global entity
```
hartig: This comment needs to be adapted (now that the lookup function is not an argument anymore).

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/EntityMappingImpl.java
```diff
@@ -193,4 +232,85 @@ static public Set<SolutionMapping> applyMapToSolutionMapping( final SolutionMapp
...
+			final Node n = e.getConstant().asNode();
+
+			if ( ! n.isURI() )
+				return List.of(e);
+
+			final List<Expr> exprs = new ArrayList<>();
+			final Set<Node> mappings = g2lMap.get(n);
+
+			if ( mappings == null || mappings.isEmpty() )
+				exprs.add(NodeValue.makeNode(n));
```
hartig: No need to create a new `NodeValue` here. Instead, you can simply add `e`.

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/mappings/impl/SchemaMappingImpl.java
```diff
@@ -236,6 +273,89 @@ protected Set<SolutionMapping> applyToSolutionMapping( final SolutionMapping sol
...
+	 * returned unchanged as a singleton list. Non-constant expressions are only returned
+	 * unchanged if they do not contain any mapped global URIs anywhere in their
+ 	 * structure.
+	 *
+	 * @param e the expression to expand
+	 * @param lookup lookup function that returns the term mappings for a given
+	 *               global schema term
+	 * @return the translated expressions
+	 */
+	public List<Expr> expandExpression( final Expr e ) {
```
hartig, line 324: Same two points about this method as for the same kind of method in `EntityMappingImpl`.

hartig (changes requested):
Just 2(+2) very minor things remaining.

KentThang:
Latest commit resolves the comments!

hartig (approved):

PR #664, opened 2026-07-02
Extend feddesc with interface authentication support

KentThang:
Example usages:
HttpBasicAuthentication:
```
ex:protectedSPARQL a fd:RDFBasedFederationMember ;
    fd:serviceURI "https://secure.example.com/sparql" ;
    fd:interface [
        a fd:FixedEndpointInterface ;
        fd:supportedProtocol fd:SPARQLProtocol ;
        fd:endpointAddress "https://secure.example.com/sparql" ;
        fd:authentication _:auth3
    ] .

_:auth3 a fd:HttpBasicAuthentication ;
    fd:username "${SPARQL_USER}" ;
    fd:password "${SPARQL_PASSWORD}" .
```
or HttpHeaderAuthentication:
```
ex:graphdb a fd:RDFBasedFederationMember ;

    fd:serviceURI "https://graphdb.example.com/repositories/myrepo" ;

    fd:interface [
        a fd:FixedEndpointInterface ;
        fd:supportedProtocol fd:SPARQLProtocol ;
        fd:endpointAddress "https://graphdb.example.com/repositories/myrepo" ;
        fd:authentication _:auth1
    ] .

_:auth1 a fd:HttpHeaderAuthentication ;
    fd:headers (
        [
            a fd:HttpHeader ;
            fd:headerName "Authorization" ;
            fd:headerValue "Bearer ${GRAPHDB_TOKEN}"
        ]
        [
            a fd:HttpHeader ;
            fd:headerName "Api-Key" ;
            fd:headerValue "${GRAPHDB_API_KEY}"
        ]
    ) .
```
Note that the `fd:SecretReference` hasn't been implemented as I am not sure yet which implementation of it is preferred.


hartig:
Before going further with this, I started to wonder whether we are reinventing the wheel. There may already be an RDF vocabulary for describing the authentication configuration to be used for Web APIs. If so, we may better reuse that one instead of inventing our own (similar to how we use the Hydra vocabulary to describe the URI templates for Web APIs).

It turns out there is one: the [Web of Things (WoT) Security Ontology](https://www.w3.org/2019/wot/security)!

I checked it out a bit and had the following conversation with Gemini to understand better how to use it:

---
>How exactly can the APIKeySecurityScheme of the Web of Things (WoT) Security Ontology be used? Do you have examples of its use? (ideally presented in Turtle format)
---
In the W3C Web of Things (WoT) architecture, the `APIKeySecurityScheme` is used to describe authentication configurations where a client must provide a static, opaque token (an API key) to access a Thing's resources. It is typically applied when dealing with cloud service providers or IoT platforms that use proprietary or simple token-matching mechanisms.

---

## Key Parameters of APIKeySecurityScheme

When utilizing this security scheme within the WoT Security Ontology (`https://www.w3.org/2019/wot/security#`), you must define how and where the API key should be transmitted using specific properties:

* **`wots:in`**: Specifies the location of the API key in the request. The standard values are:
* `"header"` (passed in an HTTP header)
* `"query"` (passed as a URL query parameter)
* `"body"` (passed in the payload body)
* `"cookie"` (passed via an HTTP cookie)
* `"uri"` (passed as a URI template variable)


* **`wots:name`**: The specific identifier name of the key (e.g., `X-API-KEY` or `api_token`).

---

## Usage Examples in Turtle Format

Because WoT Thing Descriptions (TDs) are written in JSON-LD, they inherently map directly to RDF triples. Below are two examples of how this security scheme is modeled using Turtle (`.ttl`) format.

### Example 1: API Key in the HTTP Header

In this scenario, a smart weather station requires an API key named `X-Station-Token` sent within the HTTP request header.

```turtle
@prefix td: <https://www.w3.org/2019/wot/td#> .
@prefix wots: <https://www.w3.org/2019/wot/security#> .
@prefix htv: <http://www.w3.org/2011/http#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

# Definition of the Thing
<urn:example:weather-station> a td:Thing ;
    td:title "Smart Weather Station" ;
    
    # Associate the security scheme with the Thing
    td:hasSecurityConfiguration <urn:example:weather-station#api-key-header-scheme> ;
    
    # A property exposed by the Thing
    td:hasPropertyAffordance [
        a td:PropertyAffordance ;
        td:title "Temperature" ;
        td:hasForm [
            a td:Form ;
            td:href "https://api.weatherstation.local/v1/temperature" ;
            td:hasContentType "application/json" ;
            htv:methodName "GET"
        ]
    ] .

# The APIKeySecurityScheme Definition
<urn:example:weather-station#api-key-header-scheme> a wots:APIKeySecurityScheme ;
    wots:scheme "apikey" ;
    wots:in "header" ;
    wots:name "X-Station-Token" .

```

### Example 2: API Key as a URI Query Parameter

In this scenario, a legacy smart plug requires the API key to be appended directly to the URL string as a query parameter called `token`.

```turtle
@prefix td: <https://www.w3.org/2019/wot/td#> .
@prefix wots: <https://www.w3.org/2019/wot/security#> .
@prefix htv: <http://www.w3.org/2011/http#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

# Definition of the Smart Plug Thing
<urn:example:smart-plug> a td:Thing ;
    td:title "Smart Power Plug" ;
    td:hasSecurityConfiguration <urn:example:smart-plug#api-key-query-scheme> ;
    
    td:hasPropertyAffordance [
        a td:PropertyAffordance ;
        td:title "Status" ;
        td:hasForm [
            a td:Form ;
            # Consumer appends ?token=YOUR_KEY during invocation
            td:href "https://plug.local/api/status" ; 
            td:hasContentType "application/json" ;
            htv:methodName "GET"
        ]
    ] .

# The APIKeySecurityScheme Definition (Query Variant)
<urn:example:smart-plug#api-key-query-scheme> a wots:APIKeySecurityScheme ;
    wots:scheme "apikey" ;
    wots:in "query" ;
    wots:name "token" .

```

### How a Consumer Uses This

When a WoT Consumer processes this Turtle (or JSON-LD) document, it learns that it cannot anonymously pull data from the `/temperature` or `/status` endpoints. It looks up its local secure credential store for a matching key associated with `<urn:example:weather-station#api-key-header-scheme>`, retrieves the secret string, and injects it into the request header as `X-Station-Token: <secret_value>`.

---
> And how would BasicSecurityScheme be used?
---
The `BasicSecurityScheme` in the Web of Things (WoT) Security Ontology corresponds to the standard HTTP Basic Authentication mechanism (defined by RFC 7617). It indicates that the client must provide a username and a password.

In the underlying network protocol (typically HTTP), these credentials are combined, Base64-encoded, and transmitted via the `Authorization` header in the format: `Authorization: Basic <base64-encoded-credentials>`.

---

## Key Features of BasicSecurityScheme

Unlike `APIKeySecurityScheme`, which requires specifying where the key goes (`wots:in`) and what it is called (`wots:name`), `BasicSecurityScheme` relies entirely on a standardized protocol behavior.

* **`wots:scheme`**: Must always be set to `"basic"`.
* **No location properties**: You do not need to specify `wots:in` because the HTTP standard strictly dictates it goes into the `Authorization` header.

---

## Usage Example in Turtle Format

Below is a Turtle (`.ttl`) representation of a Smart Camera that secures its video stream using HTTP Basic Authentication.

```turtle
@prefix td: <https://www.w3.org/2019/wot/td#> .
@prefix wots: <https://www.w3.org/2019/wot/security#> .
@prefix htv: <http://www.w3.org/2011/http#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

# Definition of the Smart Camera Thing
<urn:example:smart-camera> a td:Thing ;
    td:title "Security Camera" ;
    
    # Enforce basic authentication for the entire Thing
    td:hasSecurityConfiguration <urn:example:smart-camera#basic-auth-scheme> ;
    
    # An action affordance to take a snapshot
    td:hasActionAffordance [
        a td:ActionAffordance ;
        td:title "Take Snapshot" ;
        td:hasForm [
            a td:Form ;
            td:href "https://camera.local/api/snapshot" ;
            td:hasContentType "image/jpeg" ;
            htv:methodName "POST"
        ]
    ] .

# The BasicSecurityScheme Definition
<urn:example:smart-camera#basic-auth-scheme> a wots:BasicSecurityScheme ;
    wots:scheme "basic" .

```

---

## How it works under the hood

When a WoT Runtime or Consumer application interacts with this Thing description, it executes the following steps:

1. **Discovery:** The Consumer reads the TD and notes that `<urn:example:smart-camera#basic-auth-scheme>` requires `basic` authentication.
2. **Credential Retrieval:** The Consumer checks its secure local credential store for a pair of username and password strings assigned to this specific Thing URI or security scheme identifier.
3. **Encoding:** The Consumer combines them as `username:password` and encodes the string into Base64 (e.g., `admin:admin123` becomes `YWRtaW46YWRtaW4xMjM=`).
4. **Execution:** The Consumer dispatches the HTTP POST request to `/api/snapshot` containing the header:
```http
Authorization: Basic YWRtaW46YWRtaW4xMjM=
```
---
> Your Turtle example uses the URI https://www.w3.org/2019/wot/security#scheme but there is no such URI in the ontology.
---
You are completely right, and that is an excellent catch. I made a mistake by treating the `scheme` property as an explicit RDF predicate.

In the official W3C Web of Things (WoT) Security Ontology, there is no `wots:scheme` data property. The term `"scheme"` exists purely in the JSON-LD serialization syntax as a structural keyword to help non-RDF processors. In pure RDF/Turtle, the authentication type is derived directly from the **`rdf:type`** of the security resource (e.g., `wots:APIKeySecurityScheme` or `wots:BasicSecurityScheme`), making an explicit `scheme` predicate redundant.

Additionally, looking closer at the ontology specification:

* The `wots:in` property for an API key specifically maps to **`wots:apikeyIn`** to distinguish it from other location properties.

Here is the corrected and accurate Turtle definition for both schemas according to the official W3C vocabulary.

---

### Corrected APIKeySecurityScheme

```turtle
@prefix td: <https://www.w3.org/2019/wot/td#> .
@prefix wots: <https://www.w3.org/2019/wot/security#> .

<urn:example:weather-station#api-key-header-scheme> a wots:APIKeySecurityScheme ;
    wots:apikeyIn "header" ;
    wots:name "X-Station-Token" .

```

### Corrected BasicSecurityScheme

For HTTP Basic Auth, the location property is simply `wots:in` (not `apikeyIn`).

```turtle
@prefix td: <https://www.w3.org/2019/wot/td#> .
@prefix wots: <https://www.w3.org/2019/wot/security#> .

<urn:example:smart-camera#basic-auth-scheme> a wots:BasicSecurityScheme ;
    wots:in "header" .
```

hartig:
So, I think we can use that. The only thing that we need to add are extra properties for specifying the use name and password (for `wotsec:BasicSecurityScheme`) and the actual API key (for `wotsec:APIKeySecurityScheme`). These extra properties should go into our FD vocabulary.

Before changing the PR, can you please reproduce the examples that you have in the PR description above, but using the "wotsec" vocabulary (including extra properties that we may still have to add). The `td:hasActionAffordance` part is not needed, but the security scheme should be connected to the interface via `td:hasSecurityConfiguration`. Also, be careful, there may still be some mistakes in the Turtle examples from Gemini. So, use them as a starting point but also look into the ontology itself. And use the prefix name `wotsec:` instead of `wots:`

KentThang:
Thanks for sharing! I should've had the same instinct and looked it up myself. I'll get on it after pushing a commit for #660.

KentThang:
Here's the rewritten examples but I am not really sure how correct it is:
HttpBasicAuthentication:
```
ex:protectedSPARQL a fd:RDFBasedFederationMember ;
    fd:serviceURI "https://secure.example.com/sparql" ;

    fd:interface [
        a fd:FixedEndpointInterface ;
        fd:supportedProtocol fd:SPARQLProtocol ;
        fd:endpointAddress "https://secure.example.com/sparql" ;

        td:hasSecurityConfiguration <urn:example:secure-sparql#basic-auth>
    ] .

<urn:example:secure-sparql#basic-auth>
    a wotsec:BasicSecurityScheme ;
    wotsec:scheme "basic" ;
    fd:usernameEnv "SPARQL_USER" ;
    fd:passwordEnv "SPARQL_PASSWORD" .
```
I had some trouble and it took some time to figure out a clean rewrite for fd:HttpHeaderAuthentication. The brunt of  it was trying to figure out a good way to be able to have multiple headers but I am not sure how many endpoints need multiple headers / how common that use case is. In that case, I think we'd have to use `ComboSecurityScheme` and then potentially  `wotsec:allOf`.

I ended up just rewriting it to use the `BearerSecurityScheme` to showcase how it would be used for authentication with a bearer token. 
```
ex:graphdb a fd:RDFBasedFederationMember ;
    fd:serviceURI "https://graphdb.example.com/repositories/myrepo" ;

    fd:interface [
        a fd:FixedEndpointInterface ;
        fd:supportedProtocol fd:SPARQLProtocol ;
        fd:endpointAddress "https://graphdb.example.com/repositories/myrepo" ;

        td:hasSecurityConfiguration a wotsec:BearerSecurityScheme ;
            wotsec:scheme "bearer" ;
            fd:tokenEnv "GRAPHDB_TOKEN" .
    ] .
```
Let me know if you know how common the use case with multiple headers is and if we need it.

hartig:
A couple of comments:

In your first example, notice that the URI `urn:example:secure-sparql#basic-auth` may also be replaced by a blank node.

In both examples, there is no `wotsec:scheme` property. Remove that triple.

In the second example, the following part is incorrect Turtle:
```
fd:interface [
  # ...
  td:hasSecurityConfiguration a wotsec:BearerSecurityScheme
```

After fixing the previous issue, the `wotsec:BearerSecurityScheme` instance needs to have a `wotsec:in` property and the value of it can be `"header"`. (For the `wotsec:BasicSecurityScheme` instance in the first example, the `wotsec:in` property is not needed because, for `wotsec:BasicSecurityScheme`,  `wotsec:in` with value `"header"` is the default, as far as I understand the documentation.)

Please write the examples again based on these comments.

No need to bother about the multi-header case for the moment.

KentThang:
Do these look better?
First example rewritten:
```
ex:protectedSPARQL a fd:RDFBasedFederationMember ;
    fd:serviceURI "https://secure.example.com/sparql" ;

    fd:interface [
        a fd:FixedEndpointInterface ;
        fd:supportedProtocol fd:SPARQLProtocol ;
        fd:endpointAddress "https://secure.example.com/sparql" ;

        td:hasSecurityConfiguration [
            a wotsec:BasicSecurityScheme ;
            fd:usernameEnv "SPARQL_USER" ;
            fd:passwordEnv "SPARQL_PASSWORD"
        ]
    ] .
```
And the second example:
```
ex:graphdb a fd:RDFBasedFederationMember ;
    fd:serviceURI "https://graphdb.example.com/repositories/myrepo" ;

    fd:interface [
        a fd:FixedEndpointInterface ;
        fd:supportedProtocol fd:SPARQLProtocol ;
        fd:endpointAddress "https://graphdb.example.com/repositories/myrepo" ;

        td:hasSecurityConfiguration [
            a wotsec:BearerSecurityScheme ;
            wotsec:in "header" ;
            fd:tokenEnv "GRAPHDB_TOKEN"
        ]
    ] .
```
As for the comment regarding the `wotsec:scheme` property. I might be mistaken but if you look at the [documentation](https://www.w3.org/TR/2020/REC-wot-thing-description-20200409/?utm_source=chatgpt.com#securityscheme) for the superclass `SecurityScheme`, it explicitly lists `scheme` as a mandatory property. Shouldn't it be needed for the subclasses too or did I misunderstand your comment?

hartig:
The examples look good now.

I pushed a commit to the PR that adds the two vocabularies (`td` and `wotsec`) as Java classes which you can use for the implementation later.

> As for the comment regarding the `wotsec:scheme` property. I might be mistaken but if you look at the [documentation](https://www.w3.org/TR/2020/REC-wot-thing-description-20200409/#securityscheme) for the superclass `SecurityScheme`, it explicitly lists `scheme` as a mandatory property. Shouldn't it be needed for the subclasses too or did I misunderstand your comment?

If you look at the actual vocabulary https://w3c.github.io/wot-resources/td/v1.1/ontology/wotsec.ttl there is no property called `scheme`. It seems to me that `scheme` is some shortcut introduced by the JSON-LD context provided for this WoT spec and, thus, can only be used in JSON-LD files, not in Turtle files.

KentThang:
I see, thanks for sharing me the actual vocabulary and thanks for the clarification. Everything should be in place for me to start working on this then?

hartig:
> Everything should be in place for me to start working on this then?

Yes :-)

First step: update the fed.desc. vocabulary. Once you have done that, I can regenerate the corresponding Java class for you.

KentThang:
Alright, latest commit simplifies the fed. desc. vocabulary. The only properties that remain are the names for the environment variables, assuming that we don't want to store the credentials directly in the vocabulary, let me know if you'd prefer that instead for now.

hefquin-vocabs/feddesc.ttl
```diff
@@ -157,3 +156,18 @@ fd:brTPFProtocol a fd:DataRetrievalProtocol ;
 fd:GenericWebAPIProtocol a fd:DataRetrievalProtocol ;
     rdfs:label "Generic Web API protocol"@en ;
     rdfs:comment "The protocol to be used when interacting with a generic Web API."@en .
+
+fd:usernameEnv a owl:DatatypeProperty ;
+    rdfs:range xsd:string ;
+    rdfs:label "username environment variable"@en ;
+    rdfs:comment "The name of the environment variable from which the username is obtained."@en .
```
hartig: Even if you don't have an rdfs:domain for this property, write in this comment how this property is meant to be used.

Same for the two other new properties.

Also, is there a reason why you didn't add rdfs:domain for them?

hartig (changes requested):

hefquin-vocabs/feddesc.ttl
KentThang: I should have left a note about this!

The idea behind not adding an `rdfs:domain` was to leave room for other schemes and not tie it to one particular security scheme. The other idea was to use the more general `rdfs:domain wotsec:SecurityScheme`, is that preferred?

hartig: > The idea behind not adding an `rdfs:domain` was to leave room for other schemes and not tie it to one particular security scheme.

Makes sense!

> The other idea was to use the more general `rdfs:domain wotsec:SecurityScheme`, is that preferred?

No, that might make people think to use `wotsec:SecurityScheme` instead of the concrete class.

So, the best thing is you just extend the `rdfs: comment` with a sentence about the intended use. (for all three properties)

hartig (approved):
Looks good now!

After I have merged the PR, you can start working on the implementation of the authentication feature.

Issue #468, opened 2025-08-18
Consolidate code for cardinality estimation


hartig:
We currently have the `CardinalityEstimator` interface for logical plans and the `CardinalityEstimation` interface for physical plans. The implementation class for the former (i.e., `RequestBasedCardinalityEstimator`) has some TODOs (most importantly, the estimation for inner joins should be addressed), but at least it properly distinguishes the various types of operators and it produces `QueryPlanningInfo` objects. In contrast, the implementation classes for the latter (i.e., `CardinalityEstimationImpl` and `MinBasedCardinalityEstimationImpl`) already have slightly more sophisticated join cardinality estimation approaches from research papers, but it doesn't properly distinguish the various types of operators and it doesn't produce `QueryPlanningInfo` objects.

The task is to merge them.

PR #665, opened 2026-07-07
Rewrite addCardinalityForInnerJoin to fix incorrect quality annotation

### LINKED ISSUE #516: Incorrect quality for cardinality estimate for ExampleQuery1
The cardinality estimate for the bind-join operator in the (physical) plan for ExampleQuery1 is 'accurate', which is not correct because this is only an estimate, calculated based on 'accurate' input argument.

KentThang:
Closes #516 by refactoring the quality propagation in `addCardinalityForInnerJoin`.

The quality of the input cardinality estimates is first aggregated across all subplans, after which it is reduced once to account for the heuristic join cardinality estimation. This removes the previous iteration-dependent logic and results in more consistent cardinality estimate qualities.

For `ExampleQuery1`, instead of 'accurate' the cardinality estimate for the bind-join operator in the (physical) plan is now 'estimate based on accurates'.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/cardinality/CardinalityEstimationWorkerImpl.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/cardinality/CardinalityEstimationWorkerImpl.java
@@ -689,6 +689,12 @@ public class CardinalityEstimationWorkerImpl implements CardinalityEstimationWor
 		// estimate of 0 that is ACCURATE, which means that the join
 		// result is guaranteed to be empty and, thus, we copy the
 		// cardinality of 0 over as the join cardinality.)
+		//
+		// The quality of the resulting cardinality estimate is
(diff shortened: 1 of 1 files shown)


KentThang:
There's probably better approaches to refactoring the method still which is why I didn't remove the TODO to figure out a better approach.

I'll think about some better ideas but I'm thinking for this PR, the current change was sufficient to fix the bug. 

After looking through the list of issues, I noticed #468, which seems to address the broader TODO around improving join cardinality estimation. So the goal of this PR is limited to fixing the incorrect quality annotation exposed by `ExampleQuery1`, rather than changing the cardinality estimation heuristic itself.

hartig (approved):
Looks good!

(The issue that this PR fixes showed up today when we looked at some HeFQUIN query plans during our discussion here in Nantes 😅)

KentThang:
> (The issue that this PR fixes showed up today when we looked at some HeFQUIN query plans during our discussion here in Nantes 😅)

What a coincidence! 😅

PR #666, opened 2026-07-08
Only add --sun-misc-unsafe-memory-access flag for supporded JDK versions

### LINKED ISSUE #663: --sun-misc-unsafe-memory-access=allow is not recognized on OpenJDK 21.0.11
It seems that with the addition of the `--sun-misc-unsafe-memory-access=allow` option from the #659 merge, running any of the programs (with `OpenJDK 21.0.11`) result in the following error:
```
Unrecognized option: --sun-misc-unsafe-memory-access=allow
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

keski:


hartig (approved):
Looks okay to me, but I don't have a way of actually checking it at the moment.

@KentThang can you please check that this fix actually works on your end.

KentThang:
> @KentThang can you please check that this fix actually works on your end.

Works! Nice!

hartig:
Thanks for checking!

I will merge the PR then.

PR #667, opened 2026-07-09
Extend so filters can be pushed under l2g and g2l operators

### LINKED ISSUE #271: Extend 'FilterPushDown' to push filters under vocabulary rewriting operators
Currently, `FilterPushDown` cannot push filter conditions under a `LogicalOpLocalToGlobal` or under a `LogicalOpGlobalToLocal`. Extend the functionality of `FilterPushDown` such that the pushing of filter conditions is considered in these cases, which entails that the pushed filter conditions may need rewritten to the respective other vocabulary. The functionality for rewriting the filter conditions will be provided by #257.

KentThang:
This PR closes #271 by:
- expanding upon the work of #660 by adding `translateExpressions` and `translateExpressionsFromGlobal` (expressions can now be rewritten from local to global too).
- implementing `createPlanForL2GOrG2LUnderFilter` in `FilterPushDown`.

`createPlanForL2GOrG2LUnderFilter` now attempts to rewrite the filter expression to the vocabulary expected below the mapping operator. If rewrite succeeds, the translated filter is pushed below the l2g / g2l operator. If rewriting throws `UnsupportedOperationException`, the filter is left in its original position while the heuristic gets applied to its subplans.
 
I made this into a draft because there are a couple points I'd like some feedback on before making it a PR:
1. `FilterPushDownTest` and `VocabularyMappingUtilsTest` has some duplicate code at the moment, helper functions `createSchemaMapping`, `createEntityMapping` and `extractEqualsPairs` appear in both. Can they be moved into something like `TestUtils`?
2. In `VocabularyMappingUtils`, I added `translateExpressionsFromGlobal` but now  `translateGraphPattern( OpFilter, ... )` always calls `translateExpressions` (l2g). I assume it is desired that `translateGraphPattern` supports both directions depending on the vocabulary of the input graph pattern. Is there a preferred way to determine which translation direction should be used?

### DIFF
--- a/hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/VocabularyMapping.java
+++ b/hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/VocabularyMapping.java
@@ -24,7 +24,7 @@ public interface VocabularyMapping
 
 	/**
 	 * Applies this vocabulary mapping to the given filter expression, which is
-	 * assumed to use the global vocabulary, and returns the translated expression.
+	 * assumed to use the local vocabulary, and returns the translated expression.
 	 * If this mapping is not relevant for the given expression (i.e., applying
 	 * the mapping does not change the expression), then the result of this
 	 * function is simply the given expression itself.
@@ -34,6 +34,19 @@ public interface VocabularyMapping
 	 * throws an {@link UnsupportedOperationException}.
 	 */
 	Expr translateExpression( Expr e );
+
+	/**
+	 * Applies this vocabulary mapping to the given filter expression, which is
+	 * assumed to use the global vocabulary, and returns the translated expression.
+	 * If this mapping is not relevant for the given expression (i.e., applying
+	 * the mapping does not change the expression), then the result of this
+	 * function is simply the given expression itself.
+	 * <p>
+	 * It may not be possible to translate every expression; if the
+	 * given expression cannot be translated, then this method
+	 * throws an {@link UnsupportedOperationException}.
+	 */
+	Expr translateExpressionFromGlobal( Expr e );
 
 	/**
 	 * Applies this vocabulary mapping to the given solution mapping expressed
--- a/hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/VocabularyMapping.java
+++ b/hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/VocabularyMapping.java
@@ -33,7 +33,7 @@ public interface VocabularyMapping
 	 * given expression cannot be translated, then this method
 	 * throws an {@link UnsupportedOperationException}.
 	 */
-	Expr translateExpression( Expr e );
+	Expr translateExpressionFromLocal( Expr e );
 
 	/**
 	 * Applies this vocabulary mapping to the given filter expression, which is
--- a/hefquin-base/pom.xml
+++ b/hefquin-base/pom.xml
@@ -60,5 +60,18 @@
         </includes>
       </resource>
     </resources>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>test-jar</goal>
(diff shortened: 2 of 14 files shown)


hartig:
Let me just respond to your two questions, which I assume will cause further changes to this PR.

> 1. `FilterPushDownTest` and `VocabularyMappingUtilsTest` has some duplicate code at the moment, helper functions `createSchemaMapping`, `createEntityMapping` and `extractEqualsPairs` appear in both. Can they be moved into something like `TestUtils`?

Yes. Good idea.

> 2. In `VocabularyMappingUtils`, I added `translateExpressionsFromGlobal` but now  `translateGraphPattern( OpFilter, ... )` always calls `translateExpressions` (l2g). I assume it is desired that `translateGraphPattern` supports both directions depending on the vocabulary of the input graph pattern. Is there a preferred way to determine which translation direction should be used?

I don't see a need to ever translate query patterns from local to global. The reason is that the queries given to HeFQUIN are always assumed to be written in terms of the global vocabulary; then, every part of the query that is assigned to a federation member that uses a different vocabulary locally, this part needs to be translated to that local vocabulary.

The reason why filter and bind expressions may need to be translated in both directions is because they may appear separately from 
graph patterns in query plans, namely in filter and bind operators of the plan, and these operators might be pushed under both g2l and l2g operators.

KentThang:
In the latest commit, I've extracted the duplicate methods into `hefquin-base/src/test/java/se/liu/ida/hefquin/testutils/TestUtils.java` (wasn't sure where to put it) but `FilterPushDownTest` can't seem to resolve it even though it mananged to import and `hefquin-base` is a dependency in the `pom.xml`. Is it because it is under the test folder?

> I don't see a need to ever translate query patterns from local to global.

I assume it is fine as is then and my concern doesn't need to be addressed?

hartig:
> In the latest commit, I've extracted the duplicate methods into `hefquin-base/src/test/java/se/liu/ida/hefquin/testutils/TestUtils.java` (wasn't sure where to put it) but `FilterPushDownTest` can't seem to resolve it even though it mananged to import and `hefquin-base` is a dependency in the `pom.xml`. Is it because it is under the test folder?

Yes pretty much. If I remember correctly, the problem is that, by default, the code of the unit tests (i.e., in the test folder) of `hefquin-base` is not exposed as part of the library that other modules (such as `hefquin-engine`) can use even if they have `hefquin-base` as a dependency. But there should be a way to expose it, and then use it as a dependency for the unit test in `hefquin-engine`. Can you please take a look and try to figure out how to adapt the two `pom.xml` files such that it works (the `pom.xml` file in `hefquin-base` and the one in `hefquin-engine`).

> > I don't see a need to ever translate query patterns from local to global.
> 
> I assume it is fine as is then and my concern doesn't need to be addressed?

I think it is fine. Or, which concern exactly do you mean?
Just to be clear again: given that ` translateGraphPattern` translates from global to local, it needs to use `translateExpressionsFromGlobal`.

KentThang:
> Or, which concern exactly do you mean?

Yup, that was it, regarding `translateGraphPattern` only translating from global to local!

Latest commit has made changes to the two `pom.xml` files so the test folder is now exposed to `hefquin-engine`. Since both of my points have been resolved, I'll mark this ready for review!

hefquin-base/src/main/java/se/liu/ida/hefquin/base/data/VocabularyMapping.java
```diff
@@ -24,7 +24,7 @@ public interface VocabularyMapping
 
 	/**
 	 * Applies this vocabulary mapping to the given filter expression, which is
-	 * assumed to use the global vocabulary, and returns the translated expression.
+	 * assumed to use the local vocabulary, and returns the translated expression.
```
hartig, line 27: It is better to rename the method name accordingly, i.e. `translateExpressionFromLocal`. Likewise, in `VocabularyMappingUtils`.

(We should do the same for the methods about solution mappings, but in a separate PR late.)

hartig (changes requested):
Looks good. Just a minor request.

hartig (approved):

PR #668, opened 2026-07-13
Change fullResult data structure to hash index instead of iterable

### LINKED ISSUE #444: Use suitable data structure for full-retrieval mode of the bind-join variations
PR #443 has extended the `BaseForExecOpBindJoinWithRequestOps` class (which is the base class for all variations of the bind join algorithm) with the capability to automatically switch into a full-retrieval mode as soon as there is an input solution mapping for which it is necessary to retrieve all solution mappings for the query pattern of the bind join operator (namely, that happens for input solution mappings that do not bind any of the join variables). However, the current implementation of this capability is not efficient. At the moment, the retrieved solution mappings are simply collected into a list (see the `switchToFullRetrievalMode` function and the `fullResult` member), which is then used in a nested loops join when doing the remaining joins locally (see the `joinInFullRetrievalMode` function). Instead of a list, it would be much more efficient to use some form of a hash index that can speed up the local joins. The data structure to be used may perhaps depend on whether the join variables are all certain variables or not.

KentThang:
Closes #444. 

As said in the issue thread, should I also handle the case in which certainJoinVars.size() == 0 / certainJoinVars.isEmpty() and in that case use SolutionMappingsIndexNoJoinVars?


hartig (approved):
This one looks good. Thanks!

PR #675, opened 2026-08-19
Fix output control flags for hefquin-rmlmat

KentThang:
Previously, only the output stream format was considered when writing the output.

With this PR, flags `--formatted` and `--compress` work properly, using modified logic from `MaterializeRDFViewOfLPG`.


hartig (approved):
Thanks for the prompt fix!

PR #677, opened 2026-08-20
Multi-Request Operator

hartig:
This PR adds a new logical operator called _mreq_ (multi-request), together with a corresponding pair of a physical and an executable operator.

The use case of this operator is to improve the processing of query patterns of the following type.
```
# ...
VALUES ?s ( <http://example.org/endpoint1>
            <http://example.org/endpoint2>
            <http://example.org/endpoint3> )
SERVICE ?s {
    # ... some pattern P
}
# ...
```
The initial logical (sub)plan for this pattern without the new operator would be something like the following.
```
mu
|
+-- bind ( <http://example.org/endpoint1> AS ?s )
|   |
|   +-- req ( P at <http://example.org/endpoint1> )
|
+-- bind ( <http://example.org/endpoint2> AS ?s )
|   |
|   +-- req ( P at <http://example.org/endpoint2> )
|
+-- bind ( <http://example.org/endpoint3> AS ?s )
    |
    +-- req ( P at <http://example.org/endpoint3> )
```
Clearly, these subplans can become pretty big if the corresponding VALUES clause has many service URIs. As a consequence, during query planning, HeFQUIN will have to repeatedly traverse many (sub)subplans (e.g., the 'filter push down' heuristic would recursively go into all these subplans).

The new operator avoid this problem. The complete subplan for the pattern above is just a single _mreq_ operator:
```
mreq ( P, ?s, L )
```
where `L` is a list of the service URIs from the VALUES clause. The semantics of the operator is such that it is equivalent to the semantics of the plan above. The corresponding executable operator is obviously doing the requests to the listed endpoints in parallel.

Since one aspect of the semantics of this operator is to bind the service URIs of the accessed federation members to a given variable (?s in the example above), the implementation of the operator requires access to these service URIs. Therefore, this PR adds a `getServiceURI()` method to the `FederationMember` interface, which means that all implementations of this interface need to be extended.

**Another thing that this PR fixes** is that the bind-join algorithm was not considered for cases in which the request operator under a join has a bind operator on top (i.e., between the request operator and the join operator), but it was considered---correctly so---if the request has a filter operator on top. Now, requests with a bind operator on top are also considered for the bind-join algorithm (the bind operator would then be pulled on top of the bind join).

### DIFF
--- a/config/DefaultConfDescr.ttl
+++ b/config/DefaultConfDescr.ttl
@@ -41,7 +41,7 @@ _:bLogicalOptimizer rdf:type ec:LogicalOptimizer ;
 _:bListOfHeuristics
     ec:elementsTypeName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicForLogicalOptimization" ;
     ec:elements (
-        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.ApplyVocabularyMappings" ]
+#        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.ApplyVocabularyMappings" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemoveSubPlansWithEmptyResults" ]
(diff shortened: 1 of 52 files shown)


hartig:
~TODO before merging: Check whether more (or other types of) unit tests are required, and whether Javadoc comments are missing for the new classes and functions.~  Done

PR #679, opened 2026-08-20
678 missing hashcode method in pair class

### LINKED ISSUE #678: Missing hashCode() method in Pair class
`FederationManagerWithCache` defines the class `Key` as an extension of `Pair` to act as the cache key when caching using the generic cache implementation (`GenericCacheImpl`).`Pair` defines equality based on its contained objects but uses the default `hashCode()` implementation. This breaks the `equals()`/`hashCode()` contract: two equal `Pair` instances can have different hash codes. Consequently, when used as keys in hash-based cache structures, this results in cache misses and unnecessary cache rewrites. A `hashCode()` method should therefore be added to `Pair`:
```java
@Override
public int hashCode() {
    return Objects.hash(object1, object2);
}
```

keski:


hartig (approved):
Thanks a lot for the fix!!

Issue #669, opened 2026-07-21
Implement authentication information from federation descriptions


KentThang:
With #664 merged, the fd vocabulary now supports describing authentication configurations using the Web of Things (WoT) Security Ontology, with HeFQUIN-specific properties for retrieving credentials from environment variables. What remains is to make these authentication configurations affect the actual HTTP requests made by HeFQUIN.

Below details ideas and an outline for a possible implementation.
### 1. Parse authentication in `FederationDescriptionReader`
`FederationDescriptionReader` could parse the optional `td:hasSecurityConfiguration` property from an interface and create a Java object of the authentication configuration. For example, the reader could handle:
- `wotsec:BasicSecurityScheme` and read `fd:usernameEnv`, `fd:passwordEnv` and obtain the actual values from the corresponding environment variables.
- `wotsec:BearerSecurityScheme` and verify that `wotsec:in` is `"header"`, read `fd:tokenEnv` and obtain the token from the corresponding environment variable.

The parsing could potentially be implemented in a separate helper method or class rather than directly inside createFederationMember.
For example:
```
AuthenticationConfiguration auth =
    parseAuthenticationConfiguration(iface, fd);

```
The result could then be passed when creating the federation member:
```
return createSPARQLEndpoint(
    addrStr,
    vocabMap,
    auth
);
```
### 2. Introduce a Java abstraction for authentication
As shown just above, a new interface / abstraction could represent the authentication information independently of the RDF vocabulary.
For example:
```
public interface AuthenticationConfiguration {
    void apply(HttpRequest.Builder builder);
}
```
Possible implementations for this class could include `BasicAuthenticationConfiguration` or `BearerAuthenticationConfiguration`.

For example, the implementations could be responsible for adding the appropriate HTTP headers:
`Authorization: Basic <credentials>`
or
`Authorization: Bearer <token>`
This would avoid making every request processor know how to interpret WoT security scheme types.

An alternative would be to have the authentication configuration expose the necessary information and have a shared utility apply it to an `HttpRequest.Builder`.
### 3. Store the authentication configuration in the federation member
The parsed authentication config needs to be retained somehow after the federation description has been read.

One possible way is to add authentication information to the relevant endpoint/member abstraction, like so:

`AuthenticationConfiguration getAuthenticationConfiguration();`

or to add it to the concrete endpoint implementations.

For example, `SPARQLEndpointImpl` could receive the configuration in its constructor:
```
new SPARQLEndpointImpl(
[...]

PR #680, opened 2026-08-20
Implement AuthenticationInformation to handle authentication with BearerSecurityScheme

KentThang:
Adds authentication with the `wotsec:BearerSecurityScheme`.

The newly added `parseAuthConfig()` in `FederationDescriptionReader` parses the interface for the authentication information and creates a `BearerAuthenticationInformation` object. 

This object is then passed on and stored in the relevant federation member object. 

Lastly, in the `RequestProcessorImpl` files, the token is applied to the request via an `applyTo()` method that is in the `BearerAuthenticationInformation` object.

Also adds tests in `FederationDescriptionReaderTest` (that tests that the token is parsed properly) and `BearerAuthenticationInformationTest` (that tests that the token is applied to the request properly).

The reason I've made it into a draft is that I am not sure how I should make the `XRequestProcessorImplTest` tests. My main idea was to verify that the authentication information is actually propagated from the federation member to the request processor and ultimately added to the HTTP request. But to do this end-to-end would likely require setting up a local HTTP server or some form of HTTP mocking, which feels somewhat heavyweight for these tests.

So I'm not sure if the existing `BearerAuthenticationInformationTest` and `FederationDescriptionReaderTest` provide enough coverage for the authentication-specific functionality or if the request processor tests should be extended with some lightweight way of intercepting/inspecting the generated HTTP requests.


hartig:
I will be looking at this PR now. One question before I start: I notice that you linked issue #669 as being closed by this PR. Was that deliberate? Isn't handling HTTP basic authentication another aspect of #669 ?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/SPARQLEndpoint.java
```diff
@@ -7,6 +7,9 @@ public interface SPARQLEndpoint extends RDFBasedFederationMember
 	/** Returns the URL at which this SPARQL endpoint can be reached. */
 	String getURL();
 
+	/** Returns the authentication information of this SPARQL endpoint. */
+	AuthenticationInformation getAuthenticationInformation();
```
hartig: Is there a specific reason why you added this only to `SPARQLEndpoint`, `TPFServer`, and `RESTEndpoint`? (rather than adding it to the parent interface `FederationMember` for all types of fed.members)

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/impl/SPARQLEndpointImpl.java
```diff
@@ -1,23 +1,30 @@
...
 import se.liu.ida.hefquin.federation.members.SPARQLEndpoint;
 
 public class SPARQLEndpointImpl extends BaseForRDFBasedFederationMember
                                 implements SPARQLEndpoint
 {
 	protected final String url;
+	protected final AuthenticationInformation authInfo;
 
-	public SPARQLEndpointImpl( final String url, final VocabularyMapping vm ) {
+	public SPARQLEndpointImpl( final String url, final AuthenticationInformation authInfo, final VocabularyMapping vm ) {
```
hartig: Unfortunately, after I merged PR #677 yesterday, there is a merge conflict now for these constructors (of each of the fed.member classes that you extend here in this way). The reason is that PR #677 also adds another argument to the constructors (namely, an argument to provide each fed.member with its service URI).

The merge conflict also affects `FederationTestBase.java` in this PR.

By the way, @keski also has a PR open at the moment (PR #682), which also changes the corresponding code in `FederationTestBase.java` and, thus, also has a merge conflict because of the new constructor argument that I added in PR #677. Your addition of another constructor argument here, will also affect his PR. I suggest one of you waits before resolving the merge conflict in your respective PR; otherwise, one of you will have another merge conflict later. Assuming that @KentThang can get to it faster, I propose Kent resolves the merge conflict in this PR here first, and @keski waits until this PR is merged before attempting to resolve the conflict in PR #682.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/impl/RESTEndpointImpl.java
```diff
@@ -3,23 +3,29 @@
...
 
 public class RESTEndpointImpl extends BaseForFederationMember
                               implements RESTEndpoint
 {
 	protected final String urlTemplate;
 	protected final List<RESTEndpoint.Parameter> params;
+	protected final AuthenticationInformation authInfo;
 
-	public RESTEndpointImpl( final String urlTemplate, final List<RESTEndpoint.Parameter> params ) {
+	public RESTEndpointImpl( final String urlTemplate, final AuthenticationInformation authInfo, final List<RESTEndpoint.Parameter> params ) {
```
hartig: Besides the issue that the merge conflict mention in my other comment will affect this constructor as well, I would prefer you change the order of the arguments. The arguments about the URL template and the corresponding parameters belong together. Therefore, it is more natural to have `authInfo` as the last argument here. (and after resolving the aforementioned merge conflict, the `serviceURI` argument that was added by PR #677 would be the first one, before `urlTemplate` and `params`.

Same comment for `WrappedRESTEndpointImpl`

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/authentication/AuthenticationInformation.java
```diff
@@ -0,0 +1,28 @@
+package se.liu.ida.hefquin.federation.members;
+
+import java.net.http.HttpRequest;
+import java.util.Map;
+
+/**
+ * Provides authentication information for requests to a federation member.
+ * <p>
+ * Implementations can apply the authentication information either directly
+ * to an HTTP request builder or to a collection of HTTP headers.
```
hartig, line 12:
```suggestion
 * Provides authentication information for requests to a federation member.
 * <p>
 * The information is provided in a form that it can be applied directly
 * to an HTTP request builder or to a collection of HTTP headers.
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/AuthenticationInformation.java
```diff
@@ -0,0 +1,28 @@
+package se.liu.ida.hefquin.federation.members;
```
hartig: Move this interface, as well as the corresponding class (`BearerAuthenticationInformation`), to a new package: `se.liu.ida.hefquin.federation.authentication`

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/authentication/BearerAuthenticationInformation.java
```diff
@@ -0,0 +1,37 @@
+package se.liu.ida.hefquin.federation.members;
+
+import java.net.http.HttpRequest.Builder;
+import java.util.Map;
+
+public class BearerAuthenticationInformation implements AuthenticationInformation
```
hartig, line 16: Please add a Javadoc comment.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java
```diff
@@ -119,21 +123,22 @@ protected FederationMember createFederationMember( final Resource fedMember,
...
 		final Resource iface = fedMember.getProperty(FDVocab.interface_).getResource();
+		final AuthenticationInformation authInfo = parseAuthConfig(iface, fd);
 		final RDFNode ifaceType = fd.getRequiredProperty(iface, RDF.type).getObject();
 		final Resource protocol = iface.getProperty(FDVocab.supportedProtocol).getResource();
 
 		// Check the type of interface
 		if ( ifaceType.equals(FDVocab.FixedEndpointInterface) )
 		{
-			return handleFixedEndpointInterface( iface, protocol, vocabMap, fedMember, fd, serviceURI );
+			return handleFixedEndpointInterface( iface, protocol, authInfo, vocabMap, fedMember, fd, serviceURI );
```
hartig, line 135: Here you have the arguments in the order `iface, protocol, authInfo, vocabMap` but in the next of these kind of method calls (for `handleFragmentInterface`) you have them in the order `iface, authInfo, protocol, vocabMap`. Please make this consistent.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java
```diff
@@ -427,13 +481,14 @@ protected FederationMember createGraphQLServer( final String uri ) {
 	}
 
 	protected FederationMember createWrappedRESTEndpoint( final String uri,
+	                                                      final AuthenticationInformation authInfo,
```
hartig: Same comment as for the constructor that this method calls: it would be more natural to have the `authInfo` argument after the `params` argument.

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReaderTest.java
```diff
@@ -246,4 +248,44 @@ public void missingVocabularyMappingFileThrowsExpectedException() {
...
+				+ "PREFIX wotsec:     <https://www.w3.org/2019/wot/security#>\n"
+				+ "\n"
+				+ "ex:dbpediaSPARQL a fd:RDFBasedFederationMember ;\n"
+				+ "      fd:serviceURI \"http://dbpedia.org/sparql\"^^xsd:anyURI ;\n"
+				+ "      fd:interface [ a                         fd:FixedEndpointInterface ;\n"
+				+ "                     fd:supportedProtocol      fd:SPARQLProtocol ;\n"
+				+ "                     fd:endpointAddress        \"http://dbpedia.org/sparql\"^^xsd:anyURI ;\n"
+				+ "\n"
+				+ "                     td:hasSecurityConfiguration [ a                wotsec:BearerSecurityScheme ;\n"
+				+ "                                                   wotsec:in        \"header\" ;\n"
```
hartig, line 267: Was this a mandatory property (according to the wotsec vocabulary)?

If yes---or, actually, in any case---it would be good if the `FederationDescriptionReader` performs a check, checking that the value of this property is indeed the literal `"header"`.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImpl.java
```diff
@@ -57,15 +59,24 @@ protected SolMapsResponse performRequestWithQueryExecutionHTTP( final SPARQLRequ
 	{
 		// see https://jena.apache.org/documentation/sparql-apis/#query-execution
 
+		final Map<String, String> headers = new HashMap<>();
```
hartig: It would be better if we do not have to create this extra `HashMap` for every request. To make it possible to avoid that, please extend the `AuthenticationInformation` interface with another `applyTo` method:
```java
void applyTo( QueryExecutionHTTPBuilder b );
```
... After that, revert the code here to roughly what it was before and include the use of that new method.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/RESTRequestProcessorImpl.java
```diff
@@ -50,10 +50,12 @@ public StringResponse performRequest( final RESTRequest req,
 			throws FederationAccessException
 	{
 		final URI uri = req.getURI();
-		final HttpRequest.Builder builder = HttpRequest.newBuilder( uri)
+		final HttpRequest.Builder builder = HttpRequest.newBuilder( uri )
 				.header("Accept", "application/json;charset=UTF-8")
 				.header("User-Agent", BuildInfo.getUserAgent());
 
+		fm.getAuthenticationInformation().applyTo( builder );
```
hartig: `getAuthenticationInformation()` may return `null`

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/TPFRequestProcessorImpl.java
```diff
@@ -42,6 +42,8 @@ protected TPFResponse _performRequest( final TPFRequest req,
 			"User-Agent", BuildInfo.getUserAgent()
 		);
 
+		fm.getAuthenticationInformation().applyTo( headers );
```
hartig: `getAuthenticationInformation()` may return `null`

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/BRTPFRequestProcessorImpl.java
```diff
@@ -31,6 +31,8 @@ public TPFResponse performRequest( final BRTPFRequest req, final BRTPFServer fm
 			"User-Agent", BuildInfo.getUserAgent()
 		);
 
+		fm.getAuthenticationInformation().applyTo( headers );
```
hartig: `getAuthenticationInformation()` may return `null`

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/EngineTestBase.java
```diff
@@ -299,6 +301,12 @@ else if (req.getQueryPattern() instanceof TriplePattern) {
 			return new SolMapsResponseImpl( result, new Date() );
 		}
 
+		@Override
+		public AuthenticationInformation getAuthenticationInformation() {
+			// TODO Auto-generated method stub
```
hartig:
```suggestion
```

hartig: Same for the other such cases that this PR adds into this file.

hartig (changes requested):

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/impl/SPARQLEndpointImpl.java
hartig: PR #682 by @keski is merged now.

KentThang:
> One question before I start: I notice that you linked issue https://github.com/LiUSemWeb/HeFQUIN/issues/669 as being closed by this PR. Was that deliberate? Isn't handling HTTP basic authentication another aspect of https://github.com/LiUSemWeb/HeFQUIN/issues/669 ?

Not deliberate, you are right that it is another aspect of #669.

hartig:
> Not deliberate, you are right that it is another aspect of #669.

Okay. I have removed the link to #669.

hefquin-access/src/test/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReaderTest.java
KentThang: According to [this](https://www.w3.org/TR/2020/REC-wot-thing-description-20200409/#sec-default-values) and [this](https://www.w3.org/TR/2020/REC-wot-thing-description-20200409/#bearersecurityscheme) ( I couldn't find anything about it in the actual vocabulary, it isn't a mandatory property and comes with a default value as header.

I'll add so that it performs the check.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/SPARQLEndpoint.java
KentThang: The thought crossed my mind but I wasn't sure if I we also wanted to add `getAuthenticationInformation()` to `GraphQLEndpoint` and `Neo4jServer` too so I opted to do that instead. Would you prefer I add it to `FederationMember`?

KentThang:
Latest commit resolves the comments except for the one that I am unsure about. Did you also see my question here?

> The reason I've made it into a draft is that I am not sure how I should make the XRequestProcessorImplTest tests. My main idea was to verify that the authentication information is actually propagated from the federation member to the request processor and ultimately added to the HTTP request. But to do this end-to-end would likely require setting up a local HTTP server or some form of HTTP mocking, which feels somewhat heavyweight for these tests.
> 
> So I'm not sure if the existing BearerAuthenticationInformationTest and FederationDescriptionReaderTest provide enough coverage for the authentication-specific functionality or if the request processor tests should be extended with some lightweight way of intercepting/inspecting the generated HTTP requests.

hartig:
> Latest commit resolves the comments except for the one that I am unsure about.

Thanks. I will take a look.

> Did you also see my question here?
> 
> > The reason I've made it into a draft is that I am not sure how I should make the XRequestProcessorImplTest tests. My main idea was to verify that the authentication information is actually propagated from the federation member to the request processor and ultimately added to the HTTP request. But to do this end-to-end would likely require setting up a local HTTP server or some form of HTTP mocking, which feels somewhat heavyweight for these tests.
> > So I'm not sure if the existing BearerAuthenticationInformationTest and FederationDescriptionReaderTest provide enough coverage for the authentication-specific functionality or if the request processor tests should be extended with some lightweight way of intercepting/inspecting the generated HTTP requests.

I also don't know how this could be done in a simple / lightweight way. So, let's ignore it.

Having said that, please check manually that the functionality works end to end. That is, set up a SPARQL endpoint using GraphDB or something, add authentication to it, and then use HeFQUIN to run a query that accesses the endpoint (the query may contain a single `SERVICE` clause, just to access that one endpoint).

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/SPARQLEndpoint.java
hartig: > The thought crossed my mind but I wasn't sure if I we also wanted to add `getAuthenticationInformation()` to `GraphQLEndpoint` and `Neo4jServer` too so I opted to do that instead. Would you prefer I add it to `FederationMember`?

Yes, please.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/SPARQLRequestProcessorImpl.java
```diff
@@ -57,15 +57,42 @@ protected SolMapsResponse performRequestWithQueryExecutionHTTP( final SPARQLRequ
...
 		// see https://jena.apache.org/documentation/sparql-apis/#query-execution
 
+		// final Map<String, String> headers = new HashMap<>();
+		// headers.put( "User-Agent", BuildInfo.getUserAgent() );
+
+		// if ( fm.getAuthenticationInformation() != null )
+		// 	fm.getAuthenticationInformation().applyTo( headers );
+
+		// final QueryExecution qe;
+		// try {
```
hartig: I guess that this whole part here still needs to be cleaned up.

hartig (changes requested):
The changes look good. There is only one part where you seem to forgot to remove a commented copy of code.

Otherwise, I will wait for you to change the PR such that `getAuthenticationInformation()` is moved to the `FederationMember` interface.

KentThang:
> Having said that, please check manually that the functionality works end to end. That is, set up a SPARQL endpoint using GraphDB or something, add authentication to it, and then use HeFQUIN to run a query that accesses the endpoint (the query may contain a single SERVICE clause, just to access that one endpoint).

Took a second to setup the GraphDB SPARQL endpoint but it seems to work! Although I had to do some minor local changes for the test, changing "Bearer token" to "GDB token". 

This is the error message you get when the token is incorrect / missing:
```
1 org.apache.jena.sparql.engine.http.QueryExceptionHTTP: Unauthorized
StackTrace:
se.liu.ida.hefquin.engine.queryplan.executable.ExecOpExecutionException: Performing the request caused an exception.
        at se.liu.ida.hefquin.engine.queryplan.executable.impl.ops.ExecOpRequestSPARQL._execute(ExecOpRequestSPARQL.java:50)
        at se.liu.ida.hefquin.engine.queryplan.executable.impl.ops.NullaryExecutableOpBase.execute(NullaryExecutableOpBase.java:51)
        at se.liu.ida.hefquin.engine.queryplan.executable.impl.pushbased.PushBasedPlanThreadImplForNullaryOperator.produceOutput(PushBasedPlanThreadImplForNullaryOperator.java:30)
        at se.liu.ida.hefquin.engine.queryplan.executable.impl.pushbased.PushBasedPlanThreadImplBase.run(PushBasedPlanThreadImplBase.java:111)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: se.liu.ida.hefquin.federation.access.FederationAccessException: Getting the response to a data retrieval request caused an exception.
        at se.liu.ida.hefquin.engine.federation.access.utils.FederationAccessUtils.performRequest(FederationAccessUtils.java:162)
        at se.liu.ida.hefquin.engine.queryplan.executable.impl.ops.ExecOpRequestSPARQL._execute(ExecOpRequestSPARQL.java:47)
        ... 8 more
Caused by: java.util.concurrent.ExecutionException: HttpException: 401 Unauthorized
        at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
        at se.liu.ida.hefquin.engine.federation.access.utils.FederationAccessUtils.performRequest(FederationAccessUtils.java:156)
        ... 9 more
Caused by: HttpException: 401 Unauthorized
        at org.apache.jena.sparql.engine.http.QueryExceptionHTTP.rewrap(QueryExceptionHTTP.java:49)
        at org.apache.jena.sparql.exec.http.QueryExecHTTP.executeQuery(QueryExecHTTP.java:552)
        at org.apache.jena.sparql.exec.http.QueryExecHTTP.execRowSet(QueryExecHTTP.java:173)
        at org.apache.jena.sparql.exec.http.QueryExecHTTP.select(QueryExecHTTP.java:164)
        at org.apache.jena.sparql.exec.QueryExecutionAdapter.execSelect(QueryExecutionAdapter.java:108)
        at se.liu.ida.hefquin.federation.access.impl.reqproc.SPARQLRequestProcessorImpl.performRequestWithQueryExecutionHTTP(SPARQLRequestProcessorImpl.java:101)
        at se.liu.ida.hefquin.federation.access.impl.reqproc.SPARQLRequestProcessorImpl.performRequest(SPARQLRequestProcessorImpl.java:50)
        at se.liu.ida.hefquin.federation.access.impl.reqproc.SPARQLRequestProcessorImpl.performRequest(SPARQLRequestProcessorImpl.java:28)
        at se.liu.ida.hefquin.federation.access.impl.AsyncFederationAccessManagerImpl$1.get(AsyncFederationAccessManagerImpl.java:97)
        at se.liu.ida.hefquin.federation.access.impl.AsyncFederationAccessManagerImpl$1.get(AsyncFederationAccessManagerImpl.java:93)
        at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
        ... 3 more
```
Should I try to shorten it / make it more "user friendly"?

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/FederationMember.java
```diff
@@ -20,6 +21,9 @@ public interface FederationMember
 	 */
 	String getServiceURI();
 
+	/** Returns the authentication information of this federation member. */
```
hartig:
```suggestion
	/**
	 * Returns the authentication information of this federation member
	 * or {@code null} if there is no authentication information for this
	 * federation member (in which case it can be assumed that this
	 * federation member can be accessed without authentication).
	*/
```

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/impl/SPARQLEndpointImpl.java
```diff
@@ -3,25 +3,33 @@
...
 
 import se.liu.ida.hefquin.base.data.VocabularyMapping;
+import se.liu.ida.hefquin.federation.authentication.AuthenticationInformation;
 import se.liu.ida.hefquin.federation.members.SPARQLEndpoint;
 
 public class SPARQLEndpointImpl extends BaseForRDFBasedFederationMember
                                 implements SPARQLEndpoint
 {
 	protected final String url;
+	protected final AuthenticationInformation authInfo;
```
hartig: Instead of having such a member variable in each of these classes here, it makes sense to move it to the base class that all these classes derive from: `BaseForFederationMember`

The same should be done for the implementation of the `getAuthenticationInformation()` method.

hartig (changes requested):
Two more things in the comments.

Additionally, just adding `AuthenticationInformation` to `GraphQLEndpointImpl` and `Neo4JServerImpl` is not enough to have this information actually being used ;-) To implement this for GraphQL, extend the `performRequest` method of `GraphQLConnection` by adding `AuthenticationInformation` as a new argument (make it the third argument, between `url` and `connectionTimeout`). For Neo4j, you can extend the first `connect` function of `Neo4jConnectionFactory` by appending `AuthenticationInformation` as a second argument and, then, extending the first constructor of `Neo4jConnection` accordingly.

KentThang:
> For Neo4j, you can extend the first connect function of Neo4jConnectionFactory by appending AuthenticationInformation as a second argument and, then, extending the first constructor of Neo4jConnection accordingly.

Do we want `authInfo` to override the current `authHeader` if `authInfo != null` ?

KentThang:
Also I thought I could simply add a dependency to `hefquin-access` in the pom.xml of both `hefquin-graphqlconnector` and `hefquin-pgconnector` (so that I can import `AuthenticationInformation` in those relevant files) but apparently that causes a cyclic reference. Should I extract some code into a new module or maybe move some files from module to module?

hartig:
> > For Neo4j, you can extend the first connect function of Neo4jConnectionFactory by appending AuthenticationInformation as a second argument and, then, extending the first constructor of Neo4jConnection accordingly.
> 
> Do we want `authInfo` to override the current `authHeader` if `authInfo != null` ?

No. Given that `authInfo` and `authHeader` would come in via separate constructors, they are mutually exclusive: either one of them is set or the other, but never both (yet, another option is that none of them is set; i.e., both are `null`).

hartig:
> Also I thought I could simply add a dependency to `hefquin-access` in the pom.xml of both `hefquin-graphqlconnector` and `hefquin-pgconnector` (so that I can import `AuthenticationInformation` in those relevant files) but apparently that causes a cyclic reference. Should I extract some code into a new module or maybe move some files from module to module?

Ah, yes :-(  Didn't we come across this issue in another context before? (Or maybe that was with @keski?) My ideal scenario would be that `hefquin-graphqlconnector` and `hefquin-pgconnector` depend on `hefquin-access` instead of the other way around. I believe I had made some progress in this direction but I cannot immediately remember what is still causing the need for the current dependency direction. Can you please quickly investigate where in `hefquin-access` we use something from `hefquin-graphqlconnector` and from `hefquin-pgconnector`.

KentThang:
> Didn't we come across this issue in another context before?

I believe we encountered something similar, yes, but not that exact error. I was curious so I looked into the commit histories of the various pom.xml files and what I was thinking of was this: https://github.com/LiUSemWeb/HeFQUIN/pull/667. However, I'm not sure if that solution works here since that was an issue of files in the test folder not being accessible (I'll look into it some).

> Can you please quickly investigate where in hefquin-access we use something from hefquin-graphqlconnector and from hefquin-pgconnector.

Yes. It seems to be in multiple files in `hefquin-access`. It's needed in `PersistentCacheKey`, `GraphQLRequestProcessorImpl`, `Neo4jRequestProcessorImpl`, `RecordsResponseImpl`, `GraphQLRequest`, `RecordsResponse`, `FederationDescriptionReader`, `GraphQLEndpointImpl`, `GraphQLEndpoint`, `CacheTestBase`, `GraphQLRequestProcessorImplTest`, `AsyncFederationAccessManagerImplTest`, `FederationAccessManagerWithCacheTest` and `FederationTestBase`.

(Tested by commenting out the dependencies in the pom.xml for `hefquin-access` and looking at the errors)

hartig:
> I believe we encountered something similar, yes, but not that exact error. I was curious so I looked into the commit histories of the various pom.xml files and what I was thinking of was this: #667. However, I'm not sure if that solution works here since that was an issue of files in the test folder not being accessible (I'll look into it some).

No, that was something else.

> > Can you please quickly investigate where in hefquin-access we use something from hefquin-graphqlconnector and from hefquin-pgconnector.
> 
> [...]
> (Tested by commenting out the dependencies in the pom.xml for `hefquin-access` and looking at the errors)

Thanks! Can you please do a second such test: using `grep` for `import se.liu.ida.hefquin.engine.wrappers.graphql` and for `import se.liu.ida.hefquin.engine.wrappers.lpg` within the source files in `hefquin-access`, just for the sake of confirming the completeness of the list of dependent classes you extracted.

KentThang:
I'm not sure how you want it displayed so I'll just dump the outputs.
`grep` for `import se.liu.ida.hefquin.engine.wrappers.graphql`:
```
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLArgument;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.impl.GraphQLArgumentImpl;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.GraphQLQuery;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/cache/CacheTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.impl.GraphQLQueryImpl;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLArgument;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLEntrypoint;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLField;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLSchema;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.impl.GraphQLArgumentImpl;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.impl.GraphQLEntrypointType;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.impl.GraphQLFieldType;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.GraphQLQuery;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImplTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.impl.GraphQLQueryImpl;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithCacheTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.GraphQLQuery;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/access/impl/FederationAccessManagerWithCacheTest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.impl.GraphQLQueryImpl;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLEntrypoint;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLField;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLSchema;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.impl.GraphQLEntrypointType;
hefquin-access/src/test/java/se/liu/ida/hefquin/federation/FederationTestBase.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.impl.GraphQLFieldType;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/GraphQLRequest.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.GraphQLQuery;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/req/GraphQLRequestImpl.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.GraphQLQuery;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImpl.java:import se.liu.ida.hefquin.engine.wrappers.graphql.conn.GraphQLConnection;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImpl.java:import se.liu.ida.hefquin.engine.wrappers.graphql.conn.GraphQLConnectionException;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/GraphQLRequestProcessorImpl.java:import se.liu.ida.hefquin.engine.wrappers.graphql.query.GraphQLQuery;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java:import se.liu.ida.hefquin.engine.wrappers.graphql.GraphQLException;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java:import se.liu.ida.hefquin.engine.wrappers.graphql.GraphQLSchemaInitializer;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLSchema;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java:import se.liu.ida.hefquin.engine.wrappers.graphql.impl.GraphQLSchemaInitializerImpl;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/impl/GraphQLEndpointImpl.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLSchema;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/members/GraphQLEndpoint.java:import se.liu.ida.hefquin.engine.wrappers.graphql.data.GraphQLSchema;

```
and for import `se.liu.ida.hefquin.engine.wrappers.lpg`:
```
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/RecordsResponse.java:import se.liu.ida.hefquin.engine.wrappers.lpg.data.TableRecord;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/response/RecordsResponseImpl.java:import se.liu.ida.hefquin.engine.wrappers.lpg.data.TableRecord;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/Neo4jRequestProcessorImpl.java:import se.liu.ida.hefquin.engine.wrappers.lpg.Neo4jException;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/Neo4jRequestProcessorImpl.java:import se.liu.ida.hefquin.engine.wrappers.lpg.conn.Neo4jConnectionException;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/Neo4jRequestProcessorImpl.java:import se.liu.ida.hefquin.engine.wrappers.lpg.conn.Neo4jConnectionFactory;
hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/Neo4jRequestProcessorImpl.java:import se.liu.ida.hefquin.engine.wrappers.lpg.data.TableRecord;
```

hartig:
> > Having said that, please check manually that the functionality works end to end. That is, set up a SPARQL endpoint using GraphDB or something, add authentication to it, and then use HeFQUIN to run a query that accesses the endpoint (the query may contain a single SERVICE clause, just to access that one endpoint).
> 
> Took a second to setup the GraphDB SPARQL endpoint but it seems to work! Although I had to do some minor local changes for the test, changing "Bearer token" to "GDB token".

In other words, the current code does not actually work for GraphDB?

> This is the error message you get when the token is incorrect / missing:
> 
> ```
> 1 org.apache.jena.sparql.engine.http.QueryExceptionHTTP: Unauthorized
> StackTrace:
> se.liu.ida.hefquin.engine.queryplan.executable.ExecOpExecutionException: Performing the request caused an exception.
>         at se.liu.ida.hefquin.engine.queryplan.executable.impl.ops.ExecOpRequestSPARQL._execute(ExecOpRequestSPARQL.java:50)
> ```
> [...]
>
> Should I try to shorten it / make it more "user friendly"?

I assume that this is printed in lines 180-187 of `RunQueryWithoutSrcSel`, right?

An idea that I had for this the other day was to change this behavior slightly: If the CLI program is called with the `--debug` argument, then for each of the exceptions, the message of the exception itself (not of its root-cause exception as used at the moment) is printed together with the complete stack trace (as done at the moment), Without the `--debug` argument, only the message of the exception itself would be printed. Can you please implement this idea in a separate PR?

hartig:
> I'm not sure how you want it displayed so I'll just dump the outputs.

Okay, I see now that resolving the potential cyclic dependency between the modules (`hefquin-access` versus `hefquin-graphqlconnector` and `hefquin-pgconnector`) is a bit more complicated and better done outside of this PR. As a consequence, actually implementing the use of `AuthenticationInformation` for GraphQL and Neo4j (as requested in [my earlier comment above](https://github.com/LiUSemWeb/HeFQUIN/pull/680#pullrequestreview-5040072119)) should also be done outside of this PR. In other words, after this PR has been merged, please create the following series of PRs:
1. As a preparation, we the first constructor of `AsyncFederationAccessManagerImpl` needs to be extended with the `defaultParallelRequestLimitPerServer` argument that the third constructor of `AsyncFederationAccessManagerImpl` has (it should become the second argument of the first constructor, i.e., directly after `execService`). After that, the second and third constructor should be removed, and then the config file (`DefaultConfDesc.ttl`) needs to be changed such that it uses the first (and now only) constructor.
2. Now it should be possible to implement everything needed to switch the direction of the dependency between `hefquin-access` and `hefquin-pgconnector` (i.e., the latter should depend on the former). To this end, the following classes/interfaces need to be moved from `hefquin-access` to `hefquin-pgconnector`: `Neo4jRequestProcessorImpl`, `RecordsResponse`, and `RecordsResponseImpl`.
3. The next PR can then be about implementing everything needed to switch the direction of the dependency between `hefquin-access` and `hefquin-graphqlconnector`. This will be similar to the case of `hefquin-pgconnector` but will also be more complicated because `FederationDescriptionReader` (in `hefquin-access`) uses classes that are in `hefquin-graphqlconnector` (and should remain there). We have to think about a solution to this problem.
4. Once all this is done, it should be possible to implement the use of `AuthenticationInformation` for GraphQL and Neo4j as requested in [my earlier comment above](https://github.com/LiUSemWeb/HeFQUIN/pull/680#pullrequestreview-5040072119).

Please copy this plan into a separate issue in which we can keep track of these four implementation steps.

KentThang:
Latest commit moves `authInfo` to `BaseForFederationMember`. Neo4j and GraphQL does not make use of it at the moment (will be done in #695 as per discussion).

hartig (approved):

Follow ups:
1. #695 (Implement the use of AuthenticationInformation for GraphQL and Neo4j), already started in PR #697 
2. More user-friendly printing of exceptions / use of the `--debug` argument, as outlined in the following comment: https://github.com/LiUSemWeb/HeFQUIN/pull/680#issuecomment-5470435688
3. Support for the HTTP Basic Authentication scheme (which will then close issue #669)

Notice that 1 and 2 can be done in parallel. For 3, it is probably better to do it after #695 has been addressed; alternatively, 3 can be split into a part for the RDF-based and REST fed.members and a part for GraphQL + Neo4j, where the first part can be done already now (exactly like in this PR here) and the second part after #695.

KentThang:
> In other words, the current code does not actually work for GraphDB?

Ah, I didn't see this comment. No, it does not. Should another issue/PR be made for this?

> I assume that this is printed in lines 180-187 of RunQueryWithoutSrcSel, right?

Yes, exactly.

> An idea that I had for this the other day was to change this behavior slightly: If the CLI program is called with the --debug argument, then for each of the exceptions, the message of the exception itself (not of its root-cause exception as used at the moment) is printed together with the complete stack trace (as done at the moment), Without the --debug argument, only the message of the exception itself would be printed. Can you please implement this idea in a separate PR?

Yes, I'll get on that!

hartig:
> > In other words, the current code does not actually work for GraphDB?
> 
> Ah, I didn't see this comment. No, it does not. Should another issue/PR be made for this?

Yes. I suggest to create an issue first, mentioning the goal (namely, being able to authenticate via the authentication mechanism of GraphDB). After that, please work on a comment for that issue in which you describe how you think we can achieve the goal (i.e., what needs to be implemented/extended?)

Additionally, from your initial analysis of the authentication methods supported by the different triple stores, do you know whether there is any triple store that supports the Bearer token scheme as implemented in this PR? If so, it would be useful to test the functionality of this PR quickly with a SPARQL endpoint set up via that triple store.

KentThang:
> Additionally, from your initial analysis of the authentication methods supported by the different triple stores, do you know whether there is any triple store that supports the Bearer token scheme as implemented in this PR? If so, it would be useful to test the functionality of this PR quickly with a SPARQL endpoint set up via that triple store.

This wasn't in the initial document but I used something called Fuseki from Jena (to set up a SPARQL endpoint with bearer auth) to test the functionality of this PR and it works as expected!

hartig:
> > Additionally, from your initial analysis of the authentication methods supported by the different triple stores, do you know whether there is any triple store that supports the Bearer token scheme as implemented in this PR? If so, it would be useful to test the functionality of this PR quickly with a SPARQL endpoint set up via that triple store.
> 
> This wasn't in the initial document but I used something called Fuseki from Jena (to set up a SPARQL endpoint with bearer auth) to test the functionality of this PR and it works as expected!

Great! Thanks for testing it!!

PR #687, opened 2026-08-24
683 add integration test for example queries

### LINKED ISSUE #683: Add integration test for example queries
We should add an integration test that can be used to verify that all example queries (`examples/`) execute successfully with the default HeFQUIN configuration and the example federation description.

This would enable us to quickly check if changes to the code base or configuration silently breaks the examples used to demonstrate HeFQUIN.

keski:


hartig (approved):
Looks good! I will merge this PR and add an entry to the changelog.

PR #690, opened 2026-08-25
Handle non-SPARQL federation members in creation of mreq operator

### LINKED ISSUE #684: Creation of mreq operator should not fail if some fed.members are not SPARQL endpoints
The newly added multi-request operator (see PR #677) works only if all federation members assigned to it are SPARQL endpoints. The current implementation in (in PR #677) simply assumes that this is the case and throws an exception if it is not. The TODO is to change this behavior such that no exception is thrown.

The place where the exception is thrown is within the `createPlanForServicePatternWithValues` method of `ServiceClauseBasedSourcePlannerImpl` (line 329). This method takes an `OpServiceWithValues` object, turns it into a `LogicalOpMultiRequest` (the new operator), and returns a `LogicalPlan` with this operator. The idea to fix the issue would be as follows. Instead of creating just one set of federation members (as done at the moment), create two such sets: one with every relevant fed.member that is a SPARQL endpoint and one with all the other relevant fed.members. Then, for the first set, the logical plan with the multi-request operator is created (as currently done). Next, if the second set is empty, this plan with the multi-request operator can be returned (as currently done). If the second set is not empty, the plan with the multi-request operator is collected into a list and, for every federation member of the second set, a logical plan with an ordinary request operator is created (like in the `createPlanForServicePattern` method) and also collected into the list. This list is then used as the subplans of a plan with a multi-union operator.

KentThang:
Closes #684.

### DIFF
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
@@ -341,18 +341,22 @@ public class ServiceClauseBasedSourcePlannerImpl extends SourcePlannerBase
 				SPARQLfms.add(fm);
 		}
 
-		final SPARQLGraphPattern p = new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
-		final SPARQLRequest req = new SPARQLRequestImpl( p, null, mayReduce );
 
 		final List<LogicalPlan> lplansList = new ArrayList<>(nonSPARQLfms.size() + 1);
 
 		if ( SPARQLfms.size() > 1 ) {
+			final SPARQLGraphPattern p = new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
+			final SPARQLRequest req = new SPARQLRequestImpl( p, null, mayReduce );
 			final Var var = Var.alloc( jenaOp.getService() );
 			final LogicalOpMultiRequest op = new LogicalOpMultiRequest(req, var, SPARQLfms);
+
 			lplansList.add( new LogicalPlanWithNullaryRootImpl(op, null) );
 		}
 		else if ( SPARQLfms.size() == 1 ) {
+			final SPARQLGraphPattern p = new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
+			final SPARQLRequest req = new SPARQLRequestImpl( p, null, mayReduce );
 			final LogicalOpRequest<?,?> op = new LogicalOpRequest<>(SPARQLfms.iterator().next(), mayReduce, req);
+
 			lplansList.add( new LogicalPlanWithNullaryRootImpl(op, null) );
 		}
 
@@ -364,9 +368,9 @@ public class ServiceClauseBasedSourcePlannerImpl extends SourcePlannerBase
 				// We create a separate SPARQLGraphPattern and SPARQLRequest object
 				// for every request operator created here. This is to ensure that potential
 				// plan rewriting (e.g., projection push down) that is intended for one of the
-				// of these operators is not accidentally affecting other of these operators. 
+				// of these operators is not accidentally affecting other of these operators.
 				final SPARQLGraphPattern p =  new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
-				final SPARQLRequest req = new SPARQLRequestImpl( p1, null, mayReduce );
+				final SPARQLRequest req = new SPARQLRequestImpl( p, null, mayReduce );
 				final LogicalOpRequest<?,?> reqOp = new LogicalOpRequest<>(ep, mayReduce, req);
 				lplansList.add( new LogicalPlanWithNullaryRootImpl(reqOp, null) );
 			}
--- a/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
+++ b/hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
@@ -361,17 +361,20 @@ public class ServiceClauseBasedSourcePlannerImpl extends SourcePlannerBase
 				if ( ep.getNumberOfParameters() != 0 )
 					throw new IllegalArgumentException( "Invalid SERVICE clause: missing PARAMS for " + ep.toString() );
 
-				final SPARQLGraphPattern p1 =  new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
-				final SPARQLRequest req1 = new SPARQLRequestImpl( p1, null, mayReduce );
-				final LogicalOpRequest<?,?> reqOp = new LogicalOpRequest<>(ep, mayReduce, req1);
+				// We create a separate SPARQLGraphPattern and SPARQLRequest object
+				// for every request operator created here. This is to ensure that potential
+				// plan rewriting (e.g., projection push down) that is intended for one of the
+				// of these operators is not accidentally affecting other of these operators. 
+				final SPARQLGraphPattern p =  new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
+				final SPARQLRequest req = new SPARQLRequestImpl( p1, null, mayReduce );
+				final LogicalOpRequest<?,?> reqOp = new LogicalOpRequest<>(ep, mayReduce, req);
 				lplansList.add( new LogicalPlanWithNullaryRootImpl(reqOp, null) );
 			}
 			else
 				lplansList.add( createPlan( jenaOp.getSubOp(), mayReduce, fm ) );
 		}
 
-		if ( lplansList.size() == 1 )
-			return lplansList.get(0);
+		assert lplansList.size() > 1;
 
 		return new LogicalPlanWithNaryRootImpl(LogicalOpMultiwayUnion.getInstance(), null, lplansList);
 	}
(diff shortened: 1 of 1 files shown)


hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
```diff
@@ -322,20 +322,39 @@ protected LogicalPlan createPlanForServicePatternWithValues(
...
+		final Set<FederationMember> nonSPARQLfms = new HashSet<>();
 		for ( final Node n : jenaOp.getPossibleValues() ) {
 			final FederationMember fm = ctx.getFederationCatalog().getFederationMemberByURI( n.getURI() );
 			if ( ! (fm instanceof SPARQLEndpoint) )
-				throw new IllegalArgumentException( "VALUES-extended SERVICE clause with a federation member that is not a SPARQL endpoint (service URI: " + n.toString() + ")" );
-
-			fms.add(fm);
+				nonSPARQLfms.add(fm);
+			else
+				SPARQLfms.add(fm);
```
hartig: Switch the order and remove the negation (`!`) from the corresponding if-condition above, which makes the code a bit nicer to read ;-).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
```diff
@@ -322,20 +322,39 @@ protected LogicalPlan createPlanForServicePatternWithValues(
...
 		}
 
-		final SPARQLGraphPattern p =  new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
+		final SPARQLGraphPattern p = new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
 		final SPARQLRequest req = new SPARQLRequestImpl( p, null, mayReduce );
 		final Var var = Var.alloc( jenaOp.getService() );
-		final LogicalOpMultiRequest op = new LogicalOpMultiRequest(req, var, fms);
-		return new LogicalPlanWithNullaryRootImpl(op, null);
+		final LogicalOpMultiRequest op = new LogicalOpMultiRequest(req, var, SPARQLfms);
```
hartig: This operator should be created only if `SPARQLfms` is not empty.

Additionally, as a special case, if `SPARQLfms` contains only one SPARQL endpoint, then there is also no need for a multi-request operator. Instead, in this case, you can create a (sub)plan) with an ordinary request operator for `req`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
```diff
@@ -322,20 +322,39 @@ protected LogicalPlan createPlanForServicePatternWithValues(
...
+
+		if ( ! nonSPARQLfms.isEmpty() ) {
+			final List<LogicalPlan> lplansList = new ArrayList<>(nonSPARQLfms.size() + 1);
```
hartig: The creation of `lplansList` should be moved outside if this if-block. In fact, you forgot to add the multi-request operator into this list.

As another minor optimization, in the case that `jenaOp.getPossibleValues()` has a size of 1, we can avoid creating this list and the union operator altogether. You can cover this special case in a new if-block directly before creating `SPARQLfms` and `nonSPARQLfms` above.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
```diff
@@ -322,20 +322,39 @@ protected LogicalPlan createPlanForServicePatternWithValues(
...
+			lplansList.add( new LogicalPlanWithNullaryRootImpl(op, null) );
+			for ( final FederationMember fm : nonSPARQLfms ) {
+				if ( fm instanceof WrappedRESTEndpoint ep ) {
+					if ( ep.getNumberOfParameters() != 0 )
+						throw new IllegalArgumentException( "Invalid SERVICE clause: missing PARAMS for " + ep.toString() );
+
+					final LogicalOpRequest<?,?> reqOp = new LogicalOpRequest<>(ep, mayReduce, req);
```
hartig: It can be a problem to reuse `req` (and its `p`) here. Better create a new `SPARQLRequestImpl` with a new `GenericSPARQLGraphPatternImpl2` for this case.

The problem with reusing `req` and `p` for these requests is that. later during logical query planning, we may push a filter or projection variables into the request of the multi-request operator but we may not want to do that automatically also for the request of the request operators created here.

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
KentThang: > In fact, you forgot to add the multi-request operator into this list.

Did I? Is it not what I do directly below (on line 342)?

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
hartig: You forgot this request.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
```diff
@@ -322,20 +324,56 @@ protected LogicalPlan createPlanForServicePatternWithValues(
...
-
-			fms.add(fm);
+				nonSPARQLfms.add(fm);
+			else
+				SPARQLfms.add(fm);
 		}
 
-		final SPARQLGraphPattern p =  new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
+		final SPARQLGraphPattern p = new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
 		final SPARQLRequest req = new SPARQLRequestImpl( p, null, mayReduce );
```
hartig: Move these two lines into if / else-if blocks where `req` is actually used. There is no need to create `p` and `req` here if these objects are not actually used.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
```diff
@@ -322,20 +324,56 @@ protected LogicalPlan createPlanForServicePatternWithValues(
...
+		}
+
+		for ( final FederationMember fm : nonSPARQLfms ) {
+			if ( fm instanceof WrappedRESTEndpoint ep ) {
+				if ( ep.getNumberOfParameters() != 0 )
+					throw new IllegalArgumentException( "Invalid SERVICE clause: missing PARAMS for " + ep.toString() );
+
+				final SPARQLGraphPattern p1 =  new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
+				final SPARQLRequest req1 = new SPARQLRequestImpl( p1, null, mayReduce );
+				final LogicalOpRequest<?,?> reqOp = new LogicalOpRequest<>(ep, mayReduce, req1);
```
hartig: After you have applied my previous comment (about moving the creation of `p` and `req` above), you can rename the variables here. Additionally, it would be good to add a comment related to the fact that `p` and `req` are created again here.

```suggestion
				// We create a separate SPARQLGraphPattern and SPARQLRequest object
				// for every request operator created here. This is to ensure that potential
				// plan rewriting (e.g., projection push down) that is intended for one of the
				// of these operators is not accidentally affecting other of these operators. 
				final SPARQLGraphPattern p =  new GenericSPARQLGraphPatternImpl2( jenaOp.getSubOp() );
				final SPARQLRequest req = new SPARQLRequestImpl( p1, null, mayReduce );
				final LogicalOpRequest<?,?> reqOp = new LogicalOpRequest<>(ep, mayReduce, req);
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
```diff
@@ -322,20 +324,56 @@ protected LogicalPlan createPlanForServicePatternWithValues(
...
+				lplansList.add( new LogicalPlanWithNullaryRootImpl(reqOp, null) );
+			}
+			else
+				lplansList.add( createPlan( jenaOp.getSubOp(), mayReduce, fm ) );
+		}
+
+		if ( lplansList.size() == 1 )
+			return lplansList.get(0);
```
hartig: This if condition should never be true because this case is already covered by the if-block that you put at the very top of the method. hence, we can simply turn this into a unit test check.
```suggestion
		assert lplansList.size() > 1;
```

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/srcsel/ServiceClauseBasedSourcePlannerImpl.java
hartig: This one is still open. You probably didn't see my previous reminder as the thread was already marked as resolved.

KentThang: Ah, I misunderstood and thought it was another if-condition...

Been very sloppy after the vacation, I'll sharpen up! Sorry!

hartig (approved):
Looks good now.

PR #693, opened 2026-08-28
Add ProjectRemoval and tests

### LINKED ISSUE #692: Remove project operators after project push down if not needed
Currently, `ProjectPushDown` may end up leaving several project operators all over the plan. Since this is not always useful (see below), there needs to be a post processing step that cleans up the plan after `ProjectPushDown`.

Background: Pushing down project operators is useful for two reasons:
1. If a projection ends up being pushed down all the way _into_ a request operator, this request operator retrieves less data.
2. If the query uses `SELECT DISTINCT`, duplicate removal through our `mayReduce` flags may become more effective in some cases if there are project operators in the plan that project away unnecessary variables early.

Notice that, in the context of the first of these reasons, project operators have no use once projection push down has been completed; they are more like an intermediate artifacts of the project push down process. Yet, keeping them can be useful for the second reason. However, if the query does not use `DISTINCT`, then the second reason is irrelevant; in this case, the project operators should be removed after project push down has been completed.

Consequently, this issue is about creating a new `HeuristicForLogicalOptimization` called `ProjectRemoval` that traverses the plan and removes all project operators for which the `mayReduce` flag is not set (i.e., is `false`).

KentThang:
Closes #692.

### DIFF
--- a/config/DefaultConfDescr.ttl
+++ b/config/DefaultConfDescr.ttl
@@ -67,6 +67,7 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.MergeRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.FilterPushDown" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectPushDown" ]
+        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.MergeRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.CardinalityBasedJoinOrderingWithRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemoveUnnecessaryL2gAndG2l" ]
@@ -75,7 +76,6 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PushJoinUnderUnionWithRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PullUpLtgOverUnion" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemovePairsOfG2lAndL2g" ]
-        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
     ) .
 
 _:bPhysicalOptimizer rdf:type ec:PhysicalOptimizer ;
--- a/config/DefaultConfDescr.ttl
+++ b/config/DefaultConfDescr.ttl
@@ -75,6 +75,7 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PushJoinUnderUnionWithRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PullUpLtgOverUnion" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemovePairsOfG2lAndL2g" ]
+        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
     ) .
 
 _:bPhysicalOptimizer rdf:type ec:PhysicalOptimizer ;
--- a/config/DefaultConfDescrExtended.ttl
+++ b/config/DefaultConfDescrExtended.ttl
@@ -88,6 +88,7 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.MergeRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.FilterPushDown" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectPushDown" ]
+        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.MergeRequests" ]
         #[ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.GreedyBasedReordering" ;
         #  ec:constructorArguments (_:bFormulaForSelectivity) ]
@@ -98,7 +99,6 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PushJoinUnderUnionWithRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PullUpLtgOverUnion" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemovePairsOfG2lAndL2g" ]
-        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
     ) .
 
 _:bListOfPhysicalOpFactories
--- a/config/DefaultConfDescrExtended.ttl
+++ b/config/DefaultConfDescrExtended.ttl
@@ -98,6 +98,7 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PushJoinUnderUnionWithRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PullUpLtgOverUnion" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemovePairsOfG2lAndL2g" ]
+        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
     ) .
 
 _:bListOfPhysicalOpFactories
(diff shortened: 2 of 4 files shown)


config/DefaultConfDescr.ttl
```diff
@@ -75,6 +75,7 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PushJoinUnderUnionWithRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PullUpLtgOverUnion" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemovePairsOfG2lAndL2g" ]
+        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
```
hartig: It makes more sense to put this rewriting rule directly after `ProjectPushDown`.

config/DefaultConfDescrExtended.ttl
```diff
@@ -98,6 +98,7 @@ _:bListOfHeuristics
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PushJoinUnderUnionWithRequests" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.PullUpLtgOverUnion" ]
         [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemovePairsOfG2lAndL2g" ]
+        [ ec:javaClassName "se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.ProjectRemoval" ]
```
hartig: It makes more sense to put this rewriting rule directly after `ProjectPushDown`.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,61 @@
...
+
+import se.liu.ida.hefquin.engine.queryplan.logical.LogicalOperator;
+import se.liu.ida.hefquin.engine.queryplan.logical.LogicalPlan;
+import se.liu.ida.hefquin.engine.queryplan.logical.LogicalPlanUtils;
+import se.liu.ida.hefquin.engine.queryplan.logical.impl.LogicalOpProject;
+import se.liu.ida.hefquin.engine.queryproc.LogicalOptimizationException;
+import se.liu.ida.hefquin.engine.queryproc.QueryProcContext;
+import se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicForLogicalOptimization;
+
+public class ProjectRemoval implements HeuristicForLogicalOptimization
```
hartig, line 37: Please add a Javadoc comment. You can use text from the description of the issue (#692).

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,61 @@
...
+{
+
```
hartig:
```suggestion
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,61 @@
...
+			throws LogicalOptimizationException
+	{
+		return apply(inputPlan);
+	}
+
+	public LogicalPlan apply( final LogicalPlan inputPlan ) {
+		final int numberOfSubPlans = inputPlan.numberOfSubPlans();
+		if ( numberOfSubPlans == 0 ) {
+			return removeProjectIfNecessary(inputPlan);
+		}
```
hartig, line 51: If `numberOfSubPlans == 0`, then the operator of `inputPlan` cannot be a project operator and, thus, calling `removeProjectIfNecessary` does not make sense in this case.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,61 @@
...
+		final LogicalPlan newPlan;
+		final LogicalOperator rootOp = inputPlan.getRootOperator();
+		if ( noChanges ) {
+			newPlan = inputPlan;
+		}
+		else {
+			newPlan = LogicalPlanUtils.createPlanWithSubPlans(
+					rootOp,
+					null,
+					newSubPlans);
```
hartig: Creating this Java object should be avoided in case that `rootOp` is a project operator.

hefquin-engine/src/test/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemovalTest.java
```diff
@@ -0,0 +1,202 @@
...
+{
+	@Test
+	public void removeUnnecessaryProject() {
+		// A project with mayReduce = false is removed.
+
+		// set up
+		final Var v1 = Var.alloc("x");
+		final Var v2 = Var.alloc("y");
+
+		// Left request produces x
```
hartig: "Left request"?

Also, "x" may be a bit confusing, given that the program variable is called `v1`. I suggest to use the same name for the program variables as for the underlying query variables.

These comments apply to the next unit test as well.

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
KentThang: I added an if statement that returns the subplan of the plan if `rootOp` is a project operator.

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * A heuristic that removes project operators for which the {@code mayReduce}
+ * flag is not set.
+ *
+ * <p>Project push down may leave project operators in the plan after the
+ * projection has been pushed as far down as possible. Such operators are
+ * unnecessary when they do not contribute to duplicate reduction. Therefore,
+ * this heuristic removes project operators with {@code mayReduce = false},
+ * while retaining those with {@code mayReduce = true}.</p>
```
hartig:
```suggestion
 * <p>
 * This heuristic is meant to be applied after {@link ProjectPushDown} which
 * aims to push projections as far down in the plan as possible and, in this
 * process, may leave several project operators all over the plan.
 * Such operators are unnecessary when they do not contribute to duplicate reduction.
 * Therefore, this heuristic removes project operators with {@code mayReduce = false},
 * while retaining those with {@code mayReduce = true}.
 * <p>
 * More background: Pushing down project operators is useful for two reasons:
 * <ol>
 * <li>If a projection ends up being pushed down all the way into a request operator,
 *      this request operator retrieves less data.</li>
 *      <li> If the query uses {@code SELECT DISTINCT}, duplicate removal through our
 *      {@code mayReduce} flags may become more effective in some cases if there are
 *      project operators in the plan that project away unnecessary variables early.</li>
 * </ol>
 * Notice that, in the context of the first of these reasons, project operators have no
 * use once projection push down has been completed; they are more like an intermediate
 * artifacts of the project push down process. Yet, keeping them can be useful for the
 * second of the reasons mentioned above. However, if the query does not use DISTINCT,
 * then the second reason is irrelevant; in this case, the project operators should be
 * removed after project push down has been completed.
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,74 @@
...
+			if ( ! newSubPlans[i].equals(oldSubPlan) ) {
+				noChanges = false;
+			}
+		}
+
+		final LogicalPlan newPlan;
+		final LogicalOperator rootOp = inputPlan.getRootOperator();
+
+		if ( rootOp instanceof LogicalOpProject op && !op.mayReduce() )
+			return newSubPlans[0];
```
hartig: Move this part before you create the `newSubPlans` array because, in this case, we can avoid creating that array altogether. After moving, the if-block here can, of course, not return `newSubPlans[0]`; instead, it should return `apply( inputPlan.getSubPlan(0) )`

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,74 @@
...
+			newPlan = inputPlan;
+		}
+		else {
+			newPlan = LogicalPlanUtils.createPlanWithSubPlans(
+					rootOp,
+					null,
+					newSubPlans);
+		}
+
+		return removeProjectIfNecessary(newPlan);
```
hartig: It is not necessary anymore to call this function, because now this method here already handles the case of the root operator being a project operator. Therefore you can even drop the introduction of the `newPlan` variable and change the if-else-block above into the following.
```java
		if ( noChanges ) {
			return inputPlan;
		}

		return LogicalPlanUtils.createPlanWithSubPlans(rootOp, null, newSubPlans);
```

hartig (changes requested):

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,82 @@
...
+		return LogicalPlanUtils.createPlanWithSubPlans(rootOp, null, newSubPlans);
+	}
+
+	protected LogicalPlan removeProjectIfNecessary( final LogicalPlan plan ) {
+		if ( plan.getRootOperator() instanceof LogicalOpProject op && ! op.mayReduce() )
+			return plan.getSubPlan(0);
+
+		return plan;
+	}
+
```
hartig: Now this whole method can be removed ;-)

```suggestion
```

hefquin-engine/src/main/java/se/liu/ida/hefquin/engine/queryproc/impl/loptimizer/heuristics/ProjectRemoval.java
```diff
@@ -0,0 +1,82 @@
...
+ * Therefore, this heuristic removes project operators with {@code mayReduce = false},
+ * while retaining those with {@code mayReduce = true}.
+ * <p>
+ * More background: Pushing down project operators is useful for two reasons:
+ * <ol>
+ * <li>If a projection ends up being pushed down all the way into a request operator,
+ *      this request operator retrieves less data.</li>
+ *      <li> If the query uses {@code SELECT DISTINCT}, duplicate removal through our
+ *      {@code mayReduce} flags may become more effective in some cases if there are
+ *      project operators in the plan that project away unnecessary variables early.</li>
```
hartig:
```suggestion
 * <li>If a projection ends up being pushed down all the way into a request operator,
 *     this request operator retrieves less data.</li>
 * <li> If the query uses {@code SELECT DISTINCT}, duplicate removal through our
 *     {@code mayReduce} flags may become more effective in some cases if there are
 *     project operators in the plan that project away unnecessary variables early.</li>
```

hartig (approved):
Two minor edits remaining. I will apply them directly myself now and merge the PR afterwards.

PR #696, opened 2026-08-31
685 argument for cli programs to ignore the cache

### LINKED ISSUE #685: Argument for CLI programs to ignore the cache
When doing some (manual) tests for the new multi-request operator (PR #677), I noticed that it can be quite useful to be able to call the CLI program with an option to ignore/bypass the caches. That is, all requests issued when processing the current query should really be executed rather than being served from the caches. Yet, the responses received for these requests can be used to update the caches.

@keski can you please look into this (at some point after PR #590 is merged).

The implementation of this feature should be done such that it can be used also via the `hefquin-client` program, which means that it cannot be implemented simply as a flag given to the constructor of the fed.access manager. Instead, it must be implemented within the context of the current query execution. I have recently integrated the use of `QueryProcContext` objects that are passed around throughout the whole process and that, thus, seem a good foundation for implementing this feature.

As a starting point, here is some more info about `QueryProcContext`:
* The `QueryProcContext` objects are created via a `QueryProcContextBuilder`. To do this, the initial version of such a `QueryProcContextBuilder` can be obtained from the `HeFQUINEngine` instance via its `getQueryProcContextBuilder()` method.
* Next, the different setter methods of this `QueryProcContextBuilder` can be used to configure the `QueryProcContext` that is being built in the end. The `hefquin` program (`RunQueryWithoutSrcSel`) does this in the `exec()` method. The servlet does this in `createQueryProcContext` of `SparqlServlet`.
* Hence, the idea would be to add an `ignoreCaches()` method to the `QueryProcContext` interface (similar to its `skipExecution() method), add a `setIgnoreCaches` method to `QueryProcContextBuilder` (similar to it `setSkipExecution` method), wire everything together (as done for 'skip execution'), and then extend the CLI programs.

This way, you get access to the flag at the places where requests are issued during query processing (because all these places should have access to the `QueryProcContext` object, or an extended version called `QueryProcContextExt`).

The other part of the implementation is then to extend the fed.access manager such that it offers this new feature.

keski:
Adds support for bypassing data retrieval and cardinality caches during query execution. The new `--ignoreCache` and `--ignoreCardinalityCache` CLI options are propagated through `QueryProcContext` and handled by cache-aware `FederationAccessManager` methods, allowing the two to be bypassed independently during debugging. The cache is still updated with newly retrieved responses.


hartig:
Before looking in detail into the code of this PR, two general observations:
1. I would like to avoid duplicating the two methods in `FederationAccessManager` and, instead, add the `ignoreCache` argument to the two existing methods (last Thursday we talked about a similar issue regarding the different variants of the constructors in some classes). I am aware that this would mean that all calls of these methods will have to be adapted, resulting potentially in many additional changes in this PR. How bad would this be? I mean, how many other files do we have that use these methods but are not yet changed anyways by this PR?
2. When seeing `--ignoreCache` and `--ignoreCardinalityCache`, one may think that the former subsumes the latter (i.e., `--ignoreCache` ignores all of the cache whereas `--ignoreCardinalityCache` ignores only the cardinality part of the cache). To avoid this misinterpretation, I suggest to rename `--ignoreCache` to `--ignoreRetrievalCache` (or you have another idea for an better name of this argument).

keski:
> * I would like to avoid duplicating the two methods in `FederationAccessManager` and, instead, add the `ignoreCache` argument to the two existing methods (last Thursday we talked about a similar issue regarding the different variants of the constructors in some classes). I am aware that this would mean that all calls of these methods will have to be adapted, resulting potentially in many additional changes in this PR. How bad would this be? I mean, how many other files do we have that use these methods but are not yet changed anyways by this PR?
> 

Not many additional file changes required really. I've removed the duplicated method in `FederationAccessManager`. 

> * When seeing `--ignoreCache` and `--ignoreCardinalityCache`, one may think that the former subsumes the latter (i.e., `--ignoreCache` ignores all of the cache whereas `--ignoreCardinalityCache` ignores only the cardinality part of the cache). To avoid this misinterpretation, I suggest to rename `--ignoreCache` to `--ignoreRetrievalCache` (or you have another idea for an better name of this argument).

I see your point, makes sense to rename it to avoid any confusion. `--ignoreRetrievalCache` works well. I've updated the code accordingly.

keski:
This branch currently has a merge conflict. Looking into it now.

hartig:
These merge conflicts are probably all related to my recent PR #706 (the one that we talked about at the beginning of today's meeting). One of the things that this PR does is to remove the declaration of `throws FederationAccessException` from some methods in some interfaces (and their implementing classes).

keski:
This PR is now ready for review.

@hartig

hartig (approved):
Looks great. Thanks!

Issue #695, opened 2026-08-31
Implement the use of AuthenticationInformation for GraphQL and Neo4j


KentThang:
After #680 has been merged, create following PRs:
* [x] As a preparation, the first constructor of `AsyncFederationAccessManagerImpl` needs to be extended with the `defaultParallelRequestLimitPerServer` argument that the third constructor of `AsyncFederationAccessManagerImpl` has (it should become the second argument of the first constructor, i.e., directly after `execService`). After that, the second and third constructor should be removed, and then the config file (`DefaultConfDesc.ttl`) needs to be changed such that it uses the first (and now only) constructor.
* [x] Now it should be possible to implement everything needed to switch the direction of the dependency between `hefquin-access` and `hefquin-pgconnector` (i.e., the latter should depend on the former). To this end, the following classes/interfaces need to be moved from `hefquin-access` to `hefquin-pgconnector`: `Neo4jRequestProcessorImpl`, `RecordsResponse`, and `RecordsResponseImpl`.
* [ ] The next PR can then be about implementing everything needed to switch the direction of the dependency between `hefquin-access` and `hefquin-graphqlconnector`. This will be similar to the case of `hefquin-pgconnector` but will also be more complicated because `FederationDescriptionReader` (in `hefquin-access`) uses classes that are in `hefquin-graphqlconnector` (and should remain there). We have to think about a solution to this problem.
* [ ] Once all this is done, it should be possible to implement the use of `AuthenticationInformation` for GraphQL and Neo4j

KentThang:
> Now it should be possible to implement everything needed to switch the direction of the dependency between hefquin-access and hefquin-pgconnector (i.e., the latter should depend on the former). To this end, the following classes/interfaces need to be moved from hefquin-access to hefquin-pgconnector: Neo4jRequestProcessorImpl, RecordsResponse, and RecordsResponseImpl.

I assume `Neo4jRequestProcessor` also needs to be moved? With these files moved, there is still the matter of resolving the import for `Neo4jRequestProcessor` in the various `FederationAccessManager` files. Or was the idea to keep `Neo4jRequestProcessor` (the interface) in `hefquin-access` somehow while having `Neo4jRequestProcessorImpl` in `hefquin-pgconnector`? I don't see how this is the solution either as `Neo4jRequestProcessor` also needs to import `RecordsResponse`, which wouldn't work.

hartig:
[...]

PR #697, opened 2026-08-31
Change constructors of AsyncFederationAccessManagerImpl

KentThang:
Addresses the first part of #695.


hartig (changes requested):
Looks good! Just one additional request: Can you please also extend `DefaultConfDescExtended.ttl` in the same way as you have extended `DefaultConfDesc.ttl`

hartig (approved):

PR #699, opened 2026-09-01
Change printing of errors in RunQueryWithoutSrcSel

KentThang:
Implements a more user-friendly printing of exceptions with the use of the `--debug` argument, as outlined in the following comment: https://github.com/LiUSemWeb/HeFQUIN/pull/680#issuecomment-5470435688

Weirdly enough, from inspecting `processArgs` of `ModGeneral`, giving the `--debug` flag does not set the `debug` boolean to true but instead, the verbose boolean. Therefore, `isVerbose()` is used.

### DIFF
--- a/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
+++ b/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunQueryWithoutSrcSel.java
@@ -179,10 +179,11 @@ public class RunQueryWithoutSrcSel extends CmdARQ
 			System.err.println();
 			for ( int i = 0; i < numberOfExceptions; i++ ) {
 				final Exception ex = exceptions.get(i);
-				final Throwable rc = getRootCause( ex );
-				System.err.println( (i + 1) + " " + rc.getClass().getName() + ": " + rc.getMessage() );
-				System.err.println( "StackTrace:" );
-				ex.printStackTrace( System.err );
+				System.err.println( "Exception " + (i + 1) + ": " + ex.getMessage() );
+				if ( isVerbose() ) {
+					System.err.println( "StackTrace:" );
+					ex.printStackTrace( System.err );
+				}
 				System.err.println();
 			}
 		}
@@ -216,25 +217,6 @@ public class RunQueryWithoutSrcSel extends CmdARQ
 			System.err.println( "Failed to load query: " + ex.getMessage() );
 			throw new TerminationException( 1 );
 		}
-	}
-
-	/**
-	 * Returns the root cause of a throwable by traversing the cause chain.
-	 *
-	 * This method follows the chain of {@code Throwable.getCause()} until it
-	 * reaches the deepest non-null cause. If the input {@code throwable} has no
-	 * cause, the method returns the throwable itself.
-	 *
-	 * @param throwable the throwable from which to extract the root cause
-	 * @return the root cause of the throwable, or {@code null} if {@code throwable}
-	 *         is {@code null}
-	 */
-	private static Throwable getRootCause( final Throwable throwable ) {
-		Throwable cause = throwable;
-		while ( cause.getCause() != null ) {
-			cause = cause.getCause();
-		}
-		return cause;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


hartig:
> Weirdly enough, from inspecting `processArgs` of `ModGeneral`, giving the `--debug` flag does not set the `debug` boolean to true but instead, the verbose boolean. Therefore, `isVerbose()` is used.

Weird! 

From a user-experience point of view, I would still prefer this functionality to be available via the `--debug` argument. I understand that this is the case at the moment (because of the strange implementation in `ModGeneral`), but I am worried that a future version of Jena might fix this implementation and we will not notice it because we are using the `isVerbose()` method.

Therefore, I propose you add the following method into `RunQueryWithoutSrcSel` (add it in between `getCommandName()` and `exec()`).
```java
	@Override
	protected void processModulesAndArgs() {
		super.processModulesAndArgs();

		// Fix because 'ModGeneral' currently sets the verbose flag instead
		// of the debug flag whenever the --debug argument is given.
		if ( isVerbose() )
			modGeneral.debug = true;
	}
```
This fix relies on the fact the `debug` in `ModGeneral` is currently marked as `public`, which I assume will be changed if the aforementioned issue is fixed. Hence, this way, we will notice if there is a change.

After adding this method, you can use `isDebug()` instead of `isVerbose()`   ;-)

KentThang:
> Weird!

Isn't it?

> Therefore, I propose you add the following method into RunQueryWithoutSrcSel (add it in between getCommandName() and exec()).

That's a cool solution! I've added it in the latest commit.

hartig (approved):

hartig:
As a follow up to this PR, can you check whether we can have the same functionality also in `hefquin-client` ( `RunHttpQuery`) for which the exceptions need to be send over from `hefquin-server` (`RunHeFQUINServer`). I am not sure whether that's possible. Please investigate.

hartig:
I have extended the corresponding documentation page with information about this new functionality.
https://liusemweb.github.io/HeFQUIN/doc/cli.html#--debug 

/cc @keski

KentThang:
> As a follow up to this PR, can you check whether we can have the same functionality also in hefquin-client ( RunHttpQuery) for which the exceptions need to be send over from hefquin-server (RunHeFQUINServer). I am not sure whether that's possible. Please investigate.

I've had the time to look at this today and concluded that it should be possible. However, while looking at this, I noticed that `getExceptions()` in `ServletUtils` doesn't actually include the full stack trace for each exception as the Javadoc for the method suggests. From my understanding, it takes the root cause and writes the root cause class name and message (similarily to how it was done before this PR) but not the stack trace. 

My loose outline for implementing this functionality for the server and client, would be to:
1. check for the flag in the client, if it was given, attach a header to the request
2. server checks for the flag and I could potentially send it as an arg to `getExceptions()`
3. in `getExceptions()`, if `debug == true`, add the full stack trace of the exception to the JsonArray
4. print it at client-side

How does that sound? Do we want to print it out in the exact same way (as mentioned in [your comment](https://github.com/LiUSemWeb/HeFQUIN/pull/680#issuecomment-5470435688)) and no longer print the message and classname of the root cause?

hartig:
> > As a follow up to this PR, can you check whether we can have the same functionality also in hefquin-client ( RunHttpQuery) for which the exceptions need to be send over from hefquin-server (RunHeFQUINServer). I am not sure whether that's possible. Please investigate.
> 
> I've had the time to look at this today and concluded that it should be possible. However, while looking at this, I noticed that `getExceptions()` in `ServletUtils` doesn't actually include the full stack trace for each exception as the Javadoc for the method suggests. From my understanding, it takes the root cause and writes the root cause class name and message (similarily to how it was done before this PR) but not the stack trace.

Indeed. That function doesn't do what its Javadoc comment promises. That needs to be fixed.

> My loose outline for implementing this functionality for the server and client, would be to:
> [...]
> How does that sound?

Sounds good!

> Do we want to print it out in the exact same way (as mentioned in [your comment](https://github.com/LiUSemWeb/HeFQUIN/pull/680#issuecomment-5470435688)) and no longer print the message and classname of the root cause?

Yes. Notice that this also means that the string that is always printed for an exception (i.e., not matter whether `--debug` is given or not) is the message of the exception itself, rather than its root cause exception as we have done so far.

PR #702, opened 2026-09-03
Switch dependency direction between hefquin-access and hefquin-pgconnector

KentThang:
Finishes step 2 of #695.


config/DefaultConfDescr.ttl
hartig: It should not be necessary to remove the request processor here. You only need to use the new package name.

The whole point the recent change of the constructor arguments of `AsyncFederationAccessManagerImpl` was to give us that flexibility that we can here refer to request processors that are not in the `hefquin-access` module. In fact, it should even be possible to mention classes here that implement the `RequestProcessor` interface but are not part of the HeFQUIN code base, as long as they are in the Java classpath when starting the CLI program that reads this config file.

Something else about this part, which I didn't notice in the previous PR that changed this part of the config file: The `ec:argumentTypeName` is not needed in such lists; the only thing that is considered when reading this part of the config file are the `ec:javaClassName` properties. Therefore, you can simplify this part as follows (please do the same also in the other config file).
```
  ec:elements (
    [ ec:javaClassName "se.liu.ida.hefquin.engine.wrappers.lpg.access.Neo4jRequestProcessorImpl" ]
    [ ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.reqproc.RESTRequestProcessorImpl" ]
    [ ec:javaClassName "se.liu.ida.hefquin.federation.access.impl.reqproc.GraphQLRequestProcessorImpl" ]
  )
```

hefquin-pgconnector/src/main/java/se/liu/ida/hefquin/engine/wrappers/lpg/reqproc/Neo4jRequestProcessor.java
```diff
@@ -1,4 +1,4 @@
-package se.liu.ida.hefquin.federation.access.impl.reqproc;
+package se.liu.ida.hefquin.engine.wrappers.lpg.reqproc;
```
hartig: I think it is better to name this subpackage `access`; i.e.,

`se.liu.ida.hefquin.engine.wrappers.lpg.access`

and collect all access-related interfaces and classes here. That is, in addition to `Neo4jRequestprocessor`, `Neo4jRequestprocessorImpl`, and `RecordsRetrievalProcessor`, you can move `RecordsResponse` and `RecordsResponseImpl` into it. And also move the two classes that are currently in `se.liu.ida.hefquin.engine.wrappers.lpg.conn`. And don't forget to do the same changes in the parallel directory structure for the unit tests.

hartig (changes requested):

hartig (approved):
Looks good now!

Issue #694, opened 2026-08-28
Extend the fed.desc. vocabulary to allow users to specify specific limits per federation member


KentThang:
We want to give users the option to specify a request limit per federation member, via a new property in the fed.desc. vocabulary.

This issue is first about extending the vocabulary, then parsing it properly in `FederationDescriptionReader` and potentially using it in relevant files to actually limit the amount of requests.

hartig:
PR #703 is merged, which means you can continue with the next step here.

KentThang:
I assume we want to store each request limit in each federation member by extending the `FederationMember` class with a `parallelRequestLimit` property and getter?

What would the entry point be for actually enforcing the limit or would this be something for Robin to work on?

hartig:
I'm not sure actually. How exactly do you envision can this then be used further down the line?

KentThang:
Maybe some sort of concurrency limiter / semaphore can be created and then used in the `RequestProcessors`? I'm not sure exactly the best place to have this mechanism though.

hartig:
We already have `HttpClientProvider` with its (static!) `client` methods that the different request processors use to get an `HttpClient` object via which they perform the requests. This `HttpClientProvider` has a (static!) `registerEndpointLimiter` method that can be used to register a limit per endpoint address, and the `HttpClient` objects that the `client` methods of `HttpClientProvider` return should take these limits into account. Perhaps we can use this existing functionality by having the `FederationDescriptionReader` call the `registerEndpointLimiter` method?

KentThang:
Ah! I overlooked this somehow even though I saw it while refactoring that variable in that previous issue...

I think it can definitely be used in `FederationDescriptionReader` and I'll make a PR for that soon.

PR #703, opened 2026-09-03
Extend feddesc vocab with parallelrequestlimit

KentThang:
First part of #694.


hefquin-vocabs/feddesc.ttl
```diff
@@ -131,6 +131,11 @@ fd:mappingConfiguration a owl:ObjectProperty ;
     rdfs:label "mapping configuration"@en ;
     rdfs:comment "Mapping configuration for translating non-RDF data models to RDF (e.g., GraphQL, PG)."@en .
 
+fd:parallelRequestLimit a owl:DatatypeProperty ;
+    rdfs:domain fd:FederationMember ;
+    rdfs:range  xsd:integer ;
+    rdfs:label "parallel request limit"@en ;
+    rdfs:comment "Parallel request limit per federation member"@en .
```
hartig:
```suggestion
    rdfs:comment "Specifies a limit on the number of requests that HeFQUIN sends in parallel to the federation member. If the limit is reached and further requests are to be sent, these requests are queued until the number of currently active requests drops below the limit."@en .
```

hartig (changes requested):

hartig (approved):

Issue #698, opened 2026-09-01
Extend AuthenticationInformation to function with GraphDB endpoints


KentThang:
The goal of this issue is to extend the authentication information so that GraphDB endpoints with authentication enabled, can be queried. Currently, the various `applyTo` functions in `BearerAuthenticationInformation` adds a `Bearer [token]` header to requests, but GraphDB endpoints use the prefix "GDB" instead of "Bearer".

The first step is to figure out how this can be implemented and thereafter, implement it.

KentThang:
The easiest and cleanest ideas I have for this extension are all related to the federation description. The idea I am leaning towards at the moment is to make the "Bearer" in `BearerAuthenticationScheme` customizable (maybe through `wotsec:name`) and the `FederationDescriptionReader` would then check if this property isn't null (in `parseAuthConfig()`), and if not, `BearerAuthenticationInformation` would be initialized with that value.

In the various `applyTo()` methods, instead of the hardcoded "Bearer", we could instead have something like:
```
@Override
public void applyTo( final Builder builder ) {
	builder.setHeader( "Authorization", authorizationScheme + token );
}
```
with
```
ex:Test a fd:RDFBasedFederationMember ;
      fd:serviceURI "http://localhost:7200/repositories/mysparqlendpoint"^^xsd:anyURI ;
      fd:interface [ a                         fd:FixedEndpointInterface ;
                     fd:supportedProtocol      fd:SPARQLProtocol ;
                     fd:endpointAddress        "http://localhost:7200/repositories/mysparqlendpoint"^^xsd:anyURI ;

                     td:hasSecurityConfiguration [ a                wotsec:BearerSecurityScheme ;
                                                   wotsec:in        "header" ;
                                                   wotsec:name "GDB"
                                                   fd:envarForToken "TOKEN123" ; ] ] .
```
Not sure exactly what we'd call it but `authorizationScheme` is just an example and if it isn't assigned in the fd, it should have "Bearer" as the default value.

However, with this change, I'm not sure calling the class `BearerAuthenticationInformation` would be entirely accurate. It could either be renamed or a new class could be made altogether with the same idea. What do you think?

hartig:
Thanks for writing this idea! Just one more request for it: Can you please post a snippet of Turtle that illustrates how this change/extension would look like in the federation description.

As for the name of the class, what about `TokenBasedAuthenticationInformation`?

KentThang:
[...]

PR #704, opened 2026-09-03
Extend feddesc vocab to include TokenBasedSecurityScheme

KentThang:
First PR of #698.


hefquin-vocabs/feddesc.ttl
```diff
@@ -63,6 +63,10 @@ fd:DataRetrievalProtocol a owl:Class ;
     rdfs:label "Data Retrieval Protocol"@en ;
     rdfs:comment "Instances of this class represent the different protocols to be used when interacting with the interface of a federation member."@en .
 
+fd:TokenBasedSecurityScheme a owl:Class ;
+    rdfs:label "Token-based Security Scheme"@en ;
+    rdfs:comment "A security scheme that uses a token to authenticate requests."@en .
```
hartig:
```suggestion
    rdfs:comment "A security scheme that uses a token to authenticate requests, similar to wotsec:BearerSecurityScheme but with the option to specify the name of the scheme to be mentioned in the request header (where wotsec:BearerSecurityScheme uses the fixed string 'Bearer'). To specify the name use the wotsec;name property. Otherwise, this class can be used exactly like wotsec:BearerSecurityScheme. Especially, every instance of this class must have an fd:envarForToken property and the value of the wotsec:in property, if given, must be the string 'header'."@en .
```

hartig (changes requested):
Here is an edit suggestion that extends the comment of the class, documenting in a bit more detail what this is about and how it should be used.

Additionally, please add an `rdfs:subClassOf` statement to this class, specifying that it is a sub-class of `wotsec:SecurityScheme` (for which you need to declare `wotsec:` as a prefix name in the preamble of this file, after `hydra:`).

If you want to generate the Java code for the vocabulary yourself, you simply need download Apache Jena as a ZIP package, unpack it, set the environment variable `JENA_HOME` to the root directory from the unpacked ZIP package, uncomment `procFDVocab` at the end of `hefquin-vocabs/build.sh`, and then run this build script.

hartig (approved):

PR #705, opened 2026-09-04
Implement TokenBasedAuthenticationInformation

### LINKED ISSUE #698: Extend AuthenticationInformation to function with GraphDB endpoints

KentThang:
Closes #698.

I opted to make a second constructor (rather than adjusting the previous one from `BearerAuthenticationInformation`). If an "authenticationScheme" isn't specified when creating a `TokenBasedAuthenticationInformation`, `Bearer` will be used. Not sure if it is preferred to only have one constructor and in the previous code that used `BearerAuthenticationInformation`, explicitly add "Bearer" as an argument. Let me know!

I have also confirmed with my own GraphDB SPARQL endpoint that this works as intended and that sending the wrong token / authentication scheme (e.g., not "GDB [token]" but instead maybe "GD [token]") fails.


hefquin-access/src/main/java/se/liu/ida/hefquin/federation/authentication/TokenBasedAuthenticationInformation.java
```diff
@@ -0,0 +1,68 @@
...
+ * HTTP request representations used by HeFQUIN.
+ */
+public class TokenBasedAuthenticationInformation implements AuthenticationInformation
+{
+	private final String authenticationScheme;
+	private final String token;
+
+	public TokenBasedAuthenticationInformation( final String token ) {
+		this( "Bearer", token );
+	}
```
hartig: I would prefer you remove this constructor and, in the part of `FederationDescriptionReader` where you use it, you pass the string `"Bearer"` explicit.y

hartig (changes requested):
Just a minor change request.

hartig (approved):

PR #707, opened 2026-09-08
Parse request limit and use registerEndpointLimiter to enforce it

### LINKED ISSUE #694: Extend the fed.desc. vocabulary to allow users to specify specific limits per federation member

KentThang:
Closes #694.

I wasn't sure if we wanted to parse the integer before or after the various `handleXInterface()` methods but opted for the first option in the end.

I was also not sure if we want to add any tests for this? I had initially made a test in `FederationDescriptionReaderTest` but seeing as how `parallelRequestLimit` isn't stored anywhere but the HashMap in `HttpClientProvider` which is 'protected', I'm not sure how to make any assertions. I could potentially add a sort of getter in `HttpClientProvider`? Something like this:
```
static Semaphore getEndpointLimiter(final String endpointAddress) {
    return LIMITERS_BY_ENDPOINT_ADDRESS.get(endpointAddress);
}
```
but I am not sure if it is desirable.


hartig:
> Closes #694.
> 
> I wasn't sure if we wanted to parse the integer before or after the various `handleXInterface()` methods but opted for the first option in the end.

Right choice.

> I was also not sure if we want to add any tests for this? I had initially made a test in `FederationDescriptionReaderTest` but seeing as how `parallelRequestLimit` isn't stored anywhere but the HashMap in `HttpClientProvider` which is 'protected', I'm not sure how to make any assertions. I could potentially add a sort of getter in `HttpClientProvider`? Something like this:
> 
> ```
> static Semaphore getEndpointLimiter(final String endpointAddress) {
>     return LIMITERS_BY_ENDPOINT_ADDRESS.get(endpointAddress);
> }
> ```
> 
> but I am not sure if it is desirable.

No, it is not. The `Semaphore` shouldn't become accessible outside of the `HttpClientProvider`.

I also don't see a good way to have this new functionality covered by a unit test. So, let's ignore that.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java
```diff
@@ -301,6 +334,10 @@ protected FederationMember handleTemplateInterface( final Resource iface,
 
 			final String uriTemplateString = ModelUtils.getSingleMandatoryProperty_XSDString(uriTemplate, HydraVocab.template);
 
+			if ( parallelRequestLimit != null ) {
+				HttpClientProvider.registerEndpointLimiter(uriTemplateString, parallelRequestLimit);
```
hartig: I have looked at the PR. Everything looks good except for this case, which is a tricky case.

The issue is that `uriTemplateString` can be an arbitrary URI template. Such a template may contain placeholders to be substituted when the template is instantiated, and these placeholders may be anywhere within the template. For instance, we may have a template `http://example.org/api/{type}/property/{pname}` in which `{type}` and `{pname}` are placeholders; this template may be instantiated to the URI `http//example.org/api/Student/property/LiUID` but also to `http//example.org/api/Course/property/CourseCode`. Registering the fed.member-specific limiter based on the template string doesn't work (the `resolveLimiter` function in `HttpClientProvider` wouldn't return the right `Semaphore` for the two instantiated URIs mentioned in my example).

Do you have an idea how to deal with this issue?

hartig (changes requested):

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/catalog/FederationDescriptionReader.java
KentThang: I assume it isn't as simple as just registering the endpoint limiter somewhere after the template has been instantiated?

The other idea I had would be to store the limiter in the federation member itself but again, I am not sure how it would be later used and it feels quite heavy for one specific case.

hartig: > The other idea I had would be to store the limiter in the federation member itself but again, I am not sure how it would be later used and it feels quite heavy for one specific case.

Your response just gave me the following idea: An option would be to do this only for instances of `RESTEndpoint`, where the implementation(s) of `RESTRequestProcessor` would then register the limiter after the template has been instantiated. For all other types of federation members, it can be done as already implemented in this PR.

To make this more efficient, the implementation(s) of `RESTRequestProcessor` should try to avoid to repeatedly register the limiter for the same endpoint, each time it processes a request for the same fed.member. To this end, it may maintain a `HashSet` of the endpoints for which it has already registered the limiter.

KentThang: Good idea! Latest commit tries to implement this, let me know if I interpreted it correctly.

hefquin-access/src/main/java/se/liu/ida/hefquin/federation/access/impl/reqproc/RESTRequestProcessorImpl.java
```diff
@@ -48,6 +51,10 @@ public RESTRequestProcessorImpl( final long connectionTimeout,
 	public StringResponse performRequest( final RESTRequest req,
 	                                      final RESTEndpoint fm ) {
 		final URI uri = req.getURI();
+
+		if ( fm.getParallelRequestLimit() != null && endpointsWithLimiters.add(uri.toString()) )
```
hartig: The second condition (after `&&`) is wrong. `HttpClientProvider` does not necessarily use the complete URI for registering the limiter. Use the method that `HttpClientProvider` has to reduce the URIs to the ones that it uses for registering limiters. If that method is currently not accessible outside of `HttpClientProvider`, then make it accessible.

hartig (changes requested):

hartig (approved):
Looks good now.

PR #708, opened 2026-09-09
Change exception printing and handle debug flag in hefquin-client

KentThang:
Follow up of #699.

Changes the exception printing of `RunHttpQuery` to no longer print the class name and message of the root cause but instead the exception message and, if the `--debug` argument was given, the full stack trace of the exception.

Since the root cause messages are no longer printed, two tests fail in `SparqlServletTest`. To resolve this, I changed the expected message in the assertion.

The relevant tests are: `testMissingFederationMember()` and `testUseOfHeFQUINParser()`

`InspectServlet` uses the `getExceptions` method but I don’t think it ever needs to consider the case in which the full stack trace is needed. Therefore, I just call it with `debug` set to `false`.


hefquin-service/src/main/java/se/liu/ida/hefquin/service/ServletUtils.java
```diff
@@ -55,59 +56,50 @@ public static ResultsFormat convert( final String mimeType ) {
...
+				final Exception ex = exceptions.get(i);
 				final StringWriter sw = new StringWriter();
-				sw.append( rootCause.getClass().getName() + ": " + rootCause.getMessage() );
+				final PrintWriter pw = new PrintWriter(sw);
+				pw.append( "Exception " + (i + 1) + ": " + ex.getMessage() );
+				if ( debug ) {
+					pw.append( "StackTrace:" );
+					ex.printStackTrace( pw );
+				}
 				list.add( sw.toString() );
```
hartig: Just for safety, it would be good to close `pw` after this line.

```suggestion
				list.add( sw.toString() );
				pw.close();
```

hefquin-service/src/main/java/se/liu/ida/hefquin/service/ServletUtils.java
```diff
@@ -55,59 +56,50 @@ public static ResultsFormat convert( final String mimeType ) {
...
+	public static JsonArray getExceptions( final List<Exception> exceptions, final boolean debug ) {
 		final JsonArray list = new JsonArray();
 		if ( exceptions != null && ! exceptions.isEmpty() ) {
 			for ( int i = 0; i < exceptions.size(); i++ ) {
-				final Throwable rootCause = getRootCause( exceptions.get( i ) );
+				final Exception ex = exceptions.get(i);
 				final StringWriter sw = new StringWriter();
-				sw.append( rootCause.getClass().getName() + ": " + rootCause.getMessage() );
+				final PrintWriter pw = new PrintWriter(sw);
+				pw.append( "Exception " + (i + 1) + ": " + ex.getMessage() );
```
hartig: There is no need to print `"Exception " + (i + 1) + ": "` here. That should be done by the client instead.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -230,6 +245,25 @@ protected void exec( final String serverURI,
 		}
 
 		if ( response.statusCode() != 200 ) {
```
hartig: I don't think this works. The HeFQUIN service would not respond with a status code different from 200 even if there have been exceptions during the query execution.

hartig (changes requested):
In addition to the comments in this review, I wonder what the output of `hefquin-client` with exceptions would actually look like after this PR, both with and without `--debug`. Did you try it? Can you please post some example output here (you may try with a query for which an authorization-based request fails, assuming that this causes an exception).

KentThang:
> In addition to the comments in this review, I wonder what the output of hefquin-client with exceptions would actually look like after this PR, both with and without --debug. Did you try it? Can you please post some example output here (you may try with a query for which an authorization-based request fails, assuming that this causes an exception).

Yes, I forgot to include some examples. 

Here is an example making a request with the wrong auth token, without `--debug`:
```
Exception 1: Exception occurred when outputting the result of a SELECT query using the Jena machinery.
Request failed with HTTP status 500
``` 
Here is with:
```
Exception 1: Exception occurred when outputting the result of a SELECT query using the Jena machinery.StackTrace:java.lang.Exception: Exception occurred when outputting the result of a SELECT query using the Jena machinery.
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execSelectQuery(HeFQUINEngine.java:354)
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execAndPrint(HeFQUINEngine.java:307)
        at se.liu.ida.hefquin.engine.HeFQUINEngine.executeQueryAndPrintResult(HeFQUINEngine.java:172)
        at se.liu.ida.hefquin.service.SparqlServlet.execute(SparqlServlet.java:250)
        at se.liu.ida.hefquin.service.SparqlServlet.executeRequest(SparqlServlet.java:161)
        at se.liu.ida.hefquin.service.SparqlServlet.doPost(SparqlServlet.java:107)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:520)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
        at org.eclipse.jetty.servlet.ServletHolder$NotAsync.service(ServletHolder.java:1410)
        at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:764)
        at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:529)
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:131)
        at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:578)
        at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:122)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:223)
        at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:1580)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:221)
        at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1381)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:176)
        at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:484)
        at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1553)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:174)
        at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1303)
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:129)
        at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:122)
        at org.eclipse.jetty.server.Server.handle(Server.java:563)
        at org.eclipse.jetty.server.HttpChannel$RequestDispatchable.dispatch(HttpChannel.java:1598)
        at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:753)
        at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:501)
        at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:287)
        at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:314)
        at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
        at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
        at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.runTask(AdaptiveExecutionStrategy.java:421)
        at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.consumeTask(AdaptiveExecutionStrategy.java:390)
        at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.tryProduce(AdaptiveExecutionStrategy.java:277)
        at org.eclipse.jetty.util.thread.strategy.AdaptiveExecutionStrategy.produce(AdaptiveExecutionStrategy.java:193)
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:969)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1194)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1149)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.IllegalStateException
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalitiesForRequests(RequestBasedCardinalityEstimator.java:263)
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalitiesForRequests(RequestBasedCardinalityEstimator.java:127)
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalities(RequestBasedCardinalityEstimator.java:43)
        at se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemoveSubPlansWithEmptyResults.apply(RemoveSubPlansWithEmptyResults.java:76)
        at se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicsBasedLogicalOptimizerImpl.optimize(HeuristicsBasedLogicalOptimizerImpl.java:73)
        at se.liu.ida.hefquin.engine.queryproc.impl.planning.QueryPlannerImpl.createPlan(QueryPlannerImpl.java:90)
        at se.liu.ida.hefquin.engine.queryproc.impl.QueryProcessorImpl.processQuery(QueryProcessorImpl.java:85)
        at se.liu.ida.hefquin.engine.queryproc.impl.QueryProcessorImpl.processQuery(QueryProcessorImpl.java:73)
        at se.liu.ida.hefquin.jenaintegration.sparql.engine.main.OpExecutorHeFQUIN$MainQueryIterator.nextStage(OpExecutorHeFQUIN.java:208)
        at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.makeNextStage(QueryIterRepeatApply.java:100)
        at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.hasNextBinding(QueryIterRepeatApply.java:60)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIterSlice.hasNextBinding(QueryIterSlice.java:76)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:37)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:37)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.exec.RowSetStream.hasNext(RowSetStream.java:48)
        at org.apache.jena.riot.rowset.rw.RowSetWriterJSON$ResultSetWriterTableJSON.writeRows(RowSetWriterJSON.java:143)
        at org.apache.jena.riot.rowset.rw.RowSetWriterJSON$ResultSetWriterTableJSON.write(RowSetWriterJSON.java:130)
        at org.apache.jena.riot.rowset.rw.RowSetWriterJSON.write(RowSetWriterJSON.java:89)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:186)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:156)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:149)
        at org.apache.jena.sparql.util.QueryExecUtils.outputResultSet(QueryExecUtils.java:158)
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execSelectQuery(HeFQUINEngine.java:351)
        ... 40 more

Request failed with HTTP status 500
```

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
KentThang: It works for the case where the auth token is incorrect, but maybe we could also print exceptions even if the response status code is 200? I've pushed a suggestion in latest commit.

hartig:
Thanks for posting the example output! Two observations about it: 
1. I wonder why the "Request failed with HTTP status 500" part appears on a separate line (in both cases), where in the case of using `--debug`, it even appears after the stack trace. How does the output for the two cases look like if you use `hefquin` instead of `hefquin-client`?
2. For the case of using `--debug`, the first line of the stack trace is appended into the line of the exception message. This needs to be changed; the stack trace should begin in its own line as is certainly the case if you use `hefquin` instead of `hefquin-client`. There are two ways to achieve that: One option is to simply add a newline character (line separator) into the string that is put together in `getExceptions` of `ServletUtils`. The other option is to extend the format of the JSON array that `getExceptions` produces; instead of simply creating an array of strings (one per exception), create an array of JSON objects where each of these objects has three fields: i) `type`, with `ex.getClass().getName()` as the value, ii) `message` (or simply `msg`), with `ex.getMessage()` as the value, and iii) `stacktrace` (or `trace`), with the stack trace as the value. I would prefer the second option because it represents the transmitted data in a more structured way. Of course, in this case, you will then also have to adapt/expand the client-side code that receives the JSON array.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
hartig: > It works for the case where the auth token is incorrect, but maybe we could also print exceptions even if the response status code is 200?

Really? I wonder in which cases the HeFQUIN service (`SparqlServlet`) responds with a status code different from 200? And what would this status code be?

Regarding the auth token, the token may be incorrect for only one of the fed.members; other parts of the overall query (over the federation) may still succeed and we may get a nonempty query result based on these parts. As the most simple example of such a case, consider the following query.
```
SELECT * WHERE {
  { SERVICE <http://example.org/endpoint> { ?s ?p ?o } }
  UNION
  { SERVICE <http://example.com/endpoint> { ?s ?p ?o } }
}
```
Hence, there is not even a join in this query. Now assume that the `example.org` endpoint does not require an auth token but the `example.com` one does, and the authorization for that one fails when HeFQUIN executes the query. As a consequence, HeFQUIN cannot output solution mappings for the second SERVICE clause, but it can output the solution mappings obtained for the first SERVICE clause. Therefore, if `hefquin-client` requests the execution of the query from the HeFQUIN service  (`SparqlServlet`), the response of the service should be produced query result (i.e., the solution mappings obtained for the first SERVICE clause) together with the information about exceptions that occurred while producing the result (such as the auth-related exception). And then it doesn't make sense that the status code for this response is anything else than 200.

KentThang:
> I wonder why the "Request failed with HTTP status 500" part appears on a separate line (in both cases), where in the case of using --debug, it even appears after the stack trace.

Ah, that would be because of this I assume:
```
if ( response.statusCode() != 200 ) {
	final String errorBody;
	try {
		errorBody = new String(
			response.body().readAllBytes(),
			StandardCharsets.UTF_8
		);
	}
	catch ( final IOException e ) {
		cmdError( "Failed to read error response from server: " + e.getMessage(), true );
		return;
	}

	final JsonObject errorJson = JSON.parse( errorBody ).getAsObject();
	final JsonArray errors = errorJson.get( "error" ).getAsArray();

	printExceptions(errors);

	cmdError( "Request failed with HTTP status " + response.statusCode(), true );
        return;
}
```
Notice how I haven't deleted the original printout `cmdError( "Request failed with HTTP status " + response.statusCode(), true );` and only added the `printException` stuff before it?

This is with bin/hefquin, with no `--debug`:
```
Attention: The query result may be incomplete because the following exception was caught when executing the query plan

Exception 1: Exception occurred when outputting the result of a SELECT query using the Jena machinery.
```
and with `--debug`:
```
Attention: The query result may be incomplete because the following exception was caught when executing the query plan

Exception 1: Exception occurred when outputting the result of a SELECT query using the Jena machinery.
StackTrace:
java.lang.Exception: Exception occurred when outputting the result of a SELECT query using the Jena machinery.
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execSelectQuery(HeFQUINEngine.java:354)
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execAndPrint(HeFQUINEngine.java:307)
        at se.liu.ida.hefquin.engine.HeFQUINEngine.executeQueryAndPrintResult(HeFQUINEngine.java:172)
        at se.liu.ida.hefquin.cli.RunQueryWithoutSrcSel.exec(RunQueryWithoutSrcSel.java:157)
        at org.apache.jena.cmd.CmdMain.mainMethod(CmdMain.java:86)
        at org.apache.jena.cmd.CmdMain.mainRun(CmdMain.java:55)
        at org.apache.jena.cmd.CmdMain.mainRun(CmdMain.java:43)
        at se.liu.ida.hefquin.cli.RunQueryWithoutSrcSel.main(RunQueryWithoutSrcSel.java:51)
Caused by: java.lang.IllegalStateException
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalitiesForRequests(RequestBasedCardinalityEstimator.java:263)
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalitiesForRequests(RequestBasedCardinalityEstimator.java:127)
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalities(RequestBasedCardinalityEstimator.java:43)
        at se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemoveSubPlansWithEmptyResults.apply(RemoveSubPlansWithEmptyResults.java:76)
        at se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicsBasedLogicalOptimizerImpl.optimize(HeuristicsBasedLogicalOptimizerImpl.java:73)
        at se.liu.ida.hefquin.engine.queryproc.impl.planning.QueryPlannerImpl.createPlan(QueryPlannerImpl.java:90)
        at se.liu.ida.hefquin.engine.queryproc.impl.QueryProcessorImpl.processQuery(QueryProcessorImpl.java:85)
        at se.liu.ida.hefquin.engine.queryproc.impl.QueryProcessorImpl.processQuery(QueryProcessorImpl.java:73)
        at se.liu.ida.hefquin.jenaintegration.sparql.engine.main.OpExecutorHeFQUIN$MainQueryIterator.nextStage(OpExecutorHeFQUIN.java:208)
        at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.makeNextStage(QueryIterRepeatApply.java:100)
        at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.hasNextBinding(QueryIterRepeatApply.java:60)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIterSlice.hasNextBinding(QueryIterSlice.java:76)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:37)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:37)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at java.base/java.util.Iterator.forEachRemaining(Iterator.java:132)
        at org.apache.jena.sparql.exec.RowSetStream.forEachRemaining(RowSetStream.java:79)
        at org.apache.jena.sparql.exec.RowSetMem.<init>(RowSetMem.java:74)
        at org.apache.jena.sparql.exec.RowSetMem.create(RowSetMem.java:44)
        at org.apache.jena.sparql.exec.RowSetStream.rewindable(RowSetStream.java:64)
        at org.apache.jena.riot.rowset.rw.RowSetWriterText.output$(RowSetWriterText.java:130)
        at org.apache.jena.riot.rowset.rw.RowSetWriterText.output(RowSetWriterText.java:118)
        at org.apache.jena.riot.rowset.rw.RowSetWriterText.output(RowSetWriterText.java:114)
        at org.apache.jena.riot.rowset.rw.RowSetWriterText.write(RowSetWriterText.java:59)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:186)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:156)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:149)
        at org.apache.jena.sparql.util.QueryExecUtils.outputResultSet(QueryExecUtils.java:158)
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execSelectQuery(HeFQUINEngine.java:351)
        ... 7 more
```

KentThang:
> The other option is to extend the format of the JSON array that getExceptions produces; instead of simply creating an array of strings (one per exception), create an array of JSON objects where each of these objects has three fields: i) type, with ex.getClass().getName() as the value, ii) message (or simply msg), with ex.getMessage() as the value, and iii) stacktrace (or trace), with the stack trace as the value. I would prefer the second option because it represents the transmitted data in a more structured way. Of course, in this case, you will then also have to adapt/expand the client-side code that receives the JSON array.

I'll work on this, the second option!

KentThang:
The latest commit resolves your comment regarding returning an array of JSON objects instead!
It now looks like so:
```
Exception 1: Exception occurred when outputting the result of a SELECT query using the Jena machinery.
StackTrace:
java.lang.Exception: Exception occurred when outputting the result of a SELECT query using the Jena machinery.
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execSelectQuery(HeFQUINEngine.java:354)
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execAndPrint(HeFQUINEngine.java:307)
        at se.liu.ida.hefquin.engine.HeFQUINEngine.executeQueryAndPrintResult(HeFQUINEngine.java:172)
        at se.liu.ida.hefquin.service.SparqlServlet.execute(SparqlServlet.java:250)
        at se.liu.ida.hefquin.service.SparqlServlet.executeRequest(SparqlServlet.java:161)
        at se.liu.ida.hefquin.service.SparqlServlet.doPost(SparqlServlet.java:107)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:520)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:587)
        at org.eclipse.jetty.servlet.ServletHolder$NotAsync.service(ServletHolder.java:1410)
        at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:764)
        at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:529)
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:131)
        at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:578)
        at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:122)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:223)
        at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:1580)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:221)
        at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1381)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:176)
        at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:484)
        at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1553)
        at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:174)
        at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1303)
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:129)
        at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:122)
        at org.eclipse.jetty.server.Server.handle(Server.java:563)
        at org.eclipse.jetty.server.HttpChannel$RequestDispatchable.dispatch(HttpChannel.java:1598)
        at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:753)
        at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:501)
        at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:287)
        at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:314)
        at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
        at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:969)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1194)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1149)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.IllegalStateException
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalitiesForRequests(RequestBasedCardinalityEstimator.java:263)
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalitiesForRequests(RequestBasedCardinalityEstimator.java:127)
        at se.liu.ida.hefquin.engine.queryproc.impl.cardinality.RequestBasedCardinalityEstimator.addCardinalities(RequestBasedCardinalityEstimator.java:43)
        at se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.heuristics.RemoveSubPlansWithEmptyResults.apply(RemoveSubPlansWithEmptyResults.java:76)
        at se.liu.ida.hefquin.engine.queryproc.impl.loptimizer.HeuristicsBasedLogicalOptimizerImpl.optimize(HeuristicsBasedLogicalOptimizerImpl.java:73)
        at se.liu.ida.hefquin.engine.queryproc.impl.planning.QueryPlannerImpl.createPlan(QueryPlannerImpl.java:90)
        at se.liu.ida.hefquin.engine.queryproc.impl.QueryProcessorImpl.processQuery(QueryProcessorImpl.java:85)
        at se.liu.ida.hefquin.engine.queryproc.impl.QueryProcessorImpl.processQuery(QueryProcessorImpl.java:73)
        at se.liu.ida.hefquin.jenaintegration.sparql.engine.main.OpExecutorHeFQUIN$MainQueryIterator.nextStage(OpExecutorHeFQUIN.java:208)
        at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.makeNextStage(QueryIterRepeatApply.java:100)
        at org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply.hasNextBinding(QueryIterRepeatApply.java:60)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIterSlice.hasNextBinding(QueryIterSlice.java:76)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:37)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorWrapper.hasNextBinding(QueryIteratorWrapper.java:37)
        at org.apache.jena.sparql.engine.iterator.QueryIteratorBase.hasNext(QueryIteratorBase.java:122)
        at org.apache.jena.sparql.exec.RowSetStream.hasNext(RowSetStream.java:48)
        at org.apache.jena.riot.rowset.rw.RowSetWriterJSON$ResultSetWriterTableJSON.writeRows(RowSetWriterJSON.java:143)
        at org.apache.jena.riot.rowset.rw.RowSetWriterJSON$ResultSetWriterTableJSON.write(RowSetWriterJSON.java:130)
        at org.apache.jena.riot.rowset.rw.RowSetWriterJSON.write(RowSetWriterJSON.java:89)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:186)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:156)
        at org.apache.jena.sparql.resultset.ResultsWriter.write(ResultsWriter.java:149)
        at org.apache.jena.sparql.util.QueryExecUtils.outputResultSet(QueryExecUtils.java:158)
        at se.liu.ida.hefquin.engine.HeFQUINEngine._execSelectQuery(HeFQUINEngine.java:351)
        ... 36 more

Request failed with HTTP status 500
```
However, note that even though I added the `type` with `exception.getClass().getName()` as value, it isn't used anywhere at the current moment and does the same sort of printout that it does in `bin/hefquin`. That being, only the exception message and if `--debug` was given, the stack trace.

hartig:
> > I wonder why the "Request failed with HTTP status 500" part appears on a separate line (in both cases), where in the case of using --debug, it even appears after the stack trace.
> 
> Ah, that would be because of this I assume:
> [...]
> Notice how I haven't deleted the original printout `cmdError( "Request failed with HTTP status " + response.statusCode(), true );` and only added the `printException` stuff before it?

Right. That makes sense. Thanks for figuring this out.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
KentThang: Ah, thanks for providing the example!

Do you think this case is covered now, now that I added the call to `printExceptions()` in the case where the status code would be 200?

> Really? I wonder in which cases the HeFQUIN service (SparqlServlet) responds with a status code different from 200? And what would this status code be?

I'll take a closer look at that if you want but so far I've only gotten 500, as you can see from the logs.

hartig:
> The latest commit resolves your comment regarding returning an array of JSON objects instead! It now looks like so:
> [...]

Great. That looks like it should do.

Thanks for implementing the more complex version with the array of JSON objects. I will take a look at the actual source code shortly.

> However, note that even though I added the `type` with `exception.getClass().getName()` as value, it isn't used anywhere at the current moment and does the same sort of printout that it does in `bin/hefquin`. That being, only the exception message and if `--debug` was given, the stack trace.

Yes, that's fine. When I wrote the description of what these JSON objects should look like, I thought that `type` can be useful to have in this data even if it is not printed by `hefquin-client`. I should have said that instead of only thinking it  ;-)

The last thing that remains a bit mysterious is why the HeFQUIN service responds with HTTP status 500 in this case. But I already think I have an idea for the reason.   ...

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
hartig: Yes, that case is covered now.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -230,6 +245,23 @@ protected void exec( final String serverURI,
...
+					response.body().readAllBytes(),
+					StandardCharsets.UTF_8
+				);
+			}
+			catch ( final IOException e ) {
+				cmdError( "Failed to read error response from server: " + e.getMessage(), true );
+				return;
+			}
+
+			final JsonObject errorJson = JSON.parse( errorBody ).getAsObject();
```
hartig: Given that we do essentially the same also right after this if-block, can we combine both of them and move them right before the if-block?

hefquin-service/src/main/java/se/liu/ida/hefquin/service/ServletUtils.java
```diff
@@ -55,59 +57,57 @@ public static ResultsFormat convert( final String mimeType ) {
 	}
 
 	/**
-	 * Converts the exceptions of the given object into a JSON array, including
-	 * class name, message, and full stack trace for each exception.
+	 * Converts the exceptions of the given object into a JSON array.
+	 * Each exception is represented by its message and, if debug mode
+	 * is enabled, the full stack trace for each exception.
```
hartig: Please modify this Javadoc comment to reflect that the method now returns a JSON array of JSON objects.

Same for the Javadoc comment of the next method.

hefquin-service/src/main/java/se/liu/ida/hefquin/service/ServletUtils.java
```diff
@@ -55,59 +57,57 @@ public static ResultsFormat convert( final String mimeType ) {
...
-				list.add( sw.toString() );
+				final JsonObject exception = new JsonObject();
+				final Exception ex = exceptions.get(i);
+				exception.put( "type", ex.getClass().getName() );
+				exception.put( "msg", ex.getMessage() );
+
+				if ( debug ) {
+					final StringWriter sw = new StringWriter();
+					final PrintWriter pw = new PrintWriter(sw);
+					pw.println( "StackTrace:" );
```
hartig: Please remove this here and, instead, have the client print this line (i.e., in `printExceptions` of `RunHttpQuery`).

hartig (changes requested):

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
KentThang: I wasn't sure if I fully understood this comment. Do you mean move both `printException` calls before the if-block like I have done in the latest commit?

hartig: No. I meant to move only the `JSON.parse` statement before the if-block. I will do another review of the current version and indicate explicitly where things should go.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -229,13 +244,21 @@ protected void exec( final String serverURI,
 			return;
 		}
 
+		final JsonObject obj = JSON.parse(response.body());
+
+		final JsonValue exceptions = obj.get( HttpConstants.JSON_EXCEPTIONS );
+		if ( exceptions != null )
+			printExceptions( exceptions.getAsArray() );
```
hartig: This part should be moved after printing the query result; i.e., after the line that does `QueryExecUtils.outputResultSet`. This way, we would get the same order in which things are printed as we have in the `hefquin` CLI program (`RunQueryWithoutSrcSel`).

Additionally, exactly as in `RunQueryWithoutSrcSel`, you should include something like:
```java
System.err.println( "Attention: The query result may be incomplete because the following "
						+ numberOfExceptions + " exceptions were caught when executing the query plan." );
```
.. into the part where the exceptions are printed after printing the query result.

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -229,13 +244,21 @@ protected void exec( final String serverURI,
...
+
+		final JsonValue error = obj.get( "error" );
+		if ( error != null )
+			printExceptions( error.getAsArray() );
```
hartig: This part should be moved into the if-block for `response.statusCode() != 200`. There, put it **_after_** the `cmdError` that mentions the failed request and, in between them, you add another `cmdError` that says something like: "In its response, the HeFQUIN service has returned the following X exceptions that occurred during the execution of the query."

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -229,13 +244,21 @@ protected void exec( final String serverURI,
...
+
 		if ( response.statusCode() != 200 ) {
 			cmdError( "Request failed with HTTP status " + response.statusCode(), true );
```
hartig:
```suggestion
			cmdError( "The HeFQUIN service did not execute the given query but, instead, responded with HTTP status " + response.statusCode(), true );
```

hartig (changes requested):

hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
KentThang: Do we want to only include that additional print in the case where exceptions are printed after printing the query result? Or do we also want to print it in the case where `response.statusCode() != 200`? I.e., I would add 
```
System.err.println( "Attention: The query result may be incomplete because the following "
						+ numberOfExceptions + " exceptions were caught when executing the query plan." );
```
to the very start of the `printExceptions()` method.

KentThang: Ah, with the other comment you left, I guess you meant only in the case after printing out the query result...

hartig: > Ah, with the other comment you left, I guess you meant only in the case after printing out the query result...

Yupp :-) (That's also why I explicitly wrote the following in my comment above: ".. into the part where the exceptions are printed after printing the query result.")

hartig (approved):
Looks good now.

PR #711, opened 2026-09-11
Update RequestBasedCardinalityEstimator to new DataRetrievalResponse interface

### LINKED ISSUE #710: Update RequestBasedCardinalityEstimator to new DataRetrievalResponse interface
In the discussion of PR #708 I have been wondering why the HeFQUIN service responds with HTTP status code 500 for the case in which the authorization check fails. The reason is that I have not yet updated `RequestBasedCardinalityEstimator` to the changes that PR #706 has done in the `DataRetrievalResponse` interface. Especially, in line 258, `RequestBasedCardinalityEstimator` calls `getCardinality()` of a cardinality response without check first whether that response is "defective". There is only a check whether the response was an HTTP error. The fix should be easy: change line 247 from
```java
if ( resps[i].isError() ) {
```
to
```java
if ( resps[i].isError() || resps[i].isDefective() ) {
```
Additionally, line 325 needs to be changed in the same way.

@KentThang Can you please make this change and do your test again to see whether this fixes the issue. Do the test both with `hefquin-client` (for which you need a running `hefquin-service`) and with `hefquin`.

KentThang:
Closes #710.

The commit chain looks a bit silly because I thought I had pushed it to the wrong branch.

I also noticed a bug while working on this, that being the `isDebug` lacking braces in `printExceptions()`, that has been fixed.

### DIFF
--- a/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
+++ b/hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
@@ -435,16 +435,13 @@ public class RunHttpQuery extends CmdGeneral
 	 */
 	protected void printExceptions( final JsonArray exceptions ) {
 		for ( int i = 0; i < exceptions.size(); i++ ) {
-			System.err.println();
-
 			final JsonObject exception = exceptions.get(i).getAsObject();
 
 			cmdError( "Exception " + (i + 1) + ": " + exception.getString("msg"), false );
 
-			if ( isDebug() ) {
+			if ( isDebug() )
 				cmdError( "StackTrace:", false );
 				cmdError( exception.getString("stacktrace"), false );
-			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


hefquin-cli/src/main/java/se/liu/ida/hefquin/cli/RunHttpQuery.java
```diff
@@ -435,13 +435,16 @@ private static String extractOnelineTimeStats( final JsonValue statsAndException
 	 */
 	protected void printExceptions( final JsonArray exceptions ) {
 		for ( int i = 0; i < exceptions.size(); i++ ) {
+			System.err.println();
```
hartig: This should be done via `cmdError` exactly as the other statement related to printing that are here.

hartig (changes requested):
Just a (very!) minor thing.

hartig (approved):

PR #715, opened 2026-09-17
Add ignore caching arguments to hefquin-client and server

### LINKED ISSUE #685: Argument for CLI programs to ignore the cache

KentThang:
Closes #685.

I believe this is all that needs to be done but I am not sure how to verify.


hartig:
The code looks good and I agree that this should be all that needs to be done. Nonetheless, can you please test whether it really works, which you can do as follows.
1. Delete the `cache` directory.
2. Execute one of the example queries (e.g., `examples/ExampleQuery1.rq`) with `hefquin-client` using `--printFedAccessStats`. Notice that `overallNumberOfCacheHits` should be 0.
3. Execute the same query again in the same way. Now, `overallNumberOfCacheHits` should be some number greater than 0.
4. Execute the query a third time, but now with `--ignoreRetrievalCache` and `--ignoreCardinalityCache` as additional argument. Now, `overallNumberOfCacheHits` should be 0 again.

KentThang:
Hmm, `overallNumberOfCacheHits` isn't 0 again after executing a third time, but the number of `overallNumberOfCacheHits` does not increase with `--ignoreRetrievalCache` and `--ignoreCardinalityCache`. 

I looked at `bin/hefquin` to see if the same behaviour exhibited there but it does indeed seem to reset to 0, not just prevent the number from increasing. Do you know why this could be?

I'll take another look at #685 to see if I missed anything.

hartig:
Ah! Sorry! The reason is that the federation access statistics of `hefquin-server` are not cleared between the queries that `hefquin-client` sends, whereas `hefquin` starts always with fresh federation access statistics. Therefore, ...
> Hmm, overallNumberOfCacheHits isn't 0 again after executing a third time, but the number of overallNumberOfCacheHits does not increase with --ignoreRetrievalCache and --ignoreCardinalityCache.

.. is indeed the correct/expected behavior when doing this test with `hefquin-client`.

I will merge the PR.

hartig (approved):

